PR #155, opened 2018-03-01
Redirect output

### LINKED ISSUE #60: Ability to redirect stdout and stderr (tee program output)
The functionality provided by the `RedirectOutput` debug option in hew older version of PTVSD doesn't seem to be supported by the PyDev debugger.

From what I can tell this functionality is only available in `attach` scenarios. I guess this is because the debugger (PyDev) is always launched inside a terminal.

FYI - We're also missing the corresponding protocol implementation in the wrapper. I tried implementing this, however it was a moot point as there's no way to launch a program with redirection enabled as there's no way to do this in PyDev (I could be mistaken).

DonJayamanne:
Fixes #60

**Time to discuss:**

* Names of settings that are passed into the debugger
I've created a single setting that gets passed via the JSON launch args `redirectOutput`
If this is true, we redirect both stdout and stderr.
However in PyDevd i've maintained the ability to configure redirection of stdout and stderr separately. I.e. you could redirect one and not the other (as is done today).
However I'm not allowing this in the json protocol (single setting `redirectOutput`) or should I?

* How and where do we document these launch arguments 

* Command Ids that we introduce into PyDev
I've added a command `CMD_REDIRECT_OUTPUT` with the value `200` into pydevd

* Changes to PyDevd in general
Is the current approach viable. How much are we allowed/supposed to change (diverge from) pydev.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -38,6 +38,7 @@ __version__ = "4.0.0a1"
 #ipcjson._TRACE = ipcjson_trace
 
 ptvsd_sys_exit_code = 0
+WAIT_FOR_DISCONNECT_REQUEST_TIMEOUT = 2
 WAIT_FOR_THREAD_FINISH_TIMEOUT = 1
 
 def unquote(s):
@@ -390,7 +391,8 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         self.next_var_ref = 0
         self.loop = futures.EventLoop()
         self.exceptions_mgr = ExceptionsManager(self)
-
+        self.disconnect_request = None
+        self.disconnect_request_event = threading.Event()
         pydevd._vscprocessor = self
         self._closed = False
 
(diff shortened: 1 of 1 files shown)


int19h:
This needs to be split into two separate PRs - one for the bits inside `pydevd`, and one for everything else. This is so that the pydevd change can be submitted upstream later (via merge subtree). 

For launch arguments in JSON, would it be better to have some kind of namespacing there for our custom stuff? Or is all of it assumed to be specific to the debug adapter in use? That aside, unless there's some other adapter that has a similar option, I think we can just go with whatever's most convenient - which in this case would be a single option.

DonJayamanne:
> Or is all of it assumed to be specific to the debug adapter in use?

Everything in `arguments` section of `launch` request is specific to the debug adapter.

DonJayamanne:
>This needs to be split into two separate PRs - one for the bits inside pydevd, and one for everything else

I've created a separate PR #156 
Once that's merged I'll update this PR accordingly.

DonJayamanne:
@int19h 
PR has been updated, this PR contains changes scoped just to PTVSD code (`wrapper.py`).

int19h:
I think it's not supposed to have any commits that go across folder boundary for upstream merge to work properly later. Can you rebase on master and squash before merging?

Other than that, it's good to go.

DonJayamanne:
@int19h 
Ah bummer, tried to clean everything out (by resetting head and force pushing) and that didn't work.
Guess github doesn't like force pushing inside a PR.
Have created another PR #157, please review and approve.

PR #158, opened 2018-03-01
ExceptionInfo returns name and description obtained from __exception__ object

karthiknadig:
Raised exceptions:
image

Uncaught exceptions:
image

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -877,52 +877,20 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
     def on_exceptionInfo(self, request, args):
         # TODO: docstring
         pyd_tid = self.thread_map.to_pydevd(args['threadId'])
-        with self.stack_traces_lock:
-            try:
-                xframes = self.stack_traces[pyd_tid]
-            except:
-                xframes = None
-
-        name = None
-        description = None
-        if xframes:
-            # pydevd adds a __exception__ object to the frame at the top of the stack
-            # Always extract from that object to get accurate exception info.
-            try:
-                xframe = xframes[0]
-                pyd_fid = xframe['id']
-                cmdargs = '{}\t{}\tFRAME\t__exception__'.format(pyd_tid, pyd_fid)
-                _, _, resp_args = yield self.pydevd_request(pydevd_comm.CMD_GET_VARIABLE, cmdargs)
-                xml = untangle.parse(resp_args).xml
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -849,8 +849,33 @@ def on_setExceptionBreakpoints(self, request, args):
     @async_handler
     def on_exceptionInfo(self, request, args):
         # TODO: docstring
-        tid = self.thread_map.to_pydevd(args['threadId'])
-        name, description = self.get_exceptionInfo(tid)
+        pyd_tid = self.thread_map.to_pydevd(args['threadId'])
+        with self.stack_traces_lock:
```
DonJayamanne: Shouldn't all of this code go into the get_exceptionInfo method. 
Anyone using `get_exceptionInfo` won't get the right exception information

karthiknadig: pydev_request call cannot be made from the events like suspended event.

DonJayamanne: hmm, guess we'll need to find another solution for VSC.
Either way, i think we should move the code into the same method and make the decision inside that method whether to call pydevd_requests or not.
This way, the logic to get exception info is in one place.

karthiknadig: @int19h I want to make this work for `on_pydevd_thread_suspend` function. But i cannot currently move this to `get_exceptionInfo` because of the `yield`. What can we do here so that this can work for both `on_exceptionInfo` request and  `on_pydevd_thread_suspend` event?

int19h: I think `get_exceptionInfo` should be an iterator itself, then it can `yield` inside, and its caller will need to `yield` it to obtain the result. 

You'll need to make the event handler itself async for that, though. It looks like you should be able to just slap `@async_handler` on it.

karthiknadig: May be we can add `run_until_complete` to `EventLoop`?

int19h: You mean, until a specific async task is complete? But that's exactly what `yield` does, if you apply async_handler. Or do we need to not return control to pydevd main loop from `on_pydevd_thread_suspend` until we got the exception info?

karthiknadig: @int19h that worked. FYI,
This worked:
```
    @pydevd_events.handler(pydevd_comm.CMD_THREAD_SUSPEND)
    @async_handler
```
This did not:
```
    @async_handler
    @pydevd_events.handler(pydevd_comm.CMD_THREAD_SUSPEND)
```

PR #159, opened 2018-03-02
Preserve file path case on windows for file paths returned

### LINKED ISSUE #88: Breaking on exception opens a duplicate editor window without breakpoints

* Contents of file:
```python
print(x)
```
* Start debugging

DonJayamanne:
Fixes #88

### DIFF
--- a/ptvsd/pathutils.py
+++ b/ptvsd/pathutils.py
@@ -1,0 +1,65 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+from __future__ import print_function, with_statement, absolute_import
+
+import sys
+import os.path
+import platform
+from glob import glob
+
+
+__author__ = "Microsoft Corporation <ptvshelp@microsoft.com>"
+__version__ = "4.0.0a1"
+
+MAX_FILES_TO_CACHE = 1000
+
+class PathUnNormcase(object):
+    """Ensures path names of files are returned as they exist on the fs."""
+
+    def __init__(self):
+        self._dict = {}
+        self._enabled = False
+
+    def enable(self):
+        self._enabled = platform.system() == 'Windows'
+        
+    def un_normcase(self, file_path):
+        if not self._enabled or len(file_path) == 0:
+            return file_path
+        if file_path in self._dict:
+            return self._dict[file_path]
+        file_path_to_return = self._get_actual_filename(file_path)
+        self.track_file_path_case(file_path_to_return)
+        return file_path_to_return
+
+    def track_file_path_case(self, file_path):
+        if not self._enabled:
+            return
+        if len(self._dict) > MAX_FILES_TO_CACHE:
+            self._dict.clear()
+        self._dict[file_path] = file_path
+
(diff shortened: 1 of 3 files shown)


ptvsd/pathutils.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+from __future__ import print_function, with_statement, absolute_import
```
int19h: No need for `with_statement` (it's from the days when we had to support Python 2.5 etc).

ptvsd/wrapper.py
```diff
@@ -545,12 +546,27 @@ def on_configurationDone(self, request, args):
...
+        Process the launch arguments to configure the debugger.
+        Further information can be found here https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes
+        {
+            type:'python',
+            request:'launch'|'attach',
+            name:'friendly name for debug config',
+            // Custom attributes supported by PTVSD.
+            redirectStdout:true,
+            redirectStderr:true,
+            fixFilePathCase:false
```
int19h: We might as well do this always, or depending on the platform, since there's no reason why a protocol client would prefer lowercased path (if they are doing case-insensitive comparisons, they just need to do it on their end).

int19h:
I wonder if we could make this a lot easier if we can just add an option to not lowercase filenames to pydevd itself. Doesn't even have to be in the protocol - since the wrapper is in the same process, it can just reach out and flip a global variable or something.

ptvsd/wrapper.py
DonJayamanne: > We might as well do this always, or depending on the platform

We don't need this for VS, this is a problem in VS Code. 

> since there's no reason why a protocol client would prefer lowercased path

Agreed, however on Windows, file paths are meant to be case insensitive. I'd rather have this a feature that can be used on demand.

int19h: A Windows client that needs case-insensitive paths should use case-insensitive comparisons for the case-preserved paths that it receives, rather than relying on lowercased paths to begin with. VS is already doing that, so I wouldn't expect it to be negatively affected by paths not being lowercased.

On the other hand, lowercasing will sometimes be visible to the user even in VS, if not quite as annoying as in VSC. For example, suppose that you have a breakpoint or an exception inside some file that is *not* opened in VS. It will open it when it receives the stop event - but the file name and path in the tab will use the path as reported by debugger, rather than the proper case of the file on disk.

mingwandroid: > Agreed, however on Windows, file paths are meant to be case insensitive

This is not strictly true. Windows is case-insensitive but case-preserving and that distinction is very important. Any time code ever messes with the case of a filename and expects to be able to access it after is either broken at worst and unnecessary at best.

It is also [very easy to tweak Windows to be case-sensitive](https://superuser.com/questions/266110/how-do-you-make-windows-7-fully-case-sensitive-with-respect-to-the-filesystem), and I've ran Windows systems like this for years. It aids interoperability greatly not doing this.

PR #162, opened 2018-03-02
Preserve file path case on windows for file paths returned

### LINKED ISSUE #88: Breaking on exception opens a duplicate editor window without breakpoints

DonJayamanne:
Fixes #88

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -22,7 +22,6 @@ except Exception:
 import _pydevd_bundle.pydevd_comm as pydevd_comm
 import _pydevd_bundle.pydevd_extension_api as pydevd_extapi
 import _pydevd_bundle.pydevd_extension_utils as pydevd_extutil
-import pydevd_file_utils
 #from _pydevd_bundle.pydevd_comm import pydevd_log
 
 import ptvsd.ipcjson as ipcjson
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -512,11 +512,8 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
     def on_initialize(self, request, args):
         # TODO: docstring
         cmd = pydevd_comm.CMD_VERSION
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -514,6 +515,8 @@ def on_initialize(self, request, args):
         os_id = 'WINDOWS' if platform.system() == 'Windows' else 'UNIX'
         msg = '1.1\t{}\tID'.format(os_id)
         yield self.pydevd_request(cmd, msg)
+        #Ensure pydev does not normcase file paths
+        pydevd_file_utils.set_ide_os('UNIX')
```
int19h: Can't we just pass it as `os_id` above?

DonJayamanne: Not following you.

int19h: The few lines above where we send the handshake with `os_id` inside - the handler of that message inside `pydevd` will invoke `set_ide_os` with the string that it gets (and, so far as I can tell, that's the only thing that it does to that string). So you should be able to just unconditionally set `os_id` to `UNIX` to get that result.

int19h: https://github.com/Microsoft/ptvsd/blob/855559370ee44cb4f7925a83597a22ca443879cc/ptvsd/pydevd/_pydevd_bundle/pydevd_process_net_command.py#L73

DonJayamanne: Hmm, you're right, but initially when it was set to just `WINDOWS` breaktpoints didn't work on the Mac, when when I had to conditionally set `os_is` to `WINDOWS|UNIX` based on the os.

int19h: The code that sets it currently is not quite right in any case, because it uses the current platform of the debugged process, while the intent of that message is to communicate the current platform of the IDE.

But I think it should work right, so long as IDE does all comparisons in a case-insensitive way on its side - which VS does. Worst case, we might need to ensure that pathnames that come from the IDE (when creating breakpoints, for example) are properly cased.

Issue #163, opened 2018-03-02
Debugger doesn't stop at breakpoints


huguesv:
On Visual Studio. It's not stopping at breakpoints in VS (it works in VSCode).

Log from VS:
```
 1> [DebugAdapter] <--   R (launch-2): {"type": "response", "seq": 2, "request_seq": 2, "success": true, "command": "launch", "message": "", "body": {}}
 1> [DebugAdapter] --> C (setExceptionBreakpoints-3): {"command":"setExceptionBreakpoints","arguments":{"filters":[],"exceptionOptions":[{"path":[{"names":["Python Exceptions"]}],"breakMode":"userUnhandled"},{"path":[{"names":["Python Exceptions"]},{"names":["KeyError","AttributeError","IndexError","StopIteration","GeneratorExit"]}],"breakMode":"unhandled"}]},"seq":3,"type":"request"}
 1> [DebugAdapter] <--   R (setExceptionBreakpoints-3): {"type": "response", "seq": 3, "request_seq": 3, "success": true, "command": "setExceptionBreakpoints", "message": "", "body": {}}
 1> [DebugAdapter] --> C (configurationDone-4): {"command":"configurationDone","arguments":{},"seq":4,"type":"request"}
 1> [DebugAdapter] <--   R (configurationDone-4): {"type": "response", "seq": 4, "request_seq": 4, "success": true, "command": "configurationDone", "message": "", "body": {}}
 1> [DebugAdapter] <--   E (process): {"type": "event", "seq": 5, "event": "process", "body": {"name": "C:\\Users\\huvalo\\source\\repos\\PythonApplication67\\PythonApplication67\\PythonApplication67.py", "systemProcessId": 19820, "isLocalProcess": true, "startMethod": "launch"}}
 1> [DebugAdapter] <--   E (exited): {"type": "event", "seq": 6, "event": "exited", "body": {"exitCode": 0}}
 1> [DebugAdapter] <--   E (terminated): {"type": "event", "seq": 7, "event": "terminated", "body": {}}
 1> [DebugAdapter] --> C (disconnect-5): {"command":"disconnect","arguments":{},"seq":5,"type":"request"}
 1> [DebugAdapter] <--   R (disconnect-5): {"type": "response", "seq": 8, "request_seq": 5, "success": true, "command": "disconnect", "message": "", "body": {}}

```

Log from VSCode:
```
2:36:42 PM, 3/2/2018
[...]

PR #166, opened 2018-03-03
Fix file path case on windows

### LINKED ISSUE #88: Breaking on exception opens a duplicate editor window without breakpoints

DonJayamanne:
Fixes #88 
@int19h 
I had to revert the change where we were setting the `os_id` to `UNIX` even on `Widows OS`.
This was because breakpoints didn't work at all on Windows for Visual Studio (#163).
However breakpoints continued to work for VS Code.

**We have two solutions:**

1. Ensure `os_id` is always set to `UNIX` for VS Code (on all OS).
However in light of the recent issue in VS (#163), I'm tempted not to go down this path. 
Its better to have duplicate editors than not having breakpoints being hit at all. Who knows what else could break.

2. The solution in the PR
Not elegant, but i believe this is the safest, at least until we fully understand why changing `os_id` to 'UNIX` caused breakpoints to fail on Windows for VS and NOT for VSCode.
Possibly because the breakpoints were sent to VS Code as `

### DIFF
--- a/ptvsd/pathutils.py
+++ b/ptvsd/pathutils.py
@@ -1,0 +1,65 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+from __future__ import print_function, with_statement, absolute_import
+
+import sys
+import os.path
+import platform
+from glob import glob
+
+
+__author__ = "Microsoft Corporation <ptvshelp@microsoft.com>"
+__version__ = "4.0.0a1"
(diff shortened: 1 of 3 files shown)


int19h:
Looking at #163, it appears that the issue is that they're still lowercasing the drive letter somehow? Basically, VS path comes with `C:\...`, but VSCode says `c:\...`.

Anyway, let's go with this for now for preview 1, but look into what it would take to just make pydevd do the right thing here.

Issue #65, opened 2018-02-13
Tool tips try to evaluate invalid expressions


zooba:
![image](https://user-images.githubusercontent.com/1693688/36126945-c3813694-100f-11e8-910a-f731eb366601.png)
![image](https://user-images.githubusercontent.com/1693688/36126951-d16d9da6-100f-11e8-8c1f-0cfa1443d4e6.png)

We should probably not return anything when a `SyntaxError` is raised while evaluating, and also do the extra effort to make sure that we don't evaluate function calls here.

karthiknadig:
fixed by #168

PR #168, opened 2018-03-03
Checks for isErrorOnEval before responding to evaluate requests

karthiknadig:
Fixes#65

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -524,6 +524,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
             supportsConditionalBreakpoints=True,
             supportsSetVariable=True,
             supportsExceptionOptions=True,
+            supportsEvaluateForHovers=True,
             exceptionBreakpointFilters=[
                 {
                     'filter': 'raised',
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -777,6 +777,13 @@ def on_evaluate(self, request, args):
         xml = untangle.parse(resp_args).xml
         xvar = xml.var
 
+        if xvar['isErrorOnEval']=='True':
+            self.send_response(
+                request,
+                result=None,
```
int19h: What happens if the user enters an invalid expression in Watch - it would seem that they don't get any information about what the error is like?

For this particular bug (#65), it seems that the root cause is improper detection of what constitutes a hover expression in VS (i.e. it shouldn't be sending things like `def` for evaluation in the first place) - this is what `DetectSideEffectsWalker` and `GetDataTipText` in PTVS is about. If possible, we should just figure out how to hook that up into the VSCode debug adapter.  

For cases where there genuinely is an error, we do want to report an error (like the old debugger does in Watch if you put an invalid expression there). The only thing missing is the appropriate error icon (instead of the usual variable icon) on the result - I think this was in `VariablePresentationHint` in the protocol.

karthiknadig: I will check with the adapter host folks about improper detection of hover expressions. But to isolate hover cases, how about we check the `EvaluateArguments.enumDescriptions`, and only return the `None` result for those cases?

That way Watch will still get the error. But i did not see a way to provide error in `VariablePresentationHint`. I will check with the adapter host folks.

int19h: Indeed, it doesn't seem to be in the procotol. Which is strange, since it's mostly just a mapping of VS constants for variable types, and that does include errors. Must be an oversight. 

`enumDescriptions` sounds like a good way to go - then we'll have defense in depth on ptvsd side even after we fix the hover expression boundaries.

PR #174, opened 2018-03-05
Fix coverage.

ericsnowcurrently:


tests/__main__.py
```diff
@@ -39,23 +52,27 @@ def convert_argv(argv):
...
+    if runtests:
+        # We make the "executable" a single arg because unittest.main()
+        # doesn't work if we split it into 3 parts.
+        cmd = [sys.executable + ' -m unittest']
+        if not modules and not help:
+            # Do discovery.
+            if quick:
+                start = os.path.join(TEST_ROOT, 'ptvsd')
+            elif sys.version_info[0] != 3:
+                start = os.path.join(TEST_ROOT, 'ptvsd')
```
int19h: This is identical to the branch above - should it be a single branch with an `or`?

ericsnowcurrently: It doesn't really matter.  Personally I prefer to keep the distinct cases separate, even if they do the same thing currently.

Issue #78, opened 2018-02-15
Debugger not pausing in VS Code Unit tests


DonJayamanne:
Pausing works in the VS Code UI, however not when running the unit tests
The disconnect request is sent by the test adapter after waiting 5 minutes for the `stop` event.
As you can see the `stop` event is not sent back.

* Here's a log of all the communication (from `ipcjson.py`)
```
Input
{"type":"request","seq":1,"command":"initialize","arguments":{"adapterID":"pythonExperimental","linesStartAt1":true,"columnsStartAt1":true,"pathFormat":"path"}}
Output
b'{"type": "response", "seq": 0, "request_seq": 1, "success": true, "command": "initialize", "message": "", "body": {"supportsExceptionInfoRequest": true, "supportsConfigurationDoneRequest": true, "supportsConditionalBreakpoints": true, "supportsSetVariable": true, "supportsExceptionOptions": true, "exceptionBreakpointFilters": [{"filter": "raised", "label": "Raised Exceptions", "default": "true"}, {"filter": "uncaught", "label": "Uncaught Exceptions", "default": "true"}]}}'
Output
b'{"type": "event", "seq": 1, "event": "initialized", "body": {}}'
Input
{"type":"request","seq":3,"command":"configurationDone"}
Output
b'{"type": "response", "seq": 2, "request_seq": 3, "success": true, "command": "configurationDone", "message": "", "body": {}}'
Input
{"type":"request","seq":4,"command":"threads"}
Output
b'{"type": "response", "seq": 3, "request_seq": 4, "success": true, "command": "threads", "message": "", "body": {"threads": [{"id": 1, "name": "MainThread"}]}}'
Input
{"type":"request","seq":5,"command":"pause","arguments":{"threadId":1}}
Output
b'{"type": "response", "seq": 4, "request_seq": 5, "success": true, "command": "pause", "message": "", "body": {}}'
Input
{"type":"request","seq":6,"command":"disconnect"}
Output
b'{"type": "response", "seq": 5, "request_seq": 6, "success": true, "command": "disconnect", "message": "", "body": {}}'
Output
b'{"type": "event", "seq": 6, "event": "exited", "body": {"exitCode": 0}}'
Output
b'{"type": "event", "seq": 7, "event": "terminated", "body": {}}'
```

PR #179, opened 2018-03-06
Partial fix for #78

int19h:
Do not send 'process' event to the IDE until pydevd reports the first thread

(The other half will be the unit test fixes, to wait on the `process` event before pausing)


int19h:
@ericsnowcurrently, Travis blocks it because of a line exceeding the 80 char limit - can we disable that rule?

ericsnowcurrently:
@int19h, we can disable the line-length limit ([add E501 to the "ignored" list](https://github.com/Microsoft/ptvsd/blob/master/tests/__main__.py#L102)) or [make it larger](http://flake8.pycqa.org/en/2.6.0/config.html).  Alternately, we could stick with the default limit ([as recommended by PEP 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)) and apply the "# noqa" marker to the occasional lines where it makes sense to just leave them long (e.g. literals, URLs in docstrings).

Personally, I prefer the 80 column limit for a variety of reasons, including:

* long lines usually pack a lot of behavior into one line, making it harder to read and to refactor (I try to stick to 1 operation per line); with Python, doing so is less beneficial than in other languages
* past 80 characters it gets gradually harder (as a reader) to follow most lines from start to finish
* 80-column-lines is the Python community standard, as provided by PEP 8 (though not nearly as proscriptive as Go's style guidelines)
* habit :)  (due to the community standard)
* the shorter the lines, the more files I can have on my screen at once; I routinely have 6 or more editor windows up side-by-side (and sometimes up to 15) and long lines are an increasing pain point the more files I'm looking at
* [PEP 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) summarize other arguments for the 80 column limit

All that said, using a longer (or no) line length limit isn't going to hurt my feelings. :)  Obviously, some of my above reasons are personal preference.  Furthermore, none of the other reasons dictate that the project will be in danger if we use longer lines.  I just wanted to clarify my preference and the reasons for it before we settle on anything.  Ultimately I'll go with what we decide is best for the project (e.g. causes the least friction) and honestly don't want to spend time in long discussions about style.  I'd rather adapt (as I already have in other style decisions) and get back to work. :)

So with that in mind, I'll leave it to you. :)

int19h:
I did not observe that event in a single process case - thread creation was the only thing that was consistently present. We might want to revisit this in the future and add an event to pydevd that's more semantically appropriate (since what we really need here is to wait until it's initialized all threads for tracing).

karthiknadig:
Won't the delay between actual thread creation and the thread created event from pydev be a issue here?

int19h:
@karthiknadig, I reproed this problem in VS (hitting pause before process creation is reported). The workaround for now is to basically disable pause until we're fully initialized on ptvsd side - I tested this with both VS and VSC, and they both show the appropriate error message; and VS no longer gets into the weird state where it's pausing.

This should do the trick for now, and later we can add a proper pydevd initialization complete event and wire that up accordingly to avoid the thread delay - although, since requests can come asynchronously at any time, having this check for pause even then would still be a good idea.

PR #188, opened 2018-03-07
Couple of fixes

### LINKED ISSUE #116: Children and watch window ordering
image image

These should be sorted as:
* no leading underscore (`public`)
* single leading underscore (`_private`)
* double leading underscore (`__uninheritable`)
* dunder names (`__file__`)

### LINKED ISSUE #165: Copy details on exception is missing source and stack trace
Visual Studio

```
raise NameError('name')
```

Hit Copy Details in the popup, it will give you this:
```

  Message=NameError('name',)
  Source=
  StackTrace:

```

karthiknadig:
Fixes #165
Fixes #116

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -410,6 +410,40 @@ class ExceptionsManager(object):
         return category[0] == 'Python Exceptions'
 
 
+class VariablesSorter(object):
+    def __init__(self):
+        self.variables = [] # variables that do not begin with underscores
+        self.single_underscore = [] # variables that begin with underscores
+        self.double_underscore = [] # variables that begin with two underscores
+        self.dunder = [] # variables that begin and end with double underscores
+    
+    def append(self, var):
+        var_name = var['name']
+        if var_name.startswith('__'):
+            if var_name.endswith('__'):
+                self.dunder.append(var)
+                print('Apended dunder: %s' % var_name)
+            else:
+                self.double_underscore.append(var)
+                print('Apended double under: %s' % var_name)
+        elif var_name.startswith('_'):
+            self.single_underscore.append(var)
+            print('Apended single under: %s' % var_name)
+        else:
+            self.variables.append(var)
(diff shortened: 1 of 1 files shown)


karthiknadig:
@int19h I used `sort` instead of `sorted` since I am using a list and sort works in-place. Will that work?

ptvsd/wrapper.py
```diff
@@ -737,19 +739,43 @@ def on_variables(self, request, args):
...
+
+        def get_sort_key(o):
+            return o['name']
+
+        variables.sort(key=get_sort_key)
+        single_underscore.sort(key=get_sort_key)
+        double_underscore.sort(key=get_sort_key)
+        dunder.sort(key=get_sort_key)
+
+        variables = variables + single_underscore + double_underscore + dunder
```
DonJayamanne: I'd move all of this into a separate class to track and sort the variables.
If tomorrow we need to sort private variables or member variables, we need to modify this method again, when logic of sorting the list should be outside.

DonJayamanne: This stuff or sorting and tracking variables should be in its own class, something that can be easily unit tested in isolation.

DonJayamanne: I think we'll need unit tests for this sorting rule.

karthiknadig: I will move this to a class before merging.

DonJayamanne: :+1:

ptvsd/wrapper.py
```diff
@@ -739,10 +773,7 @@ def on_variables(self, request, args):
         except AttributeError:
             xvars = []
 
-        variables = [] # variables that do not begin with underscores
-        single_underscore = [] # variables that begin with underscores
-        double_underscore = [] # variables that begin with two underscores
-        dunder = [] # variables that begin and end with double underscores
+        variables = VariablesSorter()
```
DonJayamanne: 👍

PR #189, opened 2018-03-08
Adds evaluateName for variables within containers

### LINKED ISSUE #87: Add watch from QuickWatch window watches wrong expression
Visual Studio:
```
class MyClass:
    a = 12
    text = 'hello'


c_obj = MyClass()
c_obj.a = 15

import os
r = os.urandom(5100111)
print('done')
```

I selected `text` and clicked on Add Watch, and see what it added to Watch window.

image

karthiknadig:
Fixes #87

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -788,6 +788,27 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
                 pyd_child = pyd_var + (var_name,)
                 var['variablesReference'] = self.var_map.to_vscode(
                     pyd_child, autogen=True)
+
+            if len(pyd_var) > 3:
+                # This means the current variable has a parent i.e, it is not a
+                # FRAME variable. These require evaluateName to work in VS 
+                # watch window
+                eval_name = pyd_var[3]
+                for s in pyd_var[4:]:
+                    try:
+                        # In the result XML from pydevd list indexes appear 
+                        # as names. If the name is a number then it is a index.
+                        eval_name += '[' + str(int(s)) + ']'
+                    except:
+                        eval_name += '.' + s 
+
+                try:
+                    eval_name += '[' + str(int(var_name)) + ']'
+                except:
+                    eval_name += '.' + var_name
+
+                var['evaluateName'] = eval_name
+
             variables.append(var)
         self.send_response(request, variables=variables.get_sorted_variables())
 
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -739,15 +739,37 @@ def on_variables(self, request, args):
...
+            if len(pyd_var) > 3:
+                # This means the current variable has a parent i.e, it is not a
+                # FRAME variable. These require evaluateName to work in VS 
+                # watch window
+                eval_name = pyd_var[3]
+                for s in pyd_var[4:]:
+                    try:
+                        # In the result XML from pydevd list indexes appear 
+                        # as names. If the name is a number then it is a index.
+                        eval_name += '[' + str(int(s)) + ']'
```
int19h: Can you check what it does for dicts?

karthiknadig: This will not work for dicts. The `name` returned by pydevd is in a strange format. For example, consider this dict:
```
c= ('green', 'blue')
dict1 = {
    'apple': 'Red',
    0 : 'Zero',
    c: 'colors',
    }
```
This is how the `name`, `type`, and `value` appear to us:

| Name                             | Type  | Value    |
|----------------------------------|-------|----------|
| 'apple' (2200118833872)          | str   | 'Red'    |
|('green', 'blue') (2200118705544) | str   | 'colors' |
| 0 (140728987558560)              | str   | 'Zero'   |
| `__len__`                          | int   | 3        |

ptvsd/wrapper.py
```diff
@@ -750,30 +750,53 @@ def on_variables(self, request, args):
...
+            var = pyd_var_parent + (var_name,)
+            print(var)
+            eval_name = var[3]
+            for s in var[4:]:
+                try:
+                    # check if this is a dictionary key
+                    # Note: this is best effort, keys that are object references
+                    # will not work
+                    key = self.__get_dictionary_key(s)
+                    print(key)
```
int19h: Debug code.

ptvsd/wrapper.py
```diff
@@ -750,30 +750,53 @@ def on_variables(self, request, args):
...
+                    if key is not None:
+                        eval_name += '[{}]'.format(key)
+                    else:
+                        # In the result XML from pydevd list indexes appear 
+                        # as names. If the name is a number then it is a index.
+                        eval_name += '[{}]'.format(int(s))
+                except:
```
int19h: The `try`..`except` is now catching all errors from `__get_dictionary_key` and others. If that's not intentional, I think it should go just around `int(s)` - I think you could just do `key = int(s)` under `try`, and then the actual string formatting is identical for both branches.

karthiknadig: I agree, will reduce this.

PR #212, opened 2018-03-13
Support display in hex

### LINKED ISSUE #198: Hexadecimal mode
We should respect hexadecimal mode for variable values in VS, as described here:

https://github.com/Microsoft/VSDebugAdapterHost/wiki/Stack-Trace-and-Value-Formatting

karthiknadig:
Fixes #198

sample code:
```
lines = [1234, 0, 1, 10, 16, 23, 32]
records = {1: [12,13,14,15,16], 10: [120,130,140,150]}
```
Hex display:
image

Long integers:
image

### DIFF
--- a/ptvsd/safe_repr.py
+++ b/ptvsd/safe_repr.py
@@ -29,7 +29,7 @@ class SafeRepr(object):
         string_types = (str, bytes)
         set_info = (set, '{', '}', False)
         frozenset_info = (frozenset, 'frozenset({', '})', False)
-        int_types = (int)
+        int_types = (int,)
     else:
         string_types = (str, unicode)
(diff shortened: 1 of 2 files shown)


ptvsd/safe_repr.py
```diff
@@ -29,10 +29,12 @@ class SafeRepr(object):
         string_types = (str, bytes)
         set_info = (set, '{', '}', False)
         frozenset_info = (frozenset, 'frozenset({', '})', False)
+        int_types = (int)
```
ericsnowcurrently: a single-item tuple literal must have a trailing comma:

    int_types = (int,)

PR #214, opened 2018-03-15
Enable set setting values for collection items

### LINKED ISSUE #46: Cannot edit collection elements

karthiknadig:
Fixes #46


karthiknadig:
@int19h If the `var_value` contains tabs or if the `var_name` contains tabs(say dictionary key). This will not work.

int19h:
Once we switch to using proper repr for dictionary keys (which means that tabs inside of strings will become `\t`), that should mostly fix that issue with tabs.

PR #215, opened 2018-03-16
Add some initial tests that each use a live pydevd.

ericsnowcurrently:
This required a bit of refactoring in the testing framework to split off the pydevd-specific parts.

### DIFF
--- a/ptvsd/ipcjson.py
+++ b/ptvsd/ipcjson.py
@@ -99,6 +99,9 @@ class SocketIO(object):
             self.__socket.send(content)
         except BrokenPipeError:
             pass
+        except OSError as exc:
+            if exc.errno not in (errno.ENOTCONN, errno.EBADF):
+                raise
 
     def _buffered_read_line_as_ascii(self):
         """Return the next line from the buffer as a string.
--- a/ptvsd/ipcjson.py
+++ b/ptvsd/ipcjson.py
@@ -239,7 +239,7 @@ class IpcChannel(object):
         self.__exit = False
         self.__lock = thread.allocate_lock()
         self.__message = []
-        self.__exit_on_unknown_command = True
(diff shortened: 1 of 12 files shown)


ptvsd/ipcjson.py
```diff
@@ -352,7 +355,9 @@ def on_event(self, msg):
     def on_invalid_request(self, request, args):
         # TODO: docstring
         self.send_response(request, success=False, message='Unknown command')
-        if self.__exit_on_unknown_command:
+        if self._exit_on_unknown_command:
+            # TODO: Shouldn't we let VSC decide how to handle this
```
int19h: I think fast-fail is a reasonable approach. If the other side sends an unknown command, it may well be a request that requires a reply, and will block the IDE until it receives one. But it would be tricky to diagnose if the message is simply ignored.

tests/helpers/pydevd/_live.py
```diff
@@ -0,0 +1,108 @@
...
+class LivePyDevd(protocol.Daemon):
+
+    @classmethod
+    def parse_source(cls, source):
+        kind, sep, name = source.partition(':')
+        if kind == 'file':
+            return name, None, False
+        elif kind == 'module':
+            parts = (name + '.py').split('.')
+            filename = os.path.join(*parts)
```
int19h: Won't this produce something like "foo/bar/py"?

ptvsd/ipcjson.py
ericsnowcurrently: We always send a response message (with success=False).  So the IDE will know that the request failed.

tests/helpers/pydevd/_live.py
ericsnowcurrently: Yeah, that's a typo.  I'll fix it. :)

PR #217, opened 2018-03-16
Enable build and sign on Windows

### LINKED ISSUE #112: Cython speedup extension is not found
This is on Python 3.6 32-bit - the following message is printed when debugging starts:
```
warning: Debugger speedups using cython not found. Run '"C:\Python\3.6-32\python.exe" "C:\Python\3.6-32\lib\site-packages\ptvsd\pydevd\setup_cython.py" build_ext
--inplace' to build.
```
This is despite `.pyd` files being in their usual locations. Probably this is happening because Cython stuff is loaded dynamically, and by the time pydevd tries to load it, we've already removed the `sys.path` entry.

zooba:
This script can be run to generate a semi-pure wheel that includes precompiled extension modules for 3.5 and 3.6, both 32-bit and 64-bit. (Python 3.5 is when Windows started using ABI suffix on extension modules - we can't replicate this as easily for earlier versions.) I already have an internal build definition using this that can also sign all the binaries as needed.

I also fixed #112 by changing how we build the extension modules (i.e. with the correct names). I'm also not quite sure why the `default_return_value` change became necessary, but it was, so I assume it was one of our patches?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include LICENSE
+global-exclude *.py[cdo]
--- a/ptvsd/pydevd/_pydevd_bundle/pydevd_frame.py
+++ b/ptvsd/pydevd/_pydevd_bundle/pydevd_frame.py
@@ -36,7 +36,7 @@ TRACE_PROPERTY = 'pydevd_traceproperty.py'
 get_file_type = DONT_TRACE.get
 
 
-def handle_breakpoint_condition(py_db, info, breakpoint, new_frame):
+def handle_breakpoint_condition(py_db, info, breakpoint, new_frame, default_return_value=None):
     condition = breakpoint.condition
     try:
         return eval(condition, new_frame.f_globals, new_frame.f_locals)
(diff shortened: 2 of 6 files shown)


int19h:
@karthiknadig, can you take a look at `default_return_value` stuff?

karthiknadig:
@zooba `default_return_value` change was done to fix the conditional breakpoints in pydevd. It was this change that did that https://github.com/Microsoft/ptvsd/commit/63e93ec26cf60ea610f7a68be9e7ecf1b572734d

zooba:
Okay, I removed the parameter. Can you confirm that it should be returning `False`/`None` in this case?

karthiknadig:
It should return `False`, that makes the intent of returning there clearer.

zooba:
Done. The `else` block of that function needs looking at too, as it doesn't return anything (and arguably if the condition is broken it should return `True` so that we break, right?).

But file it separately - we need this PR merged before we can build for today's bug bash.

PR #218, opened 2018-03-17
Do eval and exec for repl evaluate requests

karthiknadig:
This change is part of the debug REPL changes needed for VS

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -913,7 +913,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         # Reset hex format since this is per request.
         safe_repr_provider.convert_to_hex = False
         self.send_response(request, **response)
-
+    
     @async_handler
     def on_evaluate(self, request, args):
         # pydevd message format doesn't permit tabs in expressions
@@ -923,11 +923,12 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         pyd_tid, pyd_fid = self.frame_map.to_pydevd(vsc_fid)
 
         safe_repr_provider.convert_to_hex = args.get('format', {}).get('hex', False)
-
+        
         cmd_args = (pyd_tid, pyd_fid, 'LOCAL', expr, '1')
+        msg = '\t'.join(str(s) for s in cmd_args)
         _, _, resp_args = yield self.pydevd_request(
             pydevd_comm.CMD_EVALUATE_EXPRESSION,
-            '\t'.join(str(s) for s in cmd_args))
+            msg)
         xml = untangle.parse(resp_args).xml
         xvar = xml.var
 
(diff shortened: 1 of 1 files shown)


DonJayamanne:
I'll have to check how this works with VSC as well. That because when evaluating expressions in the VSC debug console `context=repl` (My plan was to use `CMD_CONSOLE_EXEC` though that seemed to write to stdout and stderr with results).

Just out of curiosity, what happens if we enter an invalid variable or expression, such as `xyz` or `print(xyz`. Does the error message get printed out, the old debugger used to print the error message (if it was an invalid syntax of invalid variable), this is what happens when `CMD_CONSOLE_EXEC` is used.

karthiknadig:
Yep the error will be printed to the output.

ptvsd/wrapper.py
```diff
@@ -940,6 +941,27 @@ def on_evaluate(self, request, args):
...
+                pydevd_comm.CMD_EXEC_EXPRESSION,
+                msg)
+            try:
+                xml2 = untangle.parse(resp_args).xml
+                xvar2 = xml2.var
+                result_type  = unquote(xvar2['type'])
+                result = unquote(xvar2['value'])
+            except:
+                # if resp_args is not xml then it contains the error traceback
+                result_type  = unquote(xvar['type'])
```
DonJayamanne: If we're sending the traceback, then we need to filter the traceback to hide the internals.
Currently we're sending too much (internal) information back to the user. I would have thought just the error message would suffice. what's the current behaviour in VS. In VSC we just display the error message.

karthiknadig: Actually the comment there is incorrect. When i had initially written this it was returning trace back. but this is what you will see:
image

karthiknadig: The traceback you see in #219 comes from exec itself. We may have to do a bit of trickery to avoid this. I tried using `sys.displayhook` to capture this info but that did not help. @int19h has a few ides on how to fix that.

DonJayamanne: I had a similar idea (when using `CMD_CONSOLE_EXEC`) but that's where I got stuck this evening. Gets complicated when you evaluate a function that starts a thread which prints stuff (sounds too far fetched). 
Anyways, would like to see how it is that the stack trace isn't printed for your code, but only for VSC. Is it because you're ignoring what comes from `output` event?

karthiknadig: The traceback is printed in the console output, in the REPL window we print what is returned via the protocol. the screen sot was from the REPL window.

DonJayamanne: Interesting, will need to discuss this in further detail. (I thought the console===REPL).

int19h: In REPL scenarios, everything that normally goes into stdout/err is supposed to go to REPL - it just isn't wired yet. But once we do, we will have that problem.

DonJayamanne: * Out of curiosity whey aren't we using `CMD_CONSOLE_EXEC` command, after all the value for `context` is `repl` 

* >it just isn't wired yet. But once we do, we will have that problem.

    * I think I understand, VSC is getting the `stdout/stderr` messages via the protocol as VSC is passing in the value `true` for the setting [redirectOutput](https://github.com/Microsoft/ptvsd/blob/master/ptvsd/wrapper.py#L644). I'm assuming VS doesn't use this.

karthiknadig: I will give CMd_CONSOLE_EXEC a try in the morning, and update this PR if it seems better.

karthiknadig: CMD_CONSOLE_EXEC does not work the way we want it. For the following REPL statements, this is the result we get:

**In REPL window**:
```
>>> random.random()
False
>>> import random
False
>>> random.random()
False
>>> 
```

**In Console Window**:
```
  File "c:\git\ptvsd\ptvsd\pydevd\_pydevd_bundle\pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
NameError: name 'random' is not defined
<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>

<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>

0.9210117288960727
<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>
```


**XML response received from pydevd:**
```
>>> random.random()
<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>

>>> import random
<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>

>>> random.random()
<xml><var name="" type="bool" qualifier="builtins" value="False" />
</xml>
```

As you can see it always return false.

DonJayamanne: All good at my end 👍

PR #230, opened 2018-03-20
Ignore errors when closing the socket

### LINKED ISSUE #228: Running Python code to completion randomly fails with error
**Very difficult to replicate**

Python Code:
```
import sys
sys.stdout.write("OUT")
sys.stderr.write("EXs")
```

Traceoutput:
```
OUTEXError in atexit._run_exitfuncs:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 1334, in <lambda>
    atexit.register(lambda: exit_handler(proc, server_thread))
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 1293, in exit_handler
    proc.close()
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 535, in close
    self.socket.shutdown(socket.SHUT_RDWR)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 57] Socket is not connected
Error in sys.exitfunc:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 1334, in <lambda>
    atexit.register(lambda: exit_handler(proc, server_thread))
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 1293, in exit_handler
    proc.close()
  File "/Users/donjayamanne/.vscode/extensions/ms-python.python-2018.3.0-alpha/pythonFiles/experimental/ptvsd/ptvsd/wrapper.py", line 535, in close
    self.socket.shutdown(socket.SHUT_RDWR)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 57] Socket is not connected
```

DonJayamanne:
Fixes #228 
I believe its safe to just ignore all errors when shutting down & releasing resources.
At the moment, the error is `socket.error`.
Its very difficult to replicate this issue (was able to replicate it consistently earlier today).

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -558,7 +558,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
             try:
                 self.socket.shutdown(socket.SHUT_RDWR)
                 self.socket.close()
-            except:
+            except Exception:
                 pass
 
     def pydevd_notify(self, cmd_id, args):
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -555,8 +555,11 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         self.event_loop_thread.join(WAIT_FOR_THREAD_FINISH_TIMEOUT)
 
         if self.socket:
-            self.socket.shutdown(socket.SHUT_RDWR)
-            self.socket.close()
+            try:
+                self.socket.shutdown(socket.SHUT_RDWR)
+                self.socket.close()
+            except:
+                pass
 
     def pydevd_notify(self, cmd_id, args):
         # TODO: docstring


int19h:
Yep, ignoring errors during cleanup is the best we can do here.

ptvsd/wrapper.py
```diff
@@ -555,8 +555,11 @@ def close(self):
         self.event_loop_thread.join(WAIT_FOR_THREAD_FINISH_TIMEOUT)
 
         if self.socket:
-            self.socket.shutdown(socket.SHUT_RDWR)
-            self.socket.close()
+            try:
+                self.socket.shutdown(socket.SHUT_RDWR)
+                self.socket.close()
+            except:
+                pass
```
ericsnowcurrently: Knowing that closing the socket errored-out would be useful, particularly as the ptvsd code is still undergoing so many changes.  Regardless, masking errors is typically a recipe for pain.  So it may be worth logging something here.  At the very least we should emit a warning (or even just print a message).

DonJayamanne: > So it may be worth logging something here. At the very least we should emit a warning (or even just print a message). 

We are printing these, and that's what I'm trying to avoid. We shouldn't be displaying this to the user. Else it looks as though something has gone wrong in the user code or debugger has crashed (that was my initial assumption).

Is it possible that VSC closes the socket before PTVSD gets to this block of code.

> So it may be worth logging something here

Would have, but we don't have any logging mechanism in PTVSD yet.

PR #232, opened 2018-03-20
Add support for setExpression

karthiknadig:


ptvsd/wrapper.py
```diff
@@ -1012,6 +1012,36 @@ def on_evaluate(self, request, args):
...
+        lhs_expr = args.get('expression')
+        rhs_expr = args.get('value')
+        expr = '%s = %s' % (lhs_expr, rhs_expr)
+
+        # pydevd message format doesn't permit tabs in expressions
+        expr = expr.replace('\t', ' ')
+
+        cmd_args = (pyd_tid, pyd_fid, 'LOCAL', expr, '1')
+        msg = '\t'.join(str(s) for s in cmd_args)
+        _, _, _ = yield self.pydevd_request(
```
int19h: Because of the `yield` here, there's a potential re-entrancy point, which might play nasty tricks with setting/changing the hex setting, if whatever happens during the yield also tries to do the same. It might be better to just construct a new `SafeRepr` instance every time, since it's pretty cheap (the only thing that does is set the parameters).

int19h: (Might need to take a look at other places where hex is changed, too... this didn't occur to me before.)

karthiknadig: @int19h Are you suggesting that we do something like this? for each call to getVariable, setVariable, etc?
```
hex_mode = args.get('format',{}).get('hex', False)
str_handlers.insert(0, SafeReprPresentationProvider(hex_mode))
```

int19h: I missed the part where the attribute that we're changing here is on the representation provider, rather than safe repr object itself. The above doesn't really help, since the provider is still registered globally, and so they'll interfere if there's two requests to compute in parallel.

It looks like we might need some kind of locking for the presentation provider, so that message handlers that exercise it (and flip the hex bit on/off) can lock it for the duration of the use, and avoid re-entrancy.

karthiknadig: I am going to remove hex for this change. It is actually not needed with setExpression. Since we don't really return anything.

int19h: :+1:

ptvsd/wrapper.py
```diff
@@ -1012,6 +1012,31 @@ def on_evaluate(self, request, args):
...
+    @async_handler
+    def on_setExpression(self, request, args):
+        # TODO: docstring
+
+        vsc_fid = int(args['frameId'])
+        pyd_tid, pyd_fid = self.frame_map.to_pydevd(vsc_fid)
+
+        lhs_expr = args.get('expression')
+        rhs_expr = args.get('value')
+        expr = '%s = %s' % (lhs_expr, rhs_expr)
```
int19h: I think this should be:
```py
%s = (%s)
```
To prevent accidental or creative abuses, e.g. if someone stuffs a semicolon into `rhs_expr`.

karthiknadig:
The Travis CI failures do not seem to be associated with this change.

PR #240, opened 2018-03-22
Disable the schema check.

ericsnowcurrently:


DonJayamanne:
Why are we disabling this? I thought it was necessary.

ericsnowcurrently:
On Travis we're getting frequent 403 errors.  Presumably this is because the GitHub API has a limit of 500 anonymous requests per hour and checking the schema makes such a request.

DonJayamanne:
Can't we enable the test and skip it if we get a 403, that way the tests run regardless but fail ONLY when schema doesn't match.

Issue #64, opened 2018-02-13
Call stack shows unqualified names


zooba:
![image](https://user-images.githubusercontent.com/1693688/36126783-e734ffd6-100e-11e8-8eff-11ef067cf1f8.png)

Fully qualified names would be ideal, or at least including the module name.

Issue #22, opened 2018-01-24
Generate ModuleEvent


karthiknadig:
Generate moduleEvent on import, reload, and del.

Issue #21, opened 2018-01-24
Handle ModulesRequest


karthiknadig:
This request is received by debugger for attach only. Launch depends on module events for tracking modules.

PR #242, opened 2018-03-22
Modules

karthiknadig:
Using this PR to discus the right format and any performance implications of searching for modules from stack traces.

In this PR:
1. Handles modules request #21
2. Generate modules events based on modules seen in stack frames #22
3. *Does NOT report unloaded modules
4. Formats stack frames based on modules #64 (partial)

This is with **modules** and **lines** on:
image

This is with only **lines** on:
image

This is with nothing on:
image

Pending:

- [x] Update to stack trace tests to include formatting
- [x] New tests for modules request
- [x] Fixes for broken tests


zooba:
Can we format with full dotted names like you'd import them? I don’t think the length problem will be worse than in any other language, and we ought to be consistent as much as we can.

ptvsd/wrapper.py
```diff
@@ -508,6 +508,68 @@ def get_sort_key(o):
...
+            if path and mpath == path:
+                with self._lock:
+                    module_id = self._next_id
+                    self._next_id += 1
+
+                    self.path_to_module_id[module_path] = module_id
+
+                    module = {
+                        'id': module_id,
+                        'name': value.__name__,
```
int19h: Let's use `__qualname__` here if it's available, and fall back to `__name__` otherwise.

ptvsd/wrapper.py
```diff
@@ -508,6 +508,68 @@ def get_sort_key(o):
...
+
+                    self.module_id_to_details[module_id] = module
+
+                    self.proc.send_event('module', reason='new', module=module)
+                    return module
+
+        return None
+
+    def get_all(self):
+        for _, value in self.module_id_to_details.items():
```
int19h: You can use `values()` instead of `items()` to iterate just through those.

ptvsd/wrapper.py
int19h: `__qualname__` is also interesting when it's available (and can then be used in `__format_frame_name`).

karthiknadig:
@zooba, @int19h is this better:
```
__main__ : 60
_pydev_imps._pydev_execfile.execfile : 25
pydevd.run : 1035
pydevd.main : 1628
ptvsd.debugger._run : 54
ptvsd.debugger._run_file : 36
ptvsd.debugger.debug : 24
pydevd : 96  <---- this one is wrong but that is because python thinks ptvsd_launcher.py is pydevd module
```

We currently don't have a good way to get fully qualified function names. We could use custom messages to talk to analyzer for more detailed function info (including parameter type, name, and values).

int19h:
For function names, can we reuse `QualifiedVersionNameWalker` on VS side to improve the output? We could do it before, because we were providing `AD7StackFrame` ourselves. Not sure if it's possible with the adapter.

(Not for this PR, just an idea.)

int19h:
@ericsnowcurrently, can you take a look at the tests?

ericsnowcurrently:
Yeah, but it will probably be Monday.  I'm still at the DevDiv bootcamp.

int19h:
I'd like to get this in today so that it can go into the VS bundle - this is one of those things that users will notice right away. Let's go ahead with the merge, but feel free to leave any comments on how the tests can be improved, and Karthik can take care of that next week.

ericsnowcurrently:
will do

PR #247, opened 2018-03-24
Enable Django for VS

karthiknadig:
In this PR:
1. Parse launch options sent by VS
2. Enable Django and Jinja2 template debugging for VS

Notes:
For VS flask project we have to enable breakpoints in flask project, and send `FLASK_DEBUG=True` as debug option.

Pending:
- [x] Add tests to include launch options as sent by VS

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -602,6 +602,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         self.modules_mgr = ModulesManager(self)
         self.disconnect_request = None
         self.launch_arguments = None
+        self.launch_options = {}
         self.disconnect_request_event = threading.Event()
         pydevd._vscprocessor = self
         self._closed = False
@@ -776,6 +777,14 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
             name:'friendly name for debug config',
             // Custom attributes supported by PTVSD.
             redirectOutput:true|false,
+            options:'VS options semicolon separated key=value pairs'
+                    WAIT_ON_ABNORMAL_EXIT=True|False
+                    WAIT_ON_NORMAL_EXIT=True|False
+                    REDIRECT_OUTPUT=True|False
+                    VERSION=string
+                    INTERPRETER_OPTIONS=string
+                    WEB_BROWSER_URL=string url
+                    DJANGO_DEBUG=True|False
         }
         """  # noqa
         if self.launch_arguments is None:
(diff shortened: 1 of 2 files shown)


ptvsd/wrapper.py
```diff
@@ -602,6 +602,7 @@ def __init__(self, socket, pydevd, logfile=None, killonclose=True):
         self.modules_mgr = ModulesManager(self)
         self.disconnect_request = None
         self.launch_arguments = None
+        self.launch_options = {}
```
DonJayamanne: I'd check this to debug options. For instance when attaching to Django application we'll need to option DJANGO_DEBUG=True in the attach request for brains to work.

ptvsd/wrapper.py
```diff
@@ -790,6 +791,30 @@ def process_launch_arguments(self):
...
+    def __parse_launch_option_value(self, value):
+        if value == 'True' or value == 'False':
+            return bool(value)
+        return unquote(value)
+
+    def __parse_launch_options(self, launch_options):
+        """VS options semicolon separated key=value pairs
+            WAIT_ON_ABNORMAL_EXIT=True|False
+            WAIT_ON_NORMAL_EXIT=True|False
+            REDIRECT_OUTPUT=True|False
```
DonJayamanne: You're not handling REDIRECT_OUTPUT in launch request.

ptvsd/wrapper.py
```diff
@@ -1252,30 +1279,36 @@ def on_setBreakpoints(self, request, args):
         self.path_casing.track_file_path_case(path)
         src_bps = args.get('breakpoints', [])
 
+        bp_type = 'python-line'
+        if not path.lower().endswith('.py'):
+            if self.launch_options.get('DJANGO_DEBUG', False):
```
DonJayamanne: This doesn't apply to launch debugging only, attaching too would need this check, hence the suggestion to change property name

tests/highlevel/test_messages.py
```diff
@@ -1059,6 +1057,80 @@ def test_multiple_files(self):
                 '2\tpython-line\teggs.py\t17\tNone\tNone\tNone'),
         ])
 
+    def test_vs_django(self):
+        with self.launched(args={'options': 'DJANGO_DEBUG=True'}):
+            self.send_request(
```
DonJayamanne: We will also need tests for remote debugging (attach request).

ptvsd/wrapper.py
```diff
@@ -790,6 +791,30 @@ def process_launch_arguments(self):
             redirect_output = ''
         self.pydevd_request(pydevd_comm.CMD_REDIRECT_OUTPUT, redirect_output)
 
+    def __parse_launch_option_value(self, value):
+        if value == 'True' or value == 'False':
```
DonJayamanne: We need to add support for attaching as well. Or we could make that a separate PR.

ptvsd/wrapper.py
```diff
@@ -790,6 +791,30 @@ def process_launch_arguments(self):
...
+            return bool(value)
+        return unquote(value)
+
+    def __parse_launch_options(self, launch_options):
+        """VS options semicolon separated key=value pairs
```
DonJayamanne: I'd remove this message. There's no reason VSC or another debugger can't use this format. If it did then one would assume this was being used only by VS.

DonJayamanne:
Lets keep this PR simple, ignore remote debugging (attach), Please create an issue for that.

ptvsd/wrapper.py
karthiknadig: I want to keep this about DJANGO. created a bug to handle redirect separately: https://github.com/Microsoft/ptvsd/issues/250

ptvsd/wrapper.py
karthiknadig: I will handle attach separately: https://github.com/Microsoft/ptvsd/issues/248

tests/highlevel/test_messages.py
karthiknadig: Will do this as a part of attach: https://github.com/Microsoft/ptvsd/issues/248

ptvsd/wrapper.py
DonJayamanne: Hmm, bug or missing feature 😄

Issue #250, opened 2018-03-24
Handle REDIRECT_OUTPUT option from VS


karthiknadig:

DonJayamanne:
Renamed title as planning on using this same `options` setting for VSC as well.
Lets stick to this same format, else it'll become a nightmare when dealing with 2 ways for attach and launch requests.

Issue #195, opened 2018-03-12
Wait on exit


int19h:
When debuggee exits, there should be an option to:
- wait ("Press any key to continue...") if it terminated normally
- wait if it terminated with an error

DonJayamanne:
Fyi, this isn't necessary for vs code, hence this needs to be a configurable behavior.

int19h:
It is a configurable behavior for VS as well for both of these options. And it only applies to F5 scenarios, not attach, so we're just going to reuse the existing debug option for it (passed as command line argument).

karthiknadig:
Should we do this via InitializeRequest? The adapter host allows us to add custom attributes to Initialize request. Let me know if we should do this approach vs send this via command line.

int19h:
VS already does it via command line on its side (when it invokes ptvsd_launcher.py), so the easiest thing is to just enable support for those two flags (`WaitOnNormalExit` and `WaitOnAbnormalExit`) on ptvsd side.

PR #260, opened 2018-03-26
Support few more debug options from VS

karthiknadig:
In this PR:
1. Wait on normal exit and abnormal exit #195
2. Redirect output is now supported #250

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -50,6 +50,20 @@ WAIT_FOR_DISCONNECT_REQUEST_TIMEOUT = 2
 WAIT_FOR_THREAD_FINISH_TIMEOUT = 1
 
 
+def _wait_on_exit():
+    if sys.__stdout__ is not None:
+        try:
+            import msvcrt
+        except ImportError:
+            sys.__stdout__.write('Press Enter to continue . . . ')
+            sys.__stdout__.flush()
+            sys.__stdin__.read(1)
+        else:
+            sys.__stdout__.write('Press any key to continue . . . ')
+            sys.__stdout__.flush()
+            msvcrt.getch()
+
+
 class SafeReprPresentationProvider(pydevd_extapi.StrPresentationProvider):
     """
     Computes string representation of Python values by delegating them
@@ -361,7 +375,7 @@ class ExceptionsManager(object):
                                         'python-{}'.format(exception))
             self.exceptions = {}
 
-    def __find_exception(self, name):
+    def _find_exception(self, name):
         if name in self.exceptions:
             return name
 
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -613,7 +613,21 @@ def __init__(self, socket, pydevd, logfile=None, killonclose=True):
         self.event_loop_thread.daemon = True
         self.event_loop_thread.start()
 
+    def __input(self, message):
+        try:
+            raw_input(message)
```
int19h: This will only work in Python 3.x, and even then it won't do what the message says ("press *any* key to continue"). It's not always easily doable, depending on the platform, but we can do a best effort. Here's the old code that does it - let's just take that:

https://github.com/Microsoft/PTVS/blob/master/Python/Product/PythonTools/ptvsd/debugger.py#L2535

ptvsd/wrapper.py
```diff
@@ -785,19 +799,20 @@ def process_launch_arguments(self):
...
+        if self.launch_arguments.get('redirectOutput', False) \
+            or self.debug_options.get('REDIRECT_OUTPUT', False):
             redirect_output = 'STDOUT\tSTDERR'
         else:
             redirect_output = ''
         self.pydevd_request(pydevd_comm.CMD_REDIRECT_OUTPUT, redirect_output)
 
-    def __parse_launch_option_value(self, value):
+    def __parse_debug_option_value(self, value):
         if value == 'True' or value == 'False':
```
int19h: I didn't notice this before, but that seems dangerous - what if we have a string option that just happens to be True or False? I think it's better to make a dict that maps options to their types/parsers, like so:
```py
DEBUG_OPTIONS = dict(REDIRECT_OUTPUT=bool, INTERPRETER_OPTIONS=unquote, ...)
```
Then when you're parsing the options, you can iterate over that dict to see which ones you care about, and invoke the function that's needed for that particular option.

ptvsd/wrapper.py
```diff
@@ -613,7 +613,21 @@ def __init__(self, socket, pydevd, logfile=None, killonclose=True):
         self.event_loop_thread.daemon = True
         self.event_loop_thread.start()
 
+    def __input(self, message):
```
ericsnowcurrently: To facilitate testing this, I'd recommend making the "input" function an attribute of VSCodeMessageProcessor:

    def __init__(self, ..., inputfunc=None, ...):
        ...
        try:
            self.inputfunc = raw_input
        except NameError:
            self.inputfunc = input
        ...

Then in _handle_exit() just call self.inputfunc().

Then for the test machinery you'll need to pass an appropriate inputfunc.  Look for any place we pass `killonclose=False`.  There should be a couple of spots, both under tests/helpers/pydevd/.

Also, FYI, using "__"-prefixed names is discouraged in Python.  A single underscore should be sufficient.

karthiknadig:
Specific tests for the options to follow in a separate PR. Checking this in to unblock dependent PR.

PR #262, opened 2018-03-26
Add support for VSC style debug options

### LINKED ISSUE #245: Handle debugOptions sent by VS Code
Change to ensure we have a consistent approach to sending custom info to PTVSD from VS and VSC

DonJayamanne:
Fixes #245

Waiting for https://github.com/Microsoft/ptvsd/pull/260 to be completed (some overlapping changes).

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -798,11 +798,10 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         """
         Process the launch arguments to configure the debugger.
         """  # noqa
-        if self.launch_options.get('FIX_FILE_PATH_CASE', False):
+        if self.debug_options.get('FIX_FILE_PATH_CASE', False):
             self.path_casing.enable()
 
-        if self.launch_arguments.get('redirectOutput', False) \
-            or self.debug_options.get('REDIRECT_OUTPUT', False):
+        if self.debug_options.get('REDIRECT_OUTPUT', False):
             redirect_output = 'STDOUT\tSTDERR'
         else:
             redirect_output = ''
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -798,9 +798,6 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         """
         Process the launch arguments to configure the debugger.
         """  # noqa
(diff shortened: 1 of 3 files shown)


int19h:
Would it be better to go in the other direction - i.e. use JSON as the primary format, and the semicolon command-line thingy as a fallback? This way, the JSON version can have more and fancier properties if you ever want to add fine grained behavior in VSC, without changing the part that parses the custom format.

int19h:
And I'm thinking you could actually convert that custom format to JSON with a few simple string substitutions: replace all `=` with `:"`, all `;` with `",`, and put `{}` around it. Then the only thing that needs to be done is converting those strings to bools, in the same way that other PR does right now.

DonJayamanne:
@int19h 
Do you mean JSON properties coming into PTVSD?
as follows:
```json
```json
{ 
    "Django":true, 
    "Flask": true, 
    "DebugStdLib": true
}
``
```
as opposed to 
```json
{ 
    "debugOptions": ["Django", "Flask", "DebugStdLib"]
}
``

**Comment by int19h** [comment]
I think we should support both, but the array is simply a shorthand form of using `{... : true, ... : true, ...}`. We can do that conversion at the very beginning, and then just deal with the dict everywhere down the line. So your example above would be the same as:
```json
"debugOptions": {"Django": true, "Flask": true, "DebugStdLib": true}
```
This way we can add non-bool options later too, and VSC will already be able to handle them (it just passes whatever user specifies in launch.json as is, right?).

int19h:
Although looking at the schema, I see what you mean - those properties should really just go into the launch/attach request itself, it doesn't need to nest under `debugOptions`. I suppose that would be more convenient from user perspective?

DonJayamanne:
> properties should really just go into the launch/attach request itself, it doesn't need to nest under debugOptions.

Unfortunately that's how things are in VSC today. I don't think we'll change that for the new debugger. I'll have a chat with Brett. Worst case I can just massage the data and parse debug options sand add the individual properties before sending to ptvsd. Though I don't see the point if vs isn't going to use that. I.e. VSC and vs are using their own formats!

int19h:
Let's go ahead with this then, and revisit it later.

DonJayamanne:
@int19h Please could you cast your eyes on this once again, I added Jinja2 option for jinja template debugging (didn't want to re-use flask, as that feels wrong)
Added some tests as well.

PR #263, opened 2018-03-26
Fix #193: Capture usage of the experimental debugger

### LINKED ISSUE #193: Capture usage of the experimental debugger
This can be done via an `OutputEvent` with a category of `telemetry` .
https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json#L278

This is something that would be useful to both PTVS as well as VS Code.

int19h:
Provide a telemetry message incorporating ptvsd version in response to initialization request.

Add a custom message for more detailed information about ptvsd and its environment.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -7,6 +7,7 @@ from __future__ import print_function, absolute_import
 import atexit
 import contextlib
 import errno
+import getpass
 import os
 import platform
 import re
@@ -629,6 +630,12 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         self.event_loop_thread.daemon = True
         self.event_loop_thread.start()
         self.wait_on_exit_func = waitonexitfunc
+
+        self.send_event(
+            'output',
+            category='telemetry',
+            output='ptvsd',
+            data={'version': __version__})
 
     def _handle_exit(self):
         wait_on_normal_exit = self.debug_options.get(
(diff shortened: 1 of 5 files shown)


tests/helpers/socket.py
```diff
@@ -57,7 +57,10 @@ def connect():
 
 def close(sock):
     """Shutdown and close the socket."""
-    sock.shutdown(socket.SHUT_RDWR)
+    try:
+        sock.shutdown(socket.SHUT_RDWR)
+    except Exception:
```
int19h: It's basically impossible to run most tests that use live sockets on Win32 without this. I think it might be because shutting down a closed socket on other platforms isn't considered an error?

int19h:
I can't reproduce the CI test failures locally, and I believe it to be the same root cause as the other intermittent issues with non-matching events that we've been seeing lately.

ptvsd/wrapper.py
```diff
@@ -1350,6 +1357,52 @@ def on_exceptionInfo(self, request, args):
                      'source': exc.source},
         )
 
+    # Custom ptvsd message
+    def on_ptvsd_systemInfo(self, request, args):
```
karthiknadig: Should we expose this feature via capabilities `supportsPtvsdSystemInfo`? Also, are we making use of this in VS?

int19h: Not yet, but I was hoping that Attach to Process could use some of these :)

The problem with capabilities is that, so far as I can see, the schema doesn't say that we can add our things there? And in capabilities, it would be observable by everyone, not just our bits on the IDE sides (whereas with the new message type, you only see it if you issue the request in the first place). So I'm not sure if it's safe to put it there.

PR #266, opened 2018-03-27
Add explicit CLI handling for ptvsd.

ericsnowcurrently:
This change is intended to accommodate future work (adding more CLI options where using the launch config is insufficient).  It also allows us to define a consistent and easy-to-discover CLI, independent of PyDevd.

To be backward compatible, the code also supports the PyDevd CLI (for now).

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -1783,6 +1783,13 @@ def start_client(host, port, addhandlers=True):
     return pydevd
 
 
-# These are the functions pydevd invokes to get a socket to the client.
-pydevd_comm.start_server = start_server
-pydevd_comm.start_client = start_client
+def install(start_server=start_server, start_client=start_client):
+    """Configure pydevd to use our wrapper."""
+    # These are the functions pydevd invokes to get a socket to the client.
+    pydevd_comm.start_server = start_server
+    pydevd_comm.start_client = start_client
+
+    # Force a fresh pydevd.
+    sys.modules.pop('pydevd', None)
+    import pydevd  # noqa
+    return pydevd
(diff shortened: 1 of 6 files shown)


ptvsd/wrapper.py
```diff
@@ -1612,6 +1612,13 @@ def start_client(host, port, addhandlers=True):
     return pydevd
 
 
-# These are the functions pydevd invokes to get a socket to the client.
-pydevd_comm.start_server = start_server
-pydevd_comm.start_client = start_client
+def install(start_server=start_server, start_client=start_client):
```
int19h: This neuters the assert that we have in `pydevd/__init__.py` about how pydevd must not be imported before ptvsd - now it can be imported before, but if you don't call install, the detours won't work.

ericsnowcurrently: Unlike the first time I tried this, this function now forces a re-import of the "pydevd" module.  Alternately, we could manually replace `pydevd.start_server` and `pydevd.start_client`.

Also, where is `pydevd/__init__.py`?

int19h: Sorry, that was meant to be `ptvsd/__init__.py`.

What happens if some ptvsd code accidentally imports pydevd before we so the refresh below (e.g. to use some constant from it). Do we end up with two copies of that module hanging around in memory, but no harmful effects otherwise?

ericsnowcurrently: I've updated the code to monkeypatch the start functions rather than re-importing the pydevd module.  That fixes issues where pydevd was already imported.  I've also added some comments on why we have to jump through these hoops.

int19h:
Ohh, now I see what's really going on there. Thanks for figuring this out!

PR #278, opened 2018-03-28
Add package information.

int19h:


setup.py
```diff
@@ -44,13 +44,18 @@ def get_pydevd_package_data():
 
 
 PACKAGE_DATA = {
-    'ptvsd': list(get_pydevd_package_data()) + ['ThirdPartyNotices.txt'],
+    'ptvsd': (list(get_pydevd_package_data()) +
+        ['ThirdPartyNotices.txt'] +
+        ['dummy.txt'] if sys.version_info < (3,) else []
```
int19h: This is a hack to avoid building wheels that are identical for py2 and py3 (because PyPI won't let us upload them then).

PR #281, opened 2018-03-29
Add version info event

karthiknadig:
This is needed for version check in VS

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -818,8 +818,6 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
             ],
         )
         self.send_event('initialized')
-        if args.get('supportsVersionInfo', False):
-            self.send_event('versionInfo', version=__version__)
 
     @async_handler
     def on_configurationDone(self, request, args):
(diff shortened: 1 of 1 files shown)


int19h:
https://github.com/Microsoft/ptvsd/blob/master/ptvsd/wrapper.py#L654

int19h:
Ah, I see. Then maybe just add `supportsPtvsdSystemInfo` for the existing custom request that I added for attach metadata?

https://github.com/Microsoft/ptvsd/blob/master/ptvsd/wrapper.py#L1439

DonJayamanne:
If we are after information, then that implies a question being asked, hence the need for a request/response.

int19h:
Yep, and we already have that request/response - remember I added that stuff when I was doing telemetry for VSC?

DonJayamanne:
If we do then why add this event?
> self.send_event('versionInfo', version=__version_

karthiknadig:
I can switch this to request/response using the `ptvsd_systemInfo`. But we will have to do this blindly from VS. since currently there is no way to check the `capabilities` response, however, I have asked for that feature in adapter host. I could inspect the stream for this info if needed. When that feature comes in should be able read `supportsPtvsdSystemInfo` and make the request without inspecting the stream.

DonJayamanne:
> I could inspect the stream for this info if needed. When that feature comes in should be able read supportsPtvsdSystemInfo and make the request without inspecting the stream.

Oh, so you're trying to avoid having to inspect the stream. Oh well, i guess this is a much simpler enough.

int19h:
I think the easiest way to handle this would be to use the existing request on ptvsd side, and on VS side, just issue it unconditionally. If ptvsd is too old to handle it, it will just respond with an error, and then VS can catch that and treat it as "too old".

int19h:
https://github.com/Microsoft/ptvsd/blob/master/ptvsd/wrapper.py#L1460 ?

karthiknadig:
The systemInfo request contains user specif info and I wanted to avoid getting that. After offline discussion, We decided to remove the username field from system info. I will do that in a separate PR.

PR #286, opened 2018-03-30
Minor fixes to system info

karthiknadig:
* The version formatted to package style (2.7.14final0). Previously we were returning int string mixed list.
* Removed the user name part from system info

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -7,7 +7,6 @@ from __future__ import print_function, absolute_import
 import atexit
 import contextlib
 import errno
-import getpass
 import io
 import os
 import platform
@@ -1441,11 +1440,6 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
             pid = os.getpid()
         except AttributeError:
             pid = None
-
-        try:
-            username = getpass.getuser()
-        except AttributeError:
-            username = None
 
         try:
             impl_desc = platform.python_implementation()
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -1455,15 +1455,33 @@ def on_ptvsd_systemInfo(self, request, args):
...
+        except AttributeError: 
+            impl_version =  version_str(sys.version_info)
+
         sys_info = {
             'ptvsd': {
                 'version': __version__,
             },
             'python': {
-                'version': list(sys.version_info),
+                'version': version_str(sys.version_info),
```
int19h: The reason why it was an array is to make version checks easier on the receiving side (since you can just take the major or minor version and ignore the rest). Up to you whether that makes more sense or not.

ptvsd/wrapper.py
```diff
@@ -1455,15 +1455,33 @@ def on_ptvsd_systemInfo(self, request, args):
...
+                v.major,
+                v.minor,
+                v.micro,
+                v.releaselevel,
+                v.serial)
+
+        try:
+            impl_name = sys.implementation.name
+        except AttributeError:
+            impl_name = 'Python'
```
int19h: Can you change this to return None on exception, like and the same for `impl_desc` above and `impl_version` below? Basically so that the client knows whether this information was successfully obtained, or not. Falling back to "Python" is old debugger code, where it was done for ease of implementation mostly; but I think it's best for debugger to just report values as it sees them, and for the IDE to do whatever formatting and missing value substitution that's necessary to represent that data to the user.

ptvsd/wrapper.py
karthiknadig: On the other side i am using PackageVersion Parser in VS. Since we have to do that anyway for ptvsd version.

karthiknadig: Using PackageVersion made it easier to compare versions.

ptvsd/wrapper.py
karthiknadig: will do

PR #287, opened 2018-03-30
Add ability to attach, detach and attach once again

DonJayamanne:
@int19h @ericsnowcurrently @karthiknadig 
**The code is far from complete.**, but would like some early feedback on the solution. 
With the changes I've made I've managed to get re-attach working, at least on VSC.

Pending - Fix remainder of the code and what ever tests fail

Summary:
* Create new threading to check for connections
* Change how callbacks are passed into pydevd

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -249,7 +249,7 @@ class Observable(object):
             observer.notify(*args, **kwargs)
 
     @property
-    def observers_count(self):
+    def observer_count(self):
         return len(self._observers)
 
 
@@ -335,7 +335,7 @@ class PydevdSocket(Observable):
             return b''
         data = os.read(pipe_r, count)
         #self.log.write('>>>[' + data.decode('utf8') + ']\n\n')
-        # self.log.flush()
+        #self.log.flush()
         return data
 
     def recv_into(self, buf):
@@ -365,7 +365,7 @@ class PydevdSocket(Observable):
         result = len(data)
         data = self._decode_and_unquote(data)
         #self.log.write('<<<[' + data + ']\n\n')
-        # self.log.flush()
+        #self.log.flush()
         cmd_id, seq, args = data.split('\t', 2)
         cmd_id = int(cmd_id)
         seq = int(seq)
@@ -560,7 +560,7 @@ class VariablesSorter(object):
         self.single_underscore.sort(key=get_sort_key)
         self.double_underscore.sort(key=get_sort_key)
         self.dunder.sort(key=get_sort_key)
-        # print('sorted')
+        #print('sorted')
         return self.variables + self.single_underscore + self.double_underscore + self.dunder  # noqa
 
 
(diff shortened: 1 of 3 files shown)


ptvsd/wrapper.py
```diff
@@ -45,9 +45,10 @@
 __version__ = "4.0.0a5"
 
 
-#def ipcjson_trace(s):
-#    print(s)
-#ipcjson._TRACE = ipcjson_trace
+def ipcjson_trace(s):
+    with open('/Users/donjayamanne/Desktop/Development/vscode/my_forked_ptvsd/ptvsd/log.log', 'a') as file:
```
DonJayamanne: Temporary debugging code 😃

ptvsd/wrapper.py
```diff
@@ -1674,23 +1695,22 @@ def _new_sock():
...
-
-    server_thread = threading.Thread(target=proc.process_messages,
-                                     name=name)
+    server_thread = threading.Thread(target=proc.process_messages, name=name)
     server_thread.daemon = True
     server_thread.start()
 
     if addhandlers:
         _add_atexit_handler(proc, server_thread)
         _set_signal_handlers(proc)
```
DonJayamanne: I'm going to have to modify `_set_signal_handlers` to close all VS Code message processors.

ptvsd/wrapper.py
```diff
@@ -1726,6 +1746,14 @@ def start_server(port, addhandlers=True):
     server = _create_server(port)
     client, _ = server.accept()
     pydevd = _start(client, server)
+    def _wait_for_more():
+        while True:
+            client, _ = server.accept()
+            _add_socket_handler(client, pydevd, 'ptvsd.Server', killonclose=True, addhandlers=False)
+
+    connection_thread = threading.Thread(target=_wait_for_more, name='ptvsd_client_connection')
```
DonJayamanne: New thread that'll ensure we support more than one client.
@int19h When speaking to Karthik the question about limiting multiple concurrent client connections came up.
Should we have a limit on the number of concurrent connections, to just one?

int19h: Yes. We never supported anything more than that before, and just thinking of all the possibilities of things going wrong in that scenario makes my head hurt. :)

ptvsd/wrapper.py
```diff
@@ -1726,6 +1775,16 @@ def start_server(port, addhandlers=True):
...
     client, _ = server.accept()
     pydevd = _start(client, server)
+
+    def _wait_for_more_connections():
+        while True:
+            client, _ = server.accept()
+            _add_pydevd_event_handler(client, pydevd, name='ptvsd.Server')
+
+    connection_thread = threading.Thread(target=_wait_for_more_connections,
+                                         name='ptvsd_client_connection')
```
karthiknadig: Use `ptvsd.Client.Connection` as name, to be consistent with other `ptvsd` thread names

ptvsd/wrapper.py
```diff
@@ -1779,6 +1838,7 @@ def start_client(host, port, addhandlers=True):
     """
     client = _create_client()
     client.connect((host, port))
+    pydevd = PydevdSocket()
```
karthiknadig: That is not needed, right?

DonJayamanne: 👍

DonJayamanne:
@ericsnowcurrently 
Here are the requirements for remote debugging:
* Current functionality of PR
* Run a program
    - Somewhere down the line, we can import ptvsd and start the socket server (PTVSD)
    - It should listen for connections (**non-blocking**)
        This is a scenario where we have the socket server listening for connections, but in a non-blocking manner.
    - If a connection is made, then debugger should connect

DonJayamanne:
Closing in favor of #296

Issue #292, opened 2018-04-02
Unable to launch the debugger


DonJayamanne:
Getting the following error in master when debugging in VSC:
```
Could not connect to None: 60857
Traceback (most recent call last):
  File "/Users/donjayamanne/Desktop/Development/vscode/ptvsd/ptvsd/pydevd/pydevd.py", line 1620, in main
    debugger.connect(host, port)
  File "/Users/donjayamanne/Desktop/Development/vscode/ptvsd/ptvsd/pydevd/pydevd.py", line 326, in connect
    s = start_server(port)
  File "/Users/donjayamanne/Desktop/Development/vscode/ptvsd/ptvsd/wrapper.py", line 1766, in start_server
    server = _create_server(port)
  File "/Users/donjayamanne/Desktop/Development/vscode/ptvsd/ptvsd/wrapper.py", line 1701, in _create_server
    server.bind(('127.0.0.1', port))
OSError: [Errno 48] Address already in u
```

DonJayamanne:
This would indicate the live tests we have today are not sufficient or not mimicking VSC.
@karthiknadig Please could you test the latest master branch as well.

DonJayamanne:
@ericsnowcurrently 

The problem is the `--client` arg is not appended into the list when it should be.

```python
def _run_argv(address, filename, *extra):
    """Convert the given values to an argv that pydevd.main() supports."""
    if '--' in extra:
        pydevd = list(extra[:extra.index('--')])
        extra = list(extra[len(pydevd) + 1:])
    else:
        pydevd = []
        extra = list(extra)

    host, port=address
    print(host)
    #if host is None:
    #    host = '127.0.0.1'
    argv = [
        sys.argv[0],
        '--port', str(port),
    ]
    if host is not None:
        argv.extend([
            '--client', host,
        ])
```

Today I'm launching PTVSD by invoking the function `debug` in `debugger.py`:
```typescript
vspd.debug(filename, port_num, debug_id, debug_options, run_as)
```

DonJayamanne:
@ericsnowcurrently 
When possible, please do look at this issue. Should be simple. I'd fix it, but I'm not sure what the right approach is, as you have commented out the code that initializes the host to `127.0.0.1`.

Oh yes, this PR is quite crucial for VSC, as it breaks all of our PRs (we're pulling the master branch of PTVSD in VSC CI tests , i.e. none of our CI tests pass as a result of this 😄)

[...]

PR #296, opened 2018-04-03
Refactor before fixing support for re-attach.

ericsnowcurrently:
An alternative approach to fixing re-attach (in contrast to #287) requires separating some concerns that are currently mixed together in the code base.  This PR does so.


int19h:
I really like this code reorganization, much neater! 👍

DonJayamanne:
Got some issues, here's a screen dump of an error:
Basically running a program as follows fails:
`python -m ptvsd --server --port 9091 --file another.py`

screen shot 2018-04-02 at 9 32 45 pm

DonJayamanne:
@ericsnowcurrently Please can we add tests for the above error (this is a regression issue). Right now unit tests are passing, but its breaking VSC, meaning we need some tests. I don't mind us creating a separate issue to track the creation of unit tests for this (i.e. we don't need the tests in this PR).

DonJayamanne:
@ericsnowcurrently 
This PR breaks debugging using `launch` as well as `attach`.
Looks like we need some better unit tests. #292 is another example where we need unit tests for a similar if not exact scenaro.

screen shot 2018-04-02 at 9 32 45 pm
screen shot 2018-04-02 at 9 37 39 pm

DonJayamanne:
> I really like this code reorganization, much neater compared to our existing everything-in-wrapper.py! 👍

Agreed.

ericsnowcurrently:
@DonJayamanne This PR should no longer break debugging.  Please verify.  As to the tests, I agree and am working on it. :)

PR #308, opened 2018-04-04
Enable path mappings for remote debugging

### LINKED ISSUE #241: File/project mapping for remote debugging

DonJayamanne:
Fixes #241
Unfortunately I've had to modify a PyDev (If someone can think of a way around this I'm all ears).
The problem is when we initialize the path mappings, the attribute `norm_file_to_client` points to a different function, however its too late as pydev has already imported it using `from import`

### DIFF
--- a/ptvsd/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/ptvsd/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -84,7 +84,6 @@ import pydevd_tracing
 from _pydevd_bundle import pydevd_xml
 from _pydevd_bundle import pydevd_vm_type
 from pydevd_file_utils import get_abs_path_real_path_and_base_from_frame, NORM_PATHS_AND_BASE_CONTAINER, norm_file_to_client
-import pydevd_file_utils
 import sys
 import traceback
 from _pydevd_bundle.pydevd_utils import quote_smart as quote, compare_object_attrs_key, to_string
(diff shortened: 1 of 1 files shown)


karthiknadig:
This should work for VS as well. No need to be VSC specific. @int19h Do you think VS will need something different?

ptvsd/wrapper.py
```diff
@@ -930,10 +931,22 @@ def _parse_debug_options(self, debug_options):
             options[key] = DEBUG_OPTIONS_PARSER[key](value)
         return options
 
+    def _initialize_vsc_path_maps(self, args):
```
karthiknadig: Lets make this generic, since this will work for VS as well.

ptvsd/wrapper.py
```diff
@@ -930,10 +931,22 @@ def _parse_debug_options(self, debug_options):
...
+        pathMaps = []
+        for pathMapping in args.get('pathMappings', []):
```
karthiknadig: We should have a readme.md describing these custom message items for clarity.

DonJayamanne:
@karthiknadig Please review the updates to the readme.md
@int19h Please review the names (`pathMappings` passed in. I'd like to ensure both VS and VSC use the same message format where possible. Hence the names are important in this case).

int19h:
One general catch wrt this PR is that any change to pydevd stuff has to be in its own separate commit (so that we can easily upstream the subtree).

DonJayamanne:
> One general catch wrt this PR is that any change to pydevd stuff has to be in its own separate commit (so that we can easily upstream the subtree).

Aah yes, forgot about that. Will Fix that.

DonJayamanne:
@int19h 
Fixed

DonJayamanne:
@int19h 
Created a new PR https://github.com/Microsoft/ptvsd/pull/310

int19h:
Can you rebase and squish it to make sure that all commits are separate? I'm not sure if subtree merge is going to complain about a reverted commit... but if it applies them one by one, it probably will.

DonJayamanne:
@int19h 
Wouldn't `squash and merge` do that. (that's the default `Merge Operation` i choose)?

int19h:
It should, I just wanted to make sure that's what actually happens (since it'll commit immediately after). But I guess we can always remove those commits if it goes wrong.

PR #311, opened 2018-04-04
Fixes an issue with missing frozen frames

### LINKED ISSUE #280: Frame not found while switching to frozen lib frames
Clone the repo using:
```
git clone -b debuggerbug_exception_callstack https://github.com/qubitron/stackoverflow-django
```

This is where it falls over on change of frame:
```
find_frame: frame not found.
    Looking for thread_id:pid_996_id_2077876920104, frame_id:2077914384232
    Current     thread_id:pid_996_id_2077876920104, available frames:
    2077912942600  -  2077884835544  -  2077910162408  -  2077910410088  -  2077912758680
2077912573176  -

  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\xml\sax\expatreader.py", line 217, in feed
    self._parser.Parse(data, isFinal)
xml.parsers.expat.ExpatError: syntax error: line 1, column 0

During handling of the above exception, another exception occurred:

  File "C:\GIT\ptvsd\ptvsd\wrapper.py", line 730, in done
    fut.result()
  File "C:\GIT\ptvsd\ptvsd\futures.py", line 44, in result
    reraise(self._exc_info)
  File "C:\GIT\ptvsd\ptvsd\reraise3.py", line 11, in reraise
    raise exc_info[1].with_traceback(exc_info[2])
  File "C:\GIT\ptvsd\ptvsd\futures.py", line 159, in callback
    x = it.send(fut.result())
  File "C:\GIT\ptvsd\ptvsd\wrapper.py", line 1036, in on_variables
    xml = untangle.parse(resp_args).xml
  File "C:\GIT\ptvsd\ptvsd\untangle.py", line 200, in parse
    parser.parse(StringIO(filename))
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\xml\sax\expatreader.py", line 111, in parse
    xmlreader.IncrementalParser.parse(self, source)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\xml\sax\xmlreader.py", line 125, in parse
    self.feed(buffer)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\xml\sax\expatreader.py", line 221, in feed
    self._err_handler.fatalError(exc)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\xml\sax\handler.py", line 38, in fatalError
    raise exception
xml.sax._exceptions.SAXParseException: <unknown>:1:0: syntax error
```

PTVS bug: https://github.com/Microsoft/PTVS/issues/3961

karthiknadig:
Fixes #280

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -998,9 +998,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         try:
             xml = self.parse_xml_response(resp_args)
         except SAXParseException as ex:
-            self.send_error_response(
-                request,
-                message=ex.getMessage())
+            self.send_error_response(request)
             return
 
         try:
@@ -1140,9 +1138,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         try:
             xml = self.parse_xml_response(resp_args)
         except SAXParseException as ex:
-            self.send_error_response(
-                request,
-                message=ex.getMessage())
+            self.send_error_response(request)
             return
 
         try:
@@ -1258,9 +1254,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         try:
             xml = self.parse_xml_response(resp_args)
         except SAXParseException as ex:
-            self.send_error_response(
-                request,
-                message=ex.getMessage())
+            self.send_error_response(request)
             return
 
         try:
@@ -1299,9 +1293,7 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         try:
             xml = self.parse_xml_response(resp_args)
         except SAXParseException as ex:
-            self.send_error_response(
-                request,
-                message=ex.getMessage())
+            self.send_error_response(request)
             return
 
         try:
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -986,7 +986,13 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         # TODO: docstring
         cmd = pydevd_comm.CMD_LIST_THREADS
         _, _, resp_args = yield self.pydevd_request(cmd, '')
-        xml = self.parse_xml_response(resp_args)
+
+        try:
+            xml = self.parse_xml_response(resp_args)
+        except Exception:
+            self.send_response(request, success=False)
+            return
+
         try:
             xthreads = xml.thread
         except AttributeError:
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -986,7 +986,13 @@ def on_threads(self, request, args):
         # TODO: docstring
         cmd = pydevd_comm.CMD_LIST_THREADS
         _, _, resp_args = yield self.pydevd_request(cmd, '')
-        xml = self.parse_xml_response(resp_args)
+
+        try:
+            xml = self.parse_xml_response(resp_args)
+        except Exception:
```
int19h: Is there a more specific exception type to catch for malformed XML?

karthiknadig: xml.sax._exceptions.SAXParseException

ptvsd/wrapper.py
```diff
@@ -986,7 +987,13 @@ def on_threads(self, request, args):
         # TODO: docstring
         cmd = pydevd_comm.CMD_LIST_THREADS
         _, _, resp_args = yield self.pydevd_request(cmd, '')
-        xml = self.parse_xml_response(resp_args)
+
+        try:
+            xml = self.parse_xml_response(resp_args)
+        except SAXParseException:
+            self.send_response(request, success=False)
```
DonJayamanne: Shouldn't we reply back with the error message?

ptvsd/wrapper.py
```diff
@@ -990,8 +990,11 @@ def on_threads(self, request, args):
 
         try:
             xml = self.parse_xml_response(resp_args)
-        except SAXParseException:
-            self.send_response(request, success=False)
+        except SAXParseException as ex:
+            self.send_response(
+                request,
+                success=False,
+                message=ex.getMessage())
```
DonJayamanne: `getMessage()`!? Never seen that before. Hmm, I always thought the best way was to use `sys.exec_info()` (it's a tuple).
Best to wait for someone (@int19h) else to review this.
I'd create a helper method that sends the error `this.send_error_response(request)` (this way we have the opportunity to log the error and change what is sent back, error type, message, stack trace...)

DonJayamanne:
Fyi, you have failing tests

ptvsd/wrapper.py
int19h: This is specific to the exception that we're catching here (for the SAX parser). But I don't think we actually want to propagate that message, if it surfaces anywhere in UI, which IIRC it does. I mean, this would be something like "malformed XML", showing up in Locals - even though the user has no idea what it has to do with XML.

We could just pass the text that we failed to parse - that would say something like "frame ... not found". Still not perfect, but probably better, if we want to say something there.

PyDev seems to be using a generic error message when it can't retrieve frame variables.

DonJayamanne: > This is specific to the exception that we're catching here 

Aah yes,

karthiknadig: I think `this.send_error_response(request)` is a good idea. I will add that. If we have a generalized error response then `sys.exec_info()` would be a better idea. For specific cases, , we may be sending back information which may not be helpful. for `SAXExceptions` they have `getMessage()` which return human readable errors: https://docs.python.org/2/library/xml.sax.html

karthiknadig: As discussed, I need to remove the `getMessage` I will do that before merging.

PR #321, opened 2018-04-06
Add support for remote source references

### LINKED ISSUE #318: Add support for remote source references
Adds support for [sourceReference](https://github.com/Microsoft/vscode-debugadapter-node/blob/master/protocol/src/debugProtocol.ts#L818)

**Scenarios:**
1. Debugging std libs or similar
2. Debug a remote application without any source on the client side (all we need is, is one of, `breakpoint()`, `ptvsd.wait_for_attach` or `ptvsd.break_into_debugger ` or `an exception`).

I'd consider item 2 to be a great feature for release (I'm assuming this isn't already possible in PTVS, I know it isn't in VSC Python).

**When does this come into play:**
Only when remote debugging (`attach`) and if ptvsd is unable to map the server file to a client file (i.e. PTVSD is unable to map the file)

**Status**
* I've got this working, just need to test on Windows, Linux and a docker container
* Seems to work without any source
* Breakpoints/pause/breaking into exceptions (works)

**Items to consider/check:**
* UX (latency) when debugging large files

@karthiknadig @int19h 
Any thoughts/suggestions?

DonJayamanne:
Fixes #318
This is an opt in feature.
We'll be enabling this as a default feature in VSC.
* If users provide source mappings then, that wins over source references.
* If users do not provide source mappings, then source references are used
* If user provides source mappings, but PTVSD is unable to map the files (e.g. standard lib code), then source references are used

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
             "Jinja",                // Enables Jinja (Flask) Template debugging
             "FixFilePathCase",      // See FIX_FILE_PATH_CASE in wrapper.py
             "DebugStdLib"           // Whether to enable debugging of standard library functions
+            "UseSourceReferences"   // Whether to use source references (making `pathMappings` optional)
     ],
     "pathMappings": [
         {
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
             "Jinja",                // Enables Jinja (Flask) Template debugging
             "FixFilePathCase",      // See FIX_FILE_PATH_CASE in wrapper.py
             "DebugStdLib"           // Whether to enable debugging of standard library functions
-            "UseSourceReferences"   // Whether to use source references (making `pathMappings` optional)
     ],
     "pathMappings": [
         {
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -1039,11 +1039,13 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
         filename = '' if source_reference == 0 else \
             self.source_map.to_pydevd(source_reference)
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -44,6 +44,7 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
             "Jinja",                // Enables Jinja (Flask) Template debugging
             "FixFilePathCase",      // See FIX_FILE_PATH_CASE in wrapper.py
             "DebugStdLib"           // Whether to enable debugging of standard library functions
+            "UseSourceReferences"   // Whether to use source references (making `pathMappings` optional)
```
int19h: Is there any particular scenario that you have in mind when the user might want to *not* specify this? In VS, at least, all debuggers that support this functionality (e.g. JS), do so unconditionally.

ptvsd/wrapper.py
```diff
@@ -1027,6 +1032,54 @@ def on_threads(self, request, args):
...
+            cmd = pydevd_comm.CMD_LOAD_SOURCE
+            _, _, content = yield self.pydevd_request(cmd, server_filename)
+            self.send_response(request, content=content)
+
+    def get_source_reference(self, filename):
+        """Gets the source reference only in remote debugging scenarios.
+        And we know that the path returned is the same as the server path
+        (i.e. path has not been translated)"""
+
+        if self.start_reason == 'launch' or \
```
int19h: I wonder if it's worth disabling it for launch, given that "remote" attach can also be local - and often is in practice (e.g. for most people using our debugger with embedded Python scripting). It looks like the code below would take care of things either way?

README.md
DonJayamanne: I couldn't think of any reason, which is why I was planning on making this the default in VSC.

ptvsd/wrapper.py
DonJayamanne: Sorry, but I'm not following you here.

README.md
DonJayamanne: Will remove this.

PR #334, opened 2018-04-10
Fix path mappings and settings for cross-platform launch

karthiknadig:


karthiknadig:
Test errors don't look like they were caused by this change.

PR #341, opened 2018-04-12
Ability to run code without debugging

### LINKED ISSUE #339: Running python code without debugging
VSC uses the debug adapter to run code without debugging.
The plan is to move that functionality into PTVSD as well (as opposed to leaving python code in VSC).

Justification:
* Ability to run python code (file, module, etc) - Already available in PTVSD
* Ability to redirect output via protocol messages - Already available in PTVSD
* Ability to send/receive messages DAP messages- Already available in PTVSD

DonJayamanne:
Fixes #339

### DIFF
--- a/ptvsd/runner.py
+++ b/ptvsd/runner.py
@@ -13,18 +13,347 @@ import time
 import threading
 import warnings
 
-from ptvsd.daemon import DaemonClosedError
 from ptvsd import ipcjson, __version__
+from ptvsd.daemon import DaemonClosedError
+from ptvsd.pydevd_hooks import start_client
 from ptvsd.socket import close_socket
 from ptvsd.wrapper import WAIT_FOR_DISCONNECT_REQUEST_TIMEOUT, WAIT_FOR_THREAD_FINISH_TIMEOUT
-from pydevd import CheckOutputThread, init_stdout_redirect, init_stderr_redirect
-from _pydevd_bundle.pydevd_kill_all_pydevd_threads import kill_all_pydev_threads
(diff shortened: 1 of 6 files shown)


ptvsd/runner.py
```diff
@@ -0,0 +1,359 @@
...
+                    'filter': 'raised',
+                    'label': 'Raised Exceptions',
+                    'default': False
+                },
+                {
+                    'filter': 'uncaught',
+                    'label': 'Uncaught Exceptions',
+                    'default': True
+                },
+            ],
```
karthiknadig: Yet another copy of this. I created a work item to create a single location for these values. This is now duplicated in too many places.
[https://github.com/Microsoft/ptvsd/issues/342](https://github.com/Microsoft/ptvsd/issues/342)

DonJayamanne: This will not be used by VSC when running, but I see your point.
Thanks for creating that ticket.

DonJayamanne: You'll notice that I'm returning `success` for almost all requests, that's because VSC doesn't really use them, except of course when `success=false`

ptvsd/runner.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def start_message_processor(host, port_num):
+    launch_notification = threading.Event()
+
+    def on_launch():
+        launch_notification.set()
+
+    daemon = Daemon(
+        notify_launch=on_launch, addhandlers=True, killonclose=True)
```
int19h: Looks like this could just pass `launch_notification.set` directly.

Issue #14, opened 2018-01-24
Hit-count conditional breakpoints


int19h:
Need to support `hitCondition` attribute in `setBreakpoints`.

PR #346, opened 2018-04-13
Implement hit condition breakpoint

karthiknadig:
#14


ptvsd/pydevd/_pydevd_bundle/pydevd_frame.py
```diff
@@ -39,6 +39,12 @@ def send_signature_call_trace(*args, **kwargs):
 def handle_breakpoint_condition(py_db, info, breakpoint, new_frame):
     condition = breakpoint.condition
     try:
+        if breakpoint.handle_hit_condition():
+            return True
+
+        if condition is None:
+            return False
+
         return eval(condition, new_frame.f_globals, new_frame.f_locals)
```
int19h: Or maybe we should just pass `new_frame` to `handle_hit_condition`, and then do the eval there in the same way it happens here? This would potentially allow for hit conditions that are dynamic (e.g. depend on the value of a local variable), which is a neat feature.

ptvsd/pydevd/_pydevd_bundle/pydevd_breakpoints.py
```diff
@@ -44,15 +45,35 @@ def __str__(self):
...
         self.func_name = func_name
         self.expression = expression
         self.suspend_policy = suspend_policy
+        self.hit_condition = hit_condition
+        self._hit_count = 0
+        self._hit_condition_lock = threading.Lock()
         # need for frame evaluation: list of code objects, which bytecode was modified by this breakpoint
         self.code_objects = set()
 
+    def has_condition(self):
```
int19h: `@property`

ptvsd/pydevd/_pydevd_bundle/pydevd_frame.py
karthiknadig: This is a good suggestion. I will update it.

tests/highlevel/test_messages.py
```diff
@@ -944,19 +923,67 @@ def test_initial(self):
...
+             '4\tpython-line\tspam.py\t25\tNone\tNone\tNone\t@HIT@ % 5 == 0'),
         ])
 
     def test_with_existing(self):
         with self.launched():
             with self.hidden():
                 self.PYDEVD_CMD = CMD_SET_BREAK
                 self.expected_pydevd_request(
-                    '1\tpython-line\tspam.py\t10\tNone\tNone\tNone')
+                    '1\tpython-line\tspam.py\t10\tNone\tNone\tNone\tNone')
```
int19h: This is a non-backwards-compatible pydevd protocol change, which might make it harder to upstream. I think we shouldn't  be reporting this field if it's `None` - that way old breakpoints works as they always did, and only the new ones get the extra new field to parse.

huguesv:
You mixed changes to pydevd and ptvsd/wrapper.py in the same commit, is this something you still want to avoid doing?

int19h:
Oh, indeed! Thanks for pointing that out Hugues!

tests/highlevel/test_messages.py
karthiknadig: In pydevd, I support the old and the new breakpoints. However, should I just add CMD_SET_BREAK2 that can be configured using json?

int19h: Oh, sorry, I got confused about the direction here. Never mind!

DonJayamanne:
@int19h @karthiknadig 
You'll want to ensure the option `[Squas and merge]` is always used for PTVSD.
* Go into `Settings`
* Unselect all checkboxes except for `Allow squash merging` under the section `Merge Button`

This will ensure all commits are merged.
I know I've been asked to to this manually, but this isn't necessary.
This is crucial for this project to ensure changes to PyDevd don't leak out via commits.
E.g. if this PR is merged without `Squashing`, then the changes to PyDevd will go into PTVSD (even if @karthiknadig were to revert those changes in this PR).

We should set this as the default setting (we've done the same in vsc as well).
This will make the code base commits cleaner as well (i.e. one commit per merged PR), and with the added benefit of not leaking changes into Pydevd.

int19h:
Correct me if I'm wrong, but wouldn't squashing just create a single commit that touches on both ptvsd and pydevd bits?

DonJayamanne:
>  both ptvsd and pydevd bits?

As discussed, the pydevd changes would come in via a separate PR

karthiknadig:
Closing this in favor of : https://github.com/Microsoft/ptvsd/pull/359

PR #348, opened 2018-04-14
Ability to enable debugging of a program via a remote debugger

### LINKED ISSUE #251: Add support for ptvsd.enable_attach
- [ ] Pausing does not work (I guess this is because when attaching, if there are no breakpoints, then pausing won't work).
- [ ] Breaking into debugger using (`break_into_debugger`) might not work for the same reason as above.

### LINKED ISSUE #252: Add support for ptvsd.wait_for_attach

DonJayamanne:
Fixes #251 
Fixes #252

Enable this feature by using a fake socket (pydev needs one).
I have another solution that doesn't use a fake socket, but I requires more work and we cannot use `pydevd.settrace` method. Basically manually do what's done in `pydevd.settrace`. 

I believe the fake socket is cleaner.

### DIFF
--- a/ptvsd/__init__.py
+++ b/ptvsd/__init__.py
@@ -5,8 +5,9 @@
 import sys
 import os.path
 
-from ptvsd.version import __version__, __author__  # noqa
+__all__ = ['enable_attach', 'wait_for_attach', 'break_into_debugger', 'is_attached'] # noqa
 
+from ptvsd.version import __version__, __author__  # noqa
 
 PYDEVD_ROOT = os.path.join(os.path.dirname(__file__), 'pydevd')
 del os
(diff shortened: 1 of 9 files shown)


int19h:
That's much neater than I thought it would be; nice!

One thing - the `attach_server` module in the old debugger was always meant to be an implementation detail, in a sense that users are meant to import `enable_attach` etc directly from the top-level `ptvsd` package. So we don't have to duplicate that name here, or the overall arrangement. It's still descriptive enough - but if you felt like renaming it, or even just putting that stuff into `__init__.py` directly, that can be done too.

DonJayamanne:
> So we don't have to duplicate that name here, or the overall arrangement. It's still descriptive enough - but if you felt like renaming it, or even just putting that stuff into __init__.py directly, that can be done too.

I've left the stuff in `__init__.py` as this is available today. Meaning people might be using importing from `ptvsd` as follows:
```python
import ptvsd
ptvsd.enable_attach(...)
```

I can remove that.

zooba:
Since I spotted this, we consider `ptvsd.enable_attach(...)` to be essential for backwards compatibility. So please don't remove it (though you can deprecate it and make it redirect toward a new function, but there's little value).

Also, that commit list makes me hope you guys are doing squash merges 😆

DonJayamanne:
> Also, that commit list makes me hope you guys are doing squash merges

Always

DonJayamanne:
@zooba, the new API isn't 100% backwards compatible (we're missing the secret argument and some others). Is this current API going or should we try to account for the old API.

Issue #350, opened 2018-04-16
Add support for log points using the new DAP protocol


DonJayamanne:

int19h:
Is this a dupe of #313?

DonJayamanne:
:roll_eyes:  sorry
Closing in favor of #313

PR #351, opened 2018-04-16
Continue execution after evaluation of expressions in breakpoints

DonJayamanne:
Part of the fix for #350

For log points, all we need to do is, to log a message.
This PR makes use of breakpoints, but ensures we do not stop at a breakpoint (kind of link a conditional breakpoint, except for the fact that the condition never evaluates to true).

The key change is in `ptvsd/pydevd/_pydevd_bundle/pydevd_frame.py`
```python
                         if breakpoint.expression is not None:
                             handle_breakpoint_expression(breakpoint, info, new_frame)	                             handle_breakpoint_expression(breakpoint, info, new_frame)
                            if breakpoint.continue_execution == False:
                                return self.trace_dispatch
```

### DIFF
--- a/ptvsd/pydevd/_pydevd_bundle/pydevd_process_net_command.py
+++ b/ptvsd/pydevd/_pydevd_bundle/pydevd_process_net_command.py
@@ -249,11 +249,18 @@ def process_net_command(py_db, cmd_id, seq, text):
                 # command to add some breakpoint.
                 # text is file\tline. Add to breakpoints dictionary
                 suspend_policy = "NONE"
+                continue_execution = None
                 hit_condition = None
                 if py_db._set_breakpoints_with_id:
                     try:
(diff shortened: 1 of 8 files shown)


ptvsd/pydevd/_pydevd_bundle/pydevd_process_net_command.py
```diff
@@ -249,11 +249,18 @@ def process_net_command(py_db, cmd_id, seq, text):
                 # command to add some breakpoint.
                 # text is file\tline. Add to breakpoints dictionary
                 suspend_policy = "NONE"
+                continue_execution = None
```
int19h: Is there any particular reason to have this as a tri-state bool? It seems that there are no checks that distinguish `True` and `None`.

DonJayamanne:
@int19h Fixed.

PR #352, opened 2018-04-16
WIP - Support for log points

DonJayamanne:
Second part of the fix for #350

Note: The following tasks are remaining:
* Update the protocol schema
* Update the capabilities (PR #346, will make things simpler, hence waiting for that to be merged)
* Update tests


ptvsd/wrapper.py
```diff
@@ -1532,14 +1536,32 @@ def on_setBreakpoints(self, request, args):
...
+                    expression = 'print("{}")'.format(logMessage.replace('"', '\\"')) # noqa
+                else:
+                    raw_text = reduce(lambda a, b: a.replace(b, '{}'), expressions, logMessage) # noqa
+                    raw_text = raw_text.replace('"', '\\"')
+                    expression_list = ', '.join([s.strip('{').strip('}').strip() for s in expressions]) # noqa
+                    expression = 'print("{}".format({}))'.format(raw_text, expression_list) # noqa
+
             msg = msgfmt.format(vsc_bpid, bp_type, path, line,
-                                src_bp.get('condition', None))
+                                condition, expression, continue_execution)
```
karthiknadig: I think at this point we should just add CMD_SET_BREAK_JSON and do all of this via JSON? Because we have `suspend_policy` that will eventually come here along with `hit_condition`.

DonJayamanne: Absolutely no idea what that command is, will check that out.

DonJayamanne: I don't mind adding a new command, however what you're suggesting is a change to the structure of the message, i.e. change in the protocol. I'd leave that for a separate PR, its possible such a change might never get approved upstream (as we're introducing different protocols into the library).

DonJayamanne: > I think at this point we should just add CMD_SET_BREAK_JSON and do all of this via JSON? Because we have suspend_policy that will eventually come here along with hit_condition.

the plan is to merge this PR after yours

DonJayamanne:
@int19h @karthiknadig 
Tests will fail until the previous PR is merged (i.e. this PR depends on changes to PyDevd in #351)

DonJayamanne:
tests are now passing.

DonJayamanne:
@int19h @karthiknadig 
Please review and approve, as this breaks anyone using the latest ptvsd (as the api has changed in pydevd)

ptvsd/wrapper.py
```diff
@@ -1580,17 +1585,34 @@ def on_setBreakpoints(self, request, args):
...
+            if len(logMessage) == 0:
+                is_logpoint = None
+                condition = src_bp.get('condition', None)
+                expression = None
+            else:
+                is_logpoint = True
+                condition = None
+                expressions = re.findall('\{.*?\}', logMessage)
+                if len(expressions) == 0:
+                    expression = 'print("{}")'.format(logMessage.replace('"', '\\"')) # noqa
```
int19h: This also needs to escape backslashes. But I think the easiest way to do it here is to use repr, and let Python do the escaping and add the quotes: `print({}).format(repr(logMessage))`

tests/highlevel/test_live_pydevd.py
```diff
@@ -196,3 +201,73 @@ def test_no_breakpoints(self):
...
+        d = 4
+        """
+
+    @contextlib.contextmanager
+    def running(self):
+        addr = (None, 8888)
+        with self.fake.start(addr):
+                yield
+
+    def test_launch(self):
```
karthiknadig: test_basic?

ptvsd/wrapper.py
DonJayamanne: :+1:

tests/highlevel/test_live_pydevd.py
DonJayamanne: :+1:

DonJayamanne:
Fixed

PR #369, opened 2018-04-17
Check whether variable is None before checking the length

### LINKED ISSUE #367: Error in pydev breakpoints
```python
                if len(condition) <= 0 or condition is None or condition == "None":
                    condition = None

                if len(expression) <= 0 or expression is None or expression == "None":
                    expression = None

                if len(hit_condition) <= 0 or hit_condition is None or hit_condition == "None":
                    hit_condition = None
```

We need to check if variables are None before checking the length.

DonJayamanne:
Fix #367


karthiknadig:
Good catch.

PR #371, opened 2018-04-18
Wait until the end of VSCodeMessageProcessor.start() to start handling incoming messages.

ericsnowcurrently:
This guarantees that the `output` message is always sent first.


DonJayamanne:
Good catch

PR #372, opened 2018-04-18
Hiding debugger internals from stack, modules, etc

karthiknadig:
#26

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -621,34 +621,32 @@ class InternalsFilter(object):
     """Identifies debugger internal artifacts.
     """
     # TODO: Move the internal thread identifier here
-    _ignore_files = [
-        '/ptvsd_launcher.py',
-    ]
+    def __init__(self):
+        self._ignore_files = [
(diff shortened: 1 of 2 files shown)


ptvsd/wrapper.py
```diff
@@ -1180,10 +1191,38 @@ def on_stackTrace(self, request, args):
...
+            if not self.is_debugger_internal_path(frame['source']['path']):
+                user_frames.append(frame)
+
+        totalFrames = len(user_frames)
         self.send_response(request,
-                           stackFrames=stackFrames,
+                           stackFrames=user_frames,
                            totalFrames=totalFrames)
 
+    def is_debugger_internal_path(self, file_path):
```
DonJayamanne: Can we move this out of this class, right now it feels like this class is doing way more than it needs to.

DonJayamanne: This way, it gets easier to unit test the changes too (without having to run any live tests or instantiate this class).

karthiknadig:
There are a few path issues on nix. Will update with a fix.

PR #374, opened 2018-04-18
 Clean up handling of vendored projects.

ericsnowcurrently:
This allows us to keep ptvsd/__init__.py clean and focused.  It also helps isolate the vendoring-related code.

Note that I have not moved the `ptvsd/pydevd` directory to be `ptvsd/_vendored/pydevd`.  From what I understand, that's a git submodule and I am uncertain of what special considerations are required when moving the directory.  If it requires nothing special then I'll take care of it here.  Otherwise it can be addressed separately.

### DIFF
--- a/ptvsd/__init__.py
+++ b/ptvsd/__init__.py
@@ -8,11 +8,11 @@ __all__ = [
 ]
 
 
-# _vendored_pydevd must be imported first to ensure that the
-# ptvsd-vendored copy of pydevd gets used.
-from . import _vendored_pydevd
+# "force_pydevd" must be imported first to ensure (via side effects)
+# that the ptvsd-vendored copy of pydevd gets used.
+from ._vendored import force_pydevd
 from ptvsd.version import __version__, __author__
 from ptvsd.attach_server import (
     enable_attach, wait_for_attach, break_into_debugger, is_attached,
 )
-del _vendored_pydevd
+del force_pydevd
--- a/ptvsd/__init__.py
+++ b/ptvsd/__init__.py
@@ -3,46 +3,16 @@
 # for license information.
 
 __all__ = [
+    '__version__', '__author__',
     'enable_attach', 'wait_for_attach', 'break_into_debugger', 'is_attached',
 ]
 
-import sys
-import os.path
 
-from ptvsd.version import __version__, __author__  # noqa
-
-PYDEVD_ROOT = os.path.join(os.path.dirname(__file__), 'pydevd')
-del os
-
-# Ensure that pydevd is our vendored copy.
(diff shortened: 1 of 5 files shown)


int19h:
It is a subtree rather than a submodule. I am not sure how to properly move it, though - submodule would be easier, since we could just recreate it. 

Come to think of it, would a submodule out of our own separate repo (which would then be just a straight-up fork of the pydevd repo) be preferable here? It'd also take care of the issue with inadvertently making commits across subtree boundary. 

@DonJayamanne, @karthiknadig, what do you think?

zooba:
Provided we can pin versions precisely for anything our lawyers need to review, doing this should be okay.

If we can't pin and patch vendored packages, that should be a shipblocker.

int19h:
To clarify, we wouldn't do a submodule that points directly at https://github.com/fabioz/PyDev.Debugger/ - that would have all the problems you've listed. Instead, we'd fork that, apply all of our changes currently in the subtree, and point the submodule at the fork. This way we're fully in control of the bits being shipped. Indeed, it should be bit-for-bit identical (except for `.git`) to our current stuff once it's checked out.

ericsnowcurrently:
I wasn't necessarily suggesting we use submodules.  I just didn't remember what we were using. :)  It sounds like submodule and subtree each [have good and bad](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree).  My point was that I want to move the `ptvsd/pydevd directory` to `ptvsd/_vendored/pydevd`.  Is that easier with subtree (now) or with submodule?

int19h:
I don't know how to properly rename a subtree, but it seems that it's more complicated than git mv. I couldn't find any clear guidance on this. 

If we extract it to a submodule, it would go like this:
1. Create a repo that's a fork of the main pydevd repo (from the same commit that went into subtree).
2. Do a subtree merge into that repo, so that all our changes are in there.
3. Delete the subtree.
4. Create a submodule in a new place.
5. Change all the build scripts etc to do git clone --recursive. 

So for a one-off, it's probably at least as involved as moving the subtree. But it feels like, based on the way we actually work on this code, a submodule *might* be a better fit. Originally I didn't consider a submodule, because it didn't occur to me that it could point at our fork rather than upstream.

ericsnowcurrently:
@int19h (or anyone else), do you have any objections to moving the `ptvsd/pydevd` directory to `ptvsd/_vendored/pydevd`?  *How* we do it can be addressed in a separate PR. :)  FWIW, now that I've read about subtrees a bit, I'm pretty sure that it will be as trivial as a single "git mv" command.

int19h:
Let's do it!

PR #381, opened 2018-04-19
Support re-attaching to an adapter.

### LINKED ISSUE #208: Detach/reattach
It should be possible to detach from a debugged process, allow it to run, then reattach and continue debugging.

ericsnowcurrently:
(fixes #208)

> [WIP]  This is mostly done.  There's one failing test (`tests.highlevel.test_live_pydevd.LogpointTests`), which has to do with how a session connection is closed.  Once that is resolved this should be good to go.

Currently when ptvsd starts up (as server) it listens for a connection and shuts down when the debug session is done.  This prevents re-attach.  To fix this we go back to listening for a connection after getting a "disconnect" (or the client closes the connection).

### DIFF
--- a/ptvsd/_util.py
+++ b/ptvsd/_util.py
@@ -1,0 +1,142 @@
+import contextlib
+import threading
+
+
+def call_all(callables, *args, **kwargs):
+    """Return the result of calling every given object."""
+    results = []
+    for call in callables:
+        try:
+            call(*args, **kwargs)
+        except Exception as exc:
+            results.append((call, exc))
+        else:
+            results.append((call, None))
+    return results
+
+
+########################
+# closing stuff
+
+class ClosedError(RuntimeError):
+    """Indicates that the object is closed."""
+
+
+def close_all(closeables):
+    """Return the result of closing every given object."""
+    results = []
+    for obj in closeables:
+        try:
+            obj.close()
+        except Exception as exc:
+            results.append((obj, exc))
+        else:
+            results.append((obj, None))
+    return results
+
+
+class Closeable(object):
+    """A base class for types that may be closed."""
+
+    NAME = None
+    FAIL_ON_ALREADY_CLOSED = True
+
+    def __init__(self):
+        super(Closeable, self).__init__()
+        self._closed = False
+        self._closedlock = threading.Lock()
+        self._resources = []
+        self._handlers = []
+
+    def __del__(self):
+        try:
+            self.close()
+        except ClosedError:
+            pass
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    @property
+    def closed(self):
+        return self._closed
+
+    def add_resource_to_close(self, resource, before=False):
+        """Add a resource to be closed when closing."""
+        close = resource.close
+        if before:
+            def handle_closing(before):
+                if not before:
+                    return
+                close()
+        else:
+            def handle_closing(before):
+                if before:
+                    return
+                close()
+        self.add_close_handler(handle_closing)
+
+    def add_close_handler(self, handle_closing, nodupe=True):
+        """Add a func to be called when closing.
+
+        The func takes one arg: True if it was called before the main
+        close func and False if after.
+        """
+        with self._closedlock:
+            if self._closed:
+                if self.FAIL_ON_ALREADY_CLOSED:
+                    raise ClosedError('already closed')
+                return
+            if nodupe and handle_closing in self._handlers:
+                raise ValueError('close func already added')
+
+            self._handlers.append(handle_closing)
+
+    def check_closed(self):
+        """Raise ClosedError if closed."""
+        if self._closed:
+            if self.NAME:
+                raise ClosedError('{} closed'.format(self.NAME))
+            else:
+                raise ClosedError('closed')
+
+    @contextlib.contextmanager
+    def while_not_closed(self):
+        """A context manager under which the object will not be closed."""
+        with self._closedlock:
+            self.check_closed()
+            yield
+
+    def close(self):
+        """Release any owned resources and clean up."""
+        with self._closedlock:
+            if self._closed:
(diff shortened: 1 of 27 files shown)


tests/helpers/__init__.py
```diff
@@ -8,8 +8,8 @@ class Closeable(object):
     def __init__(self):
         self._closed = False
 
-    def __del__(self):
-        self.close()
+#    def __del__(self):
```
int19h: Was this for debugging, or intentional?

tests/helpers/debugadapter.py
```diff
@@ -9,40 +10,96 @@ class DebugAdapter(Closeable):
...
+            argv += list(extra)
+        return cls.start(argv, addr=addr)
+
+    @classmethod
+    def start_embedded(cls, addr, filename, redirect_output=True):
+        addr = Address.as_server(*addr)
+        with open(filename, 'r+') as scriptfile:
+            content = scriptfile.read()
+            if 'ptvsd.enable_attach' not in content:
+                raise NotImplementedError
```
int19h: This looks more like an assert to me (the script *has* to invoke enable_attach, and this helps catch that instead of failing in a weird way). And if we intend to surface it for diagnostics, then perhaps treat it as invalid argument - so `ValueError` with a message?

tests/helpers/__init__.py
ericsnowcurrently: Yeah, I didn't mean to leave that.  I'll fix it.

tests/helpers/debugadapter.py
ericsnowcurrently: Conceivably it could be handled by automatically fixing the script.  That's why I left it NotImplementedError.  However, an assert (with a comment) is sufficient until it matters.

DonJayamanne:
@ericsnowcurrently Tested and seems to be ok.

DonJayamanne:
@ericsnowcurrently , found a bug, see below:
* Start debugging (`python -m ptvsd --server --port 5678 --file one.py`)
* Connect and disconnect using the following python code:
```python
import socket
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.settimeout(1)
sock.connect(('localhost', 5678))
sock.close()
print('ok')
```
*** The above code only works 3 times, after that, PTVSD seems to be in a broken state**
I.e. `s.connect` times out.

ericsnowcurrently:
The issue with misbehaving connections should be fixed now.  Once I've gotten TravisCI to comply the PR will be ready to go.

ptvsd/_util.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def debug(*msg, **kwargs):
+    if not DEBUG:
+        return
+    tb = kwargs.pop('tb', False)
+    assert not kwargs
+    if tb:
+        import traceback
+        traceback.print_exc()
```
int19h: This should probably go to stderr, so that it can be redirected separately for more convenient debugging in case the test app has substantial amount of regular output.

ptvsd/_util.py
```diff
@@ -0,0 +1,245 @@
...
+    """A base class for types that may be closed."""
+
+    NAME = None
+    FAIL_ON_ALREADY_CLOSED = True
+
+    def __init__(self):
+        super(Closeable, self).__init__()
+        self._closed = False
+        self._closedlock = threading.Lock()
+        self._resources = []
```
int19h: Appears to be unused.

ptvsd/_util.py
```diff
@@ -0,0 +1,245 @@
...
+
+class Startable(object):
+    """A base class for types that may be started."""
+
+    RESTARTABLE = False
+    FAIL_ON_ALREADY_STOPPED = True
+
+    def __init__(self):
+        super(Startable, self).__init__()
+        self._is_running = None
```
int19h: Looks like `lambda: False` here would simplify the code that invokes this.

ptvsd/socket.py
```diff
@@ -45,6 +61,80 @@ def ignored_errno(*ignored):
             raise
 
 
+class KeepAlive(namedtuple('KeepAlive', 'interval idle, maxfails')):
```
int19h: Does the comma in the field list have any significance?

ptvsd/socket.py
```diff
@@ -45,6 +61,80 @@ def ignored_errno(*ignored):
...
+            float(maxfails) if maxfails or maxfails == 0 else cls.MAX_FAILS,
+        )
+        return self
+
+    def apply(self, sock):
+        """Set the keepalive values on the socket."""
+        sock.setsockopt(socket.SOL_SOCKET,
+                        socket.SO_KEEPALIVE,
+                        1)
+        try:
```
int19h: Might be worth moving `self.` attribute access out of the try-block here (i.e. assign to locals first), to avoid potential issues with the wrong `AttributeError` being caught and swallowed if e.g. someone renames an attribute.

MikhailArkhipov:
@ericsnowcurrently - this has been open for a while, let's finish it asap. Thanks.

ptvsd/_util.py
ericsnowcurrently: sounds good

ptvsd/_util.py
ericsnowcurrently: dropped

ptvsd/_util.py
ericsnowcurrently: fixed

ptvsd/socket.py
ericsnowcurrently: It's an innocuous typo. :)  I'll fix it.

ptvsd/socket.py
ericsnowcurrently: It shouldn't be an issue since the property would have to be actively removed.  However, I don't mind pulling those lookups out. :)  Regardless, I'm glad you made me look at this code as I found a typo. :)

Issue #315, opened 2018-04-05
CheckCommandTests fails on Windows


huguesv:
```
======================================================================
FAIL: test_match (tests.system_tests.test_schema.CheckCommandTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Projects\GitHub\Microsoft\ptvsd\tests\system_tests\test_schema.py", line 154, in test_match
    """))
AssertionError: 'chec[21 chars]e...\r\ncomparing with upstream schema file...[20 chars]\r\n' != 'chec[21 chars]e...\ncomparing with upstream schema file...\n[14 chars]ay\n'
- checking local schema file...
?                              -
+ checking local schema file...
- comparing with upstream schema file...
?                                       -
+ comparing with upstream schema file...
- schema file okay
?                 -
+ schema file okay
```

ericsnowcurrently:
Yeah, I noticed that too just the other day.  I'll look into it.

PR #390, opened 2018-04-27
Fix for DownloadCommandTests and CheckCommandTests

karthiknadig:
#315

### DIFF
--- a/tests/system_tests/test_schema.py
+++ b/tests/system_tests/test_schema.py
@@ -1,7 +1,6 @@
 import contextlib
 import io
 import os.path
-import platform
 import subprocess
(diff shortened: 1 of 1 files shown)


tests/system_tests/test_schema.py
```diff
@@ -45,19 +46,28 @@ def tearDown(self):
         self._tempdir.cleanup()
         super().tearDown()
 
+    def get_expected_stdout(self, lines):
+        is_windows = platform.system() == 'Windows'
+        newline = '\r\n' if is_windows else '\n'
```
int19h: `os.linesep` should give you this.

PR #404, opened 2018-05-02
[WIP] Add get next statement targets command that provides a list of valid locations user can go to.

karthiknadig:
This feature is needed to reach parity in VS. This is currently planned for 15.8 in VS.


zooba:
It looks right to me, though I haven’t dug into the spec. I assume the use of ord() is because it’s str on Python 2? Should we do that switch outside the loop? Or at least comment on why it’s needed.

karthiknadig:
Yes, ord() is for Python 2. I will add a comment there.

---
In reply to: [386328066](https://github.com/Microsoft/ptvsd/pull/404#issuecomment-386328066) [](ancestors = 386328066)

int19h:
@fabioz, can you take a look? This is a bigger change to pydevd than most of the other things we've done so far, and we'll want to upstream it eventually.

fabioz:
The code itself seems good to get the possible lines of code that can be executed in general in a frame, but for the set next statement there are other things that may need to be checked... 

From what I remember, the set next can't enter/exit blocks in python (i.e.: it can't enter/exit a try..except nor a with statement). 

Not sure if you want to handle this case (I don't know if Python exposes this any way -- I think you'd need to get the AST to check that), but I think that's the case which is more useful (i.e.: the user usually knows that it can't go to another method, it's the limitations inside the same method that are more difficult to handle).

The other point is that I think this feature deserves a test-case. Historically it's hard to make the debugger work on all the needed Python versions -- pydevd still supports Python 2.6 -- I'm thinking about dropping it in the future, but until then, `.format` still can't be used.

Also, I see you're not running pydevd tests -- you may want to consider that given that its code is being changed (i.e.: https://travis-ci.org/fabioz/PyDev.Debugger).

karthiknadig:
I will update this PR with tests for this. Created a workitem to enable pydevd tests in our CI #410.

@int19h, @brettcannon, any suggestions on handling `try..catch`, `with`, or any other cases to consider here? I was wondering if we could just fail the `GoTo request` for cases where `pydevd` cannot succeed with the set next statement request. That way we can avoid analyzing this ourselves. This assumes that after the failed set next statement we are not in a bad state and we can set next to a different location.

int19h:
I know that C++ and C# in VS will sometimes fail to "set next statement", and just give you an error message along the lines of "can't do that here, sorry". So we won't be shattering any expectations if we do the same.

And, unlike with invalid breakpoint locations, here you'll get that feedback immediately, rather than just quietly not hitting the breakpoint. So there's much less justification to bother with something as involved as parsing AST (or bytecode). I don't think it's worth it to try to detect it in advance, unless there really is some issue with invalid state after a failed request.

karthiknadig:
@int19h Let me know if this looks good.

fabioz:
@karthiknadig Current version seems good to me.

Issue #169, opened 2018-03-05
Add ptvsd modules to DONT_TRACE


int19h:
pydevd has a list of files that are excluded from tracing in `pydevd_dont_trace_files.py`. It's a dict, so we can add our own entries to it after loading that module.

karthiknadig:
Is this part of Just my code?

int19h:
It would probably have to be done as part of JMC, but it looks like it might have some minor perf benefits outside of that, even if we just add hardcoded list of our .py files there.

karthiknadig:
@int19h Files in that list have `pydevd_` prefix. We may need to put similar prefix `ptvsd_` for that to work correctly, without out accidentally hiding users' files. What do you think?

PR #408, opened 2018-05-03
Filter ptvsd files while tracing

karthiknadig:
Potential solution to #169

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -156,7 +156,7 @@ DONT_TRACE_FILES = {
 
 
 def filter_ptvsd_files(file_path):
-    ptvsd_path_re = r"(ptvsd[\\\/].*\.[pP][yY])"
+    ptvsd_path_re = r"(ptvsd[\\\/].*\.py)"
     matches = re.finditer(ptvsd_path_re, file_path)
     for _, match in enumerate(matches):
         for g in match.groups():
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -130,41 +130,39 @@ SafeReprPresentationProvider._instance = SafeReprPresentationProvider()
 str_handlers = pydevd_extutil.EXTENSION_MANAGER_INSTANCE.type_to_instance.setdefault(pydevd_extapi.StrPresentationProvider, [])  # noqa
 str_handlers.insert(0, SafeReprPresentationProvider._instance)
 
-PTVSD_FILE = 2
-DONT_TRACE_FILES = {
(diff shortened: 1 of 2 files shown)


ptvsd/wrapper.py
```diff
@@ -129,6 +130,45 @@ def using_format(self, fmt):
 str_handlers = pydevd_extutil.EXTENSION_MANAGER_INSTANCE.type_to_instance.setdefault(pydevd_extapi.StrPresentationProvider, [])  # noqa
 str_handlers.insert(0, SafeReprPresentationProvider._instance)
 
+PTVSD_FILE = 2
```
int19h: Is this for future extensibility? I would expect the below to be a set, rather, since all the values are the same, and only the keys differ.

Also, why 2? Does it correspond to something inside pydevd? If so, this probably needs a comment.

ptvsd/wrapper.py
```diff
@@ -129,6 +130,47 @@ def using_format(self, fmt):
...
+    """
+    Returns true if the file should not be traced.
+    """
+    ptvsd_path_re = r"(ptvsd[\\\/].*\.py)"
+    matches = re.finditer(ptvsd_path_re, file_path)
+    for _, match in enumerate(matches):
+        for g in match.groups():
+            try:
+                return g in DONT_TRACE_FILES
+            except KeyError:
```
int19h: You don't need the `except` anymore, singe `in` should never raise this for a missing key. 

Also, I think this entire loop can now become something like:
```py
return all((g in DONT_TRACE_FILES
            for m in matches
            for g in m.groups))
```

karthiknadig: Ah! yes i missed that. Will update.

ptvsd/wrapper.py
```diff
@@ -129,6 +130,47 @@ def using_format(self, fmt):
 str_handlers = pydevd_extutil.EXTENSION_MANAGER_INSTANCE.type_to_instance.setdefault(pydevd_extapi.StrPresentationProvider, [])  # noqa
 str_handlers.insert(0, SafeReprPresentationProvider._instance)
 
+DONT_TRACE_FILES = {
+    os.path.join('ptvsd', 'attach_server.py'),
```
int19h: One other thought - if you just put tuples here, i.e.:
```py
{
  ('ptvsd', 'attach_server.py'),
  ('ptvsd', 'daemon.py'),
  ...
}
```
then you can iterate over that list and just invoke `os.path.join(*elem)` for every element, to avoid having to repeat it over and over again.

PR #418, opened 2018-05-10
[WIP] Fixes issue with debuggee not terminating after 'stop debug'

### LINKED ISSUE #325: Debuggee no longer terminates
* Python program
image
* Add a break point at line 4
* Debug the above code
* Stop debugging
* You'll notice that the 4th and 5th lines are executed.

Here's the original issue https://github.com/Microsoft/vscode-python/issues/1345

**This used to work in PTVSD**. 
**Here's a commit that it used to work in 608ca082dad71b4394fdb2e5d0615a78970c0f0f**

Note:
* The above commit in PTVSD works with VSC. I.e. its a PTVSD bug. 
* Seems like we're continuing the code

karthiknadig:
Fixes #325

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -804,7 +804,6 @@ class VSCodeMessageProcessor(ipcjson.SocketIO, ipcjson.IpcChannel):
     def _stop_vsc_message_loop(self):
         self.set_exit()
         self.loop.stop()
-        self.event_loop_thread.join(WAIT_FOR_THREAD_FINISH_TIMEOUT)
         if self.socket:
             try:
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -804,7 +804,6 @@ def close(self):
     def _stop_vsc_message_loop(self):
         self.set_exit()
         self.loop.stop()
-        self.event_loop_thread.join(WAIT_FOR_THREAD_FINISH_TIMEOUT)
```
MikhailArkhipov: Is there way to perform wait and socket disconnect on a background thread rather than waiting here?

PR #424, opened 2018-05-21
Fix #325: Debuggee no longer terminates

### LINKED ISSUE #325: Debuggee no longer terminates

int19h:
Terminate the process immediately instead of closing pydevd socket when disconnecting after launch.

(fixes #325)

### DIFF
--- a/ptvsd/_vendored/force_pydevd.py
+++ b/ptvsd/_vendored/force_pydevd.py
@@ -1,6 +1,7 @@
+from importlib import import_module
 import warnings
 
-from . import check_modules, prefix_matcher, preimport
+from . import check_modules, prefix_matcher, preimport, vendored
 
 
 # Ensure that pydevd is our vendored copy.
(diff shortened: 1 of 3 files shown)


ptvsd/_vendored/force_pydevd.py
```diff
@@ -13,6 +14,16 @@
     warnings.warn(msg + ':\n {}'.format('\n  '.join(_unvendored)))
 
 
+# Constants must be set before importing any other pydevd module
```
int19h: Not related, but gets rid of the Cython warning (again).

ptvsd/wrapper.py
```diff
@@ -853,7 +848,9 @@ def _handle_disconnect(self, request):
         self.disconnect_request_event.set()
         self._notify_disconnecting(not self._closed)
         if not self._closed:
-            self.close()
+            # Closing the socket causes pydevd to resume all threads,
+            # so just terminate the process altogether.
+            sys.exit(0)
```
ericsnowcurrently: Won't this prevent re-attach?

int19h: It will, but in this scenario (launch), there's not supposed to be one - disconnect after launch means terminate debuggee, or at least that's how we have always implemented it so far in the old debugger.

ericsnowcurrently: Ah, right, ``on_disconnect()`` doesn't call ``_handle_disconnect()`` in the "attach" case.

Issue #422, opened 2018-05-16
Pull the latest pydevd changes from master


karthiknadig:

karthiknadig:
PR is ready. Pending code scan, estimated to be done by June 6th.

karthiknadig:
Moving this to June. since this wont be done till June  6th

PR #426, opened 2018-05-21
Update pydevd to 1.2

karthiknadig:
#422

Pending: regenerate the cpython files after merging. Probably will do this as a separate PR.


int19h:
Note that we'll need code scan to be complete before merging it.

karthiknadig:
@zooba is the code scanning done?

zooba:
@karthiknadig The scan is done, just waiting on an updated TPN from legal and distribution approval from management.

Issue #358, opened 2018-04-16
Performance hit debugging Django code


zooba:
While debugging Django templates, @qubitron noticed a big slowdown after adding a breakpoint. Every request after that was much slower than before.

We suspect this is because of tracing switching on due to the breakpoint, but not switching off. It also may be because tracing then goes through a lot of stdlib/Django code that it doesn't really need to trace.

qubitron:
More info, my specific repro steps are to F5 this app:
https://github.com/qubitron/stackoverflow-django

Browse to the ```/frameworks?language=Python``` page, results (using stopwatch app):
- OLD Debugger, first time load: 21s
- OLD Debugger, refresh page: 13s
- OLD debugger, run from first breakpoint to second: 13s
- OLD debugger, run from first breakpoint to second with just my code disabled: 13s
- OLD debugger, unset breakpoints and refresh: 13s
- NEW debugger, first time load: 12s
- NEW debugger, refresh page: 9s
- NEW debugger, run from first breakpoint to second: 26s
- NEW debugger, unset breakpoints and refresh: 27s

I had previously seen a greater difference between the two debuggers in the past, a 2-3x speedup on the new debugger for this exact same scenario with the page loading in about 5s on the new debugger, which makes me think something has been introduced that is causing a performance slowdown.

int19h:
Still investigating, but it appears to have something to do specifically with setting breakpoints while the app is running. If breakpoints are set before F5, then they don't affect perf substantially unless it's the same function/frame (and even then the slowdown is nowhere near as big). But if a breakpoint is set after, things slow down substantially.

This is regardless of where the breakpoint is (i.e. it can be in a different file that's not even loaded). It also applies to all code, not just Django; but Django templates are the worst affected, which is likely why you've noticed it there.

int19h:
Okay, so at this point I'm pretty sure that this is not a regression on our side, or that we're using pydevd in the wrong way - I have a condensed simplied repro showing similar results in LiClipse on stock pydevd (with none of our changes). @qubitron, can you give that a try as well, just to make sure that we're looking at the same issue here?

fabioz:
I actually started to take a look at this and my initial guess is not correct. This isn't really django-specific... to make it clear, there's still one issue with breakpoints on django templates in that it won't allow frames to be skipped, but this problem isn't really related to it.

[...]

PR #429, opened 2018-05-23
By default don't use frame evaluation as it conflicts with tracing fr…

fabioz:
…ames.

Partial fix for #358 (a final fix must take into account that the frame
evaluation and tracing debugger could be both working at the same time and avoid
a slowdown -- possibly by completely shutting down the frame evaluation
debugger).

Also includes a minor performance improvement when tracing frames.


ptvsd/_vendored/pydevd/_pydevd_frame_eval/pydevd_frame_eval_main.py
```diff
@@ -10,7 +10,7 @@
 enable_cache_frames_without_breaks = None
 dummy_trace_dispatch = None
 
-USE_FRAME_EVAL = os.environ.get('PYDEVD_USE_FRAME_EVAL', None)
+USE_FRAME_EVAL = os.environ.get('PYDEVD_USE_FRAME_EVAL', 'NO') # NO means don't use, anything else, use.
```
karthiknadig: >don't use, anything else, use [](start = 74, length = 29)

incomplete comment?

PR #449, opened 2018-05-30
Update cython with latest changes.

fabioz:


fabioz:
@karthiknadig this is a pull request to update the cython modules (which is needed after the latest changes I did).

It seems that it also updated a number of other places... so, should I actually be providing such a pull request or is this done as a part of the release process?

karthiknadig:
@zooba you had done this before. do you have any recommendation? 

Note that we may have update these files again after we are done merging #426

zooba:
I don't remember whether we regenerate the C file as part of sdist build, but pretty sure we're rebuilding it before producing the new pyd.

karthiknadig:
Since we will be taking the update to pydevd 1.2 in the next milestone. I am ok with taking this change. I will be merging this by noon. Let me know if there are any issues or reasons to block this.

PR #463, opened 2018-06-04
Clean up session handling in the daemon.

### LINKED ISSUE #459: Break on normal exit is not respected
Using Visual Studio
Enable Break on normal exit option

Result: Console does not wait for input at end of program

ericsnowcurrently:
This is part of the effort to fix gh-459.  While it doesn't fix it, it does clean up the Daemon code a little and resolves a potential issue with session lifecycle handling.  The solution for gh-459 will build on this.

### DIFF
--- a/ptvsd/__main__.py
+++ b/ptvsd/__main__.py
@@ -6,7 +6,7 @@ import argparse
 import os.path
 import sys
 
-from ptvsd._main import debug_main, run_main
+from ptvsd._local import debug_main, run_main
 from ptvsd.socket import Address
 from ptvsd.version import __version__, __author__  # noqa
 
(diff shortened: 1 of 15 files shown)


huguesv:
This didn't fix it for me :(

ericsnowcurrently:
How does PTVS use ptvsd?  Does is call the ptvsd CLI?  If so, does it run it as a server or a client?

huguesv:
From VS, we start up a process: `python.exe ptvsd_launcher.py <bunch of args>`, which calls into `ptvsd.debug()`
https://github.com/Microsoft/PTVS/blob/master/Python/Product/PythonTools/ptvsd_launcher.py#L111

karthiknadig:
PTVS uses it as a client.

ericsnowcurrently:
@huguesv Would you mind trying it again?

karthiknadig:
Tried this. It is not fixed yet.

ericsnowcurrently:
I'm going to merge this and then go from there.

karthiknadig:
Ok, go ahead

PR #467, opened 2018-06-06
Just my code

karthiknadig:


ptvsd/wrapper.py
```diff
@@ -1914,6 +1962,18 @@ def on_pydevd_thread_suspend(self, seq, args):
...
+        xframes = list(xml.thread.frame)
+        xframe = xframes[0]
+        filepath = unquote(xframe['file'])
+        if reason in STEP_REASONS or reason in EXCEPTION_REASONS:
+            if not self._should_debug(filepath):
+                self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
+                return
+            for filename in IMPORTLIB_BOOTSTRAP:
+                if filepath.endswith(filename):
+                    self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
```
fabioz: This should not be needed after I integrate the latest changes from pydevd to handle (see: https://github.com/fabioz/PyDev.Debugger/blob/master/_pydevd_bundle/pydevd_frame_utils.py#L44)

ptvsd/wrapper.py
```diff
@@ -1035,6 +1054,7 @@ def _parse_debug_options(self, debug_options):
             DJANGO_DEBUG=True|False
             WINDOWS_CLIENT=True|False
             DEBUG_STDLIB=True|False
+            JUST_NY_CODE=True|False
```
fabioz: typo: JUST_NY_CODE -> JUST_MY_CODE

ptvsd/wrapper.py
```diff
@@ -1864,6 +1886,32 @@ def version_str(v):
...
+    def on_setDebuggerProperty(self, request, args):
+        jmc = int(args.get('JustMyCodeStepping', 0))
+        self.debug_options['JUST_MY_CODE'] = jmc > 0
+        self.send_response(request)
+
+    def _is_stdlib(self, filepath):
+        for name in IMPORTLIB_BOOTSTRAP:
+            if filepath.endswith(name):
+                return True
+        for prefix in STDLIB_PATH_PREFIXES:
```
fabioz: If `IDE_PROJECT_ROOTS` and `LIBRARY_ROOTS` are properly set (after I'm able to integrate my latest changes) this could be handled by pydevd directly).

-- I plan on providing a pull request on Monday to address it as well as the other things I've worked on now that the pull request to move pydevd to 1.2 was integrated.

ptvsd/wrapper.py
```diff
@@ -127,28 +128,10 @@ def using_format(self, fmt):
...
-    ('ptvsd', 'wrapper.py'),
-    ('ptvsd', '_main.py'),
-    ('ptvsd', '_version.py'),
-    ('ptvsd', '__init__.py'),
-    ('ptvsd', '__main__.py'),
-]))
+PTVSD_DIR_PATH = os.path.dirname(__file__)
+DONT_TRACE_FILES = set((os.path.join(*elem) for elem in
+                        (('ptvsd', f) for f in os.listdir(PTVSD_DIR_PATH)
+                        if f.endswith('.py'))))
```
zooba: This is far less robust than you think. Better to do `import ptvsd` and then exclude any frame where its filename is within any of the elements of `ptvsd.__path__` (accounting for `normcase`, of course).

The aim is to avoid hitting the file system at all, and make no assumptions about whether the files actually exist on disk (or whether they are `.py` and not `.pyc` etc.)

ptvsd/wrapper.py
```diff
@@ -1864,6 +1866,51 @@ def version_str(v):
...
+        if self.debug_options.get('JUST_MY_CODE', False):
+            project_dirs = []
+            for path in sys.path + [os.getcwd()]:
+                is_stdlib = False
+                norm_path = os.path.normcase(path)
+                for prefix in STDLIB_PATH_PREFIXES:
+                    if norm_path.startswith(prefix):
+                        is_stdlib = True
+                        break
+                if path.endswith('ptvsd') or path.endswith('site-packages'):
```
zooba: How do these get hit without triggered the `STDLIB_PATH_PREFIXES` check? (And why not just put our `ptvsd` path in `STDLIB_PATH_PREFIXES` if it's because it may be bundled?)

ptvsd/wrapper.py
```diff
@@ -1864,6 +1866,51 @@ def version_str(v):
...
+                               '\t'.join(project_dirs))
+
+    def _is_stdlib(self, filepath):
+        for name in IMPORTLIB_BOOTSTRAP:
+            if filepath.endswith(name):
+                return True
+        for prefix in STDLIB_PATH_PREFIXES:
+            if prefix != '' and filepath.startswith(prefix):
+                return True
+        for name in DONT_TRACE_FILES:
```
zooba: Why not put `IMPORTLIB_BOOTSTRAP` values into `DONT_TRACE_FILES`?

ptvsd/wrapper.py
```diff
@@ -1864,6 +1866,51 @@ def version_str(v):
...
+            if filepath.endswith(name):
```
zooba: If `DONT_TRACE_FILES` is turned into a `tuple` at some point (once we know it's not going to be modified again) you can just do `if filepath.endswith(DONT_TRACE_FILES):` here for a (minor) perf improvement.

ptvsd/wrapper.py
```diff
@@ -1940,7 +1996,6 @@ def on_pydevd_thread_suspend(self, seq, args):
         if reason == 'exception':
             # Get exception info from frame
             try:
-                xframes = list(xml.thread.frame)
                 xframe = xframes[0]
```
zooba: if `xframes` is still valid here then so is `xframe`

ptvsd/wrapper.py
karthiknadig: The site-packages in the roaming directory is hard to detect, hence using explicit site-packages check (`C:\Users\kanadig\AppData\Roaming\Python\Python36\site-packages`). I had multiple, `ptvsd` folders, including one set via PYTHONPATH. This was one of the ways to get rid of all of them.

ptvsd/wrapper.py
karthiknadig: I used this list to tell pydevd to skip tracing on these files, in the frame tracing call back. See https://github.com/Microsoft/ptvsd/blob/master/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py#L730
Are you suggesting we do this manually be removing frames when we suspend threads?

ptvsd/wrapper.py
karthiknadig: IMPORTLIB_BOOTSTRAP is no longer needed. I will remove that.

ptvsd/wrapper.py
karthiknadig: The don't trace files is supposed to be updated once when ptvsd loads. So, we should be able to use this.

ptvsd/wrapper.py
```diff
@@ -1908,12 +1955,21 @@ def on_pydevd_thread_suspend(self, seq, args):
...
         EXCEPTION_REASONS = {
             pydevd_comm.CMD_STEP_CAUGHT_EXCEPTION,
             pydevd_comm.CMD_ADD_EXCEPTION_BREAK
         }
 
+        xframes = list(xml.thread.frame)
+        xframe = xframes[0]
+        filepath = unquote(xframe['file'])
+        if reason in STEP_REASONS or reason in EXCEPTION_REASONS:
+            if not self._should_debug(filepath):
```
fabioz: Is this really needed? In which cases do you expect `ptvsd` to do this auto-run which `pydevd` is not doing out of the box? 

-- the cases not supported should probably be bugs against `pydevd` as doing this auto-run is much slower than doing the skip on `pydevd` directly.

karthiknadig: This is needed for Step over and Step out. For example: 
```python
import unittest

class MyTest(unittest.TestCase):
    def test_0(self):
        pass

if __name__ == '__main__':
    unittest.main()
```

Set breakpoint on line with `pass`. Then do step over, it will enter into `case.py` in `unittest`.

fabioz: Ok, got it... this really needs a `just my code` flag globally in `pydevd` or 2 new commands (to step over my code and step return to my code) -- can you create an issue for that (and mark as `upstream-pydevd`)?

Also, given this auto-run in the example below:

```
import unittest

class MyTest(unittest.TestCase):
    def test_0(self):
        pass  # break here

if __name__ == '__main__':
    try:
        unittest.main()
    except SystemExit:
        print('finish')
```

a `step over`/`step out` should end up in `print('finish')` right? -- I think that with the current code it'd finish the program as you're just making it run and not stop at the next possible location as would be expected (or maybe I missed something?) -- if that's the case, I think the auto-run should be disabled (and the support for step over/step return would have to wait for that feature to be implemented in the `pydevd` side).

karthiknadig: You are correct, currently we just continue and will not end up in `print('finish')`. I will remove the auto run code.

karthiknadig: Created a issue for step over/step return just my code : #477

ptvsd/wrapper.py
zooba: No, just suggesting changing the comparison function to do a `.startswith()` instead of `in` comparison.

ptvsd/wrapper.py
zooba: `site.getusersitepackages()` and `site.getsitepackages()` are your friends here.

karthiknadig: Ah! it is get`user`sitepackages. I missed that one. I will update this.

ptvsd/wrapper.py
```diff
@@ -1224,6 +1215,33 @@ def _process_debug_options(self, opts):
...
+
+        Note: for now we consider DEBUG_STDLIB = False as just-my-code.
+        """
+        dbg_stdlib = self.debug_options.get('DEBUG_STDLIB', False)
+        return not dbg_stdlib
+
+    def _apply_code_stepping_settings(self):
+        if self._is_just_my_code_stepping_enabled():
+            project_dirs = []
+            for path in sys.path + [os.getcwd()]:
```
DonJayamanne: I would change the condition to as follows:
```python
if path.endswith('ptvsd'):
    is_stdlib = True
else:
    for prefix in STDLIB_PATH_PREFIXES:
        if norm_path.startswith(prefix):
            is_stdlib = True
            break
```

ptvsd/wrapper.py
```diff
@@ -2038,6 +2073,7 @@ def on_pydevd_thread_suspend(self, seq, args):
         try:
             vsc_tid = self.thread_map.to_vscode(pyd_tid, autogen=False)
         except KeyError:
+            self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
```
DonJayamanne: Why is this necessary? If `pydevd` is sending some events, why are we calling back into PyDevD?

DonJayamanne: Feels like some workaround/hack for something.

karthiknadig: This is a case we we got a suspended event before create. This `try` catch was there to handle that issue. But there was another problem. If we don't have a reference for a thread, we would just return from this event. That means, user has no clue why their code is not running. 

I don't think we need to do the check any more, since we have the fix for the thread creation event. I will remove the try/catch. if this fails again then we have bigger problems.

ptvsd/wrapper.py
karthiknadig: will do

Issue #446, opened 2018-05-30
Stepping through Python C Extension


saulshanabrook:
First off, thank you for a wonderful experience stepping through Python code.

I am debugging a Python library that has some parts written in C (Numba). I notice that if I am in the debugger, I am not able to step through the parts written in C.

I would like to be able to do this, because these are often the most confusing parts, where having a debugger would be the most helpful.

I imagine this would be difficult to do, but I is it possible? I see there is some documentation for debugging C++ code in VS Code, so I wonder if that could somehow be combined with the Python debugger. 

I wasn't sure if I should open this here or https://github.com/Microsoft/vscode-python. If it's better suited there I can open a new issue.

ericsnowcurrently:
@zooba, know anything about this relative to VS Code?

int19h:
This requires a debugger that can handle both simultaneously. We do have such a thing for VS, but not for VSCode (it requires a native debugger that is mixed-mode-capable for us to hook into, and so far only VS has one).

https://docs.microsoft.com/en-us/visualstudio/python/debugging-mixed-mode-c-cpp-python-in-visual-studio

fabioz:
As a note, sometimes an easier setup is starting the debugger on python and then use the 'attach to' from the C++ debugger. 

Although if you're using the numba jit and want to debug that, I think you'll have a hard time debugging that in the C++ debugger too (may be easier to remove the `jit` declaration temporarily and debug in pure Python and then put the `jit` declaration back).

saulshanabrook:
> Although if you're using the numba jit and want to debug that, I think you'll have a hard time debugging that in the C++ debugger too

I am actually debugging numba's source code, not a jitted function.

> As a note, sometimes an easier setup is starting the debugger on python and then use the 'attach to' from the C++ debugger.

OK I will try that, thank you!

Issue #114, opened 2018-02-23
Break on uncaught exceptions not working when exception is raised from a background thread


int19h:
Python 3.6 32-bit + VSC
```py
import threading

def thread_func(n):
    raise Exception('argh')

th = threading.Thread(target = lambda: thread_func(1))
th.setDaemon(True)
th.start()

th = threading.Thread(target = lambda: thread_func(2))
th.setDaemon(True)
th.start()

#raise Exception('argh')
x = input("?")
print(x)
```
Set checkbox on "Uncaught exceptions", clear one on "Raised exceptions", and run. Note that exception is printed out to stdout, but there's nothing in VSC. Now uncomment the `raise` before `input` and run again - that one gets reported.

huguesv:
Is there a repro for this?

int19h:
I've updated the item with code.

Also, this behavior is identical in PyDev as well, so it's a pydevd limitation, and we'll need to fix it in there.

fabioz:
I started to take a look at this issue... Initially I thought it'd be possible to just use the hooks for thread creation, but this isn't really possible (it wouldn't work if the debugger is attached to a running program and threads don't really raise their exceptions) so, changing the approach to use the tracing facilities as is the case for handled exceptions.

The idea is that only a few frames have to be traced for the exception info to trace unhandled exceptions (others can remain untraced). 

Namely:

- `threading.py` needs to trace functions: `_bootstrap_inner` , `__bootstrap_inner`.
- a frame where `frame.f_back` has the pydevd.py file with the `run` function.
- a frame where `frame.f_back` is None.

fabioz:
I've finished implementing this feature on pydevd (on https://github.com/fabioz/PyDev.Debugger/commit/77819b14274814d865bac2f4debd09726bc3942e), but there are many conflicts to bring this to ptvsd because https://github.com/Microsoft/ptvsd/pull/426 is not integrated yet, so, instead of providing a pull request (which would make integrating https://github.com/Microsoft/ptvsd/pull/426 pretty difficult), I'll wait for that pull request to be integrated and then provide a pull request for this feature.

PR #471, opened 2018-06-11
Fixes issues dealing with handled and unhandled exceptions (#114, #446).

fabioz:


karthiknadig:
@ericsnowcurrently @DonJayamanne If there are no other comments, I will merge this EOD.

PR #475, opened 2018-06-12
Wait-on-exit before sending "disconnect" response.

ericsnowcurrently:
(fixes gh-459)

We must wait-for-user before the "disconnect" response is sent.  This PR fixes that.


ericsnowcurrently:
I have verified that this fixes the problem under PTVS.

PR #476, opened 2018-06-13
Fixes #338 (output is no longer buffered)

### LINKED ISSUE #338: Output redirection is slow
Older version of PTVSD would redirect the stdout and stderr as the data was written into those streams. In the new versions there's a delay of around 300ms.

This introduces a slight delay in:
* Running a program to completion
E.g. running a program with a simple print statement takes a noticeable amount of time to complete.
* Similarly, when running a program to completion without debugging we notice the same delay

Once again, this is because of the fact that the new debugger polls the `new output` buffers for data every 300ms.

Can we implement the `push` approach the older PTVSD version had, instead of the current `pull` approach. That would make the UX more responsive.

fabioz:


ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_io.py
```diff
@@ -6,25 +6,29 @@ class IORedirector:
...
     def write(self, s):
-        for r in self._redirectTo:
-            try:
-                r.write(s)
-            except:
-                pass
+        # The original one may fail if the user tries to write a number to the stream, etc.
+        self._redirect_to[0].write(s)
+        try:
+            # This one should not fail.
```
DonJayamanne: If this should not fail, the why is it in `try`?
Should the comment read `This should be allowed to fail`?

fabioz: The comment should actually be that it is not allowed to fail, but thinking about it a bit more I don't think that the except should be needed at all (if it fails it's probably an error and we want to fix it), so, removed that try..except.

karthiknadig:
I will merge this at 11:00 AM PDT if there are no other comments.

Issue #485, opened 2018-06-14
Do not display non-user code in debugger output


DonJayamanne:
When starting the debugger we display the message `pydev debugger: starting`.
Besides this isn't something that didn't happen in the past.

Raising this as we have had users in the past asking about why we display the other commands (in the terminal) that bootstrap the debugger.

<img width="253" alt="screen shot 2018-06-14 at 1 54 00 pm" src="https://user-images.githubusercontent.com/1948812/41437917-99e25310-6fda-11e8-89e2-42bc1876035b.png">

fabioz:
This is explicitly passed in to pydevd as the default is not printing that message (so, it should be a matter of removing `--print-in-debugger-startup` from the arguments passed to pydevd in `__main__.py`).

karthiknadig:
Here i am printing `sys.argv` from `pydevd.main()`. As you can see it is printing `pydev debugger: starting`, in both cases.
This is with `--print-in-debugger-startup`:
```
['C:\\GIT\\ptvsd\\ptvsd\\__main__.py', '--port', '52873', '--client', 'localhost', '--qt-support=auto', '--print-in-debugger-startup', '--file', 'c:\\GIT\\pyscratch\\scratch9.py']
pydev debugger: starting (pid: 10032)
pydev debugger: starting
```
This is without `--print-in-debugger-startup`:
```
['C:\\GIT\\ptvsd\\ptvsd\\__main__.py', '--port', '53013', '--client', 'localhost', '--qt-support=auto', '--file', 'c:\\GIT\\pyscratch\\scratch9.py']
pydev debugger: starting
```

PR #496, opened 2018-06-18
Fix for displaying non-user code

karthiknadig:
Fixes: #485


karthiknadig:
@fabioz This should fix the immediate issue in #485. Should there be a flag just like for DEBUG and WARN messages?

fabioz:
@karthiknadig I think the fix should be removing that line altogether as it's already printed before (it was really a bug -- probably slipped because in pydev it's very rare to not use the -multiproc mode).

Issue #448, opened 2018-05-30
Breakpoints don't hit on attach


karthiknadig:
Repro steps:
1. Copy the code below into test.py
2. Make sure you are synced to latest ptvsd master
3. set PYTHONPATH to ptvsd
4. In the command prompt run `py -3.6 test.py`
5. From VS/VSC open `test.py` file and set breakpoints on line 7 and 10.
6. Attach to `tcp://localhost:5678`

![image](https://user-images.githubusercontent.com/3840081/40738084-6a4d55d0-63f7-11e8-8aaf-1610c412ac26.png)

Expected: 
Breakpoint on line 7 is hit.

Actual:
Console displays `Hello! What is your name?` and waits for user Input. The break-point on line 10 is hit after entering some text.

Repro code:
```python
import random
import ptvsd

ptvsd.enable_attach()
print("test!!!!!!!!!!!!!!!!!!!! wait for attach")
ptvsd.wait_for_attach()
print('ptvsd version: %s' % ptvsd.__version__)
guesses_made = 0
name = input('Hello! What is your name?\n')
number = random.randint(1, 20)
print('Well, {0}, I am thinking of a number between 1 and 20.'.format(name))

while guesses_made < 6:
    guess = int(input('Take a guess: '))
    guesses_made += 1
    if guess < number:
        print('Your guess is too low.')
    if guess > number:
        print('Your guess is too high.')
    if guess == number:
        break
if guess == number:
    print('Good job, {0}! You guessed my number in {1} guesses!'.format(name, guesses_made))
else:
    print('Nope. The number I was thinking of was {0}'.format(number))
```

@DonJayamanne , @ericsnowcurrently for reference. This used to work in 4.1.1.a1

ericsnowcurrently:
I got the same result (bp on line 7 not hit) on linux using VSC.

ericsnowcurrently:
I switched PTVSD to the v4.1.1a1 tag and found that breakpoints were not hit at all.

Also, I noticed that under 4.1.1a1 I got the warning about Cython, but under master I didn't.

ericsnowcurrently:
After clearing all breakpoints and then setting the two again, I got the original behavior (bp on line 7 not hit, line 10 hit).  So this does not appear to be something new.

ericsnowcurrently:
[...]

PR #502, opened 2018-06-19
Wait for handshake completion when waiting for attach.

ericsnowcurrently:
(see gh-448)

There is a race between attach and handshake completion that may result in breakpoints not being hit.  This PR fixes that by waiting for handshake completion when waiting for attach.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -989,6 +989,8 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
         self.debug_options = {}
         self._statelock = threading.Lock()
         self._debugging = debugging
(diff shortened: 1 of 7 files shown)


ericsnowcurrently:
@DonJayamanne, I'll look at adding a test.

karthiknadig:
This repros for me even with the change on Windows.

ericsnowcurrently:
@karthiknadig, is that on VS?  The patch fixes the issue for me in VSC.

ericsnowcurrently:
Unless there are any objections, I going to land this PR as-is.  (It does improve the situation.)  Then we can address the deficiency in `CMD_SET_BREAK` separately.

PR #505, opened 2018-06-20
No longer raising UnsupportedPyDevdCommandError when stdin input is requested. Fixes #490

### LINKED ISSUE #490: Debugger doesn't stop at breakpoint
VS, Python 2.7, Just My Code ON or OFF

```
res = raw_input('>')
res = 'hello'
print(res) # < breakpoint on this line
```

F5
Press enter when asked for input in console
Breakpoint on print is not hit

fabioz:
What happened was that the debugger exited and stopped tracing because of the exception.


karthiknadig:
Verified this fixes the issue.

karthiknadig:
I will merge this EOD if no other comments.

Issue #506, opened 2018-06-20
KeyError in thread suspended handler in attach scenario.


karthiknadig:
This line raises an error if the thread is seen only at suspend.
https://github.com/Microsoft/ptvsd/blob/master/ptvsd/wrapper.py#L2184

Previously we were not reporting the thread  state leaving the thread suspended. In a later change we removed it since thread created or found were reported via created handler or threads request handler. For attach scenario, it is possible that we see the thread for the first time in the suspended handler. So, only for the attach case we should allow the thread to be added to the thread map (may be send a created event as well?) and report thread suspension.

ericsnowcurrently:
FWIW, there is code precedent in `on_thread()` for adding a previously unseen thread.

PR #507, opened 2018-06-20
 Fix for KeyError in suspended event handle during attach

karthiknadig:
#506

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -2144,17 +2144,12 @@ class VSCodeMessageProcessor(VSCLifecycleMsgProcessor):
                 self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
                 return
 
-        try:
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -2144,7 +2144,17 @@ def on_pydevd_thread_suspend(self, seq, args):
                 self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
                 return
 
-        vsc_tid = self.thread_map.to_vscode(pyd_tid, autogen=False)
+        try:
+            vsc_tid = self.thread_map.to_vscode(pyd_tid, autogen=False)
```
DonJayamanne: Remove try..except
```python
autogen = self.start_reason == 'attach'
vsc_tid = self.thread_map.to_vscode(pyd_tid, autogen=autogen)
```

karthiknadig: Will update

Issue #501, opened 2018-06-19
ptvsd.break_into_debugger not breaking at the right place


DonJayamanne:
```
import ptvsd
import time
ptvsd.enable_attach()
ptvsd.wait_for_attach()
time.sleep(2)
ptvsd.break_into_debugger()
print("Hello world")
```

* Run the above code
* Attach the debugger in VSC
* Debugger breaks at the wrong, place.
It breaks in `urllib/parse.py` line `745`:
```python
def quote(string, safe='/', encoding=None, errors=None):
    """quote('abc def') -> 'abc%20def'
```

This used to work in PTVSD 4.1.1a5
Most likely related to https://github.com/Microsoft/ptvsd/issues/462

DonJayamanne:
The commit d635e22d3396f1cf3ab2fe9bc6fc9f5395b38232 is what has introduced this bug.

/cc @karthiknadig @fabioz

fabioz:
I'll take a look at this.

fabioz:
Pull request https://github.com/Microsoft/ptvsd/pull/508 should fix the part related to pydevd.

Issue #479, opened 2018-06-14
Pausing (break) not working anymore in VSC


DonJayamanne:
We have unit tests that have caught this error. 

Been ignoring failing Debugger tests in VSC as we have had a number of failing debugger tests for over a month, hence I have no idea when this issue cropped up. Either way this is something that used to work (still works in the preview version of PTVSD on pypi).

<img width="771" alt="screen shot 2018-06-13 at 5 42 57 pm" src="https://user-images.githubusercontent.com/1948812/41385618-9e751bdc-6f31-11e8-9461-a32149d5b201.png">

**Steps to reproduce:**
* Create a python file with the following code:
```python
import time

for i in range(1000):
    time.sleep(1)
    print(i)
```
* Start debugging without any breakpoints
* Once started printing to console window, hit the `Pause` button.
* **Nothing happens**

DonJayamanne:
Looks like d635e22d3396f1cf3ab2fe9bc6fc9f5395b38232 has introduced this issue.

/cc @karthiknadig @fabioz

fabioz:
Pull request https://github.com/Microsoft/ptvsd/pull/508 should fix this too.

karthiknadig:
This is fixed. Tested on windows with VSC.

PR #508, opened 2018-06-20
Fix for #501 and #479 (issues related to pausing a thread with debugger.set_suspend)

fabioz:
This patch makes sure that:
- When a thread is suspended the stepping function is properly set (otherwise it may end up not breaking if the function was previously cached to be skipped).
- pydevd threads are not traced even if the tracing function is set on them.
- Changes settrace to let users pass where exactly the tracing should pause and use that API to make the pause from ptvsd as the previous approach had some issues if a given function had been previously skipped and currently had no tracing set.

### DIFF
--- a/ptvsd/attach_server.py
+++ b/ptvsd/attach_server.py
@@ -5,21 +5,13 @@
 import threading
 
 import pydevd
-# TODO: Why import these?
-from _pydevd_bundle.pydevd_custom_frames import (  # noqa
-    CustomFramesContainer, custom_frames_container_init,
-)
-from _pydevd_bundle.pydevd_additional_thread_info import (
-    PyDBAdditionalThreadInfo,
-)
 from _pydevd_bundle.pydevd_comm import (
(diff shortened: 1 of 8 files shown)


ptvsd/attach_server.py
```diff
@@ -90,11 +81,9 @@ def break_into_debugger():
...
-        additional_info = thread.additional_info
-    except AttributeError:
-        additional_info = PyDBAdditionalThreadInfo()
-        thread.additional_info = additional_info
-
-    debugger.set_suspend(thread, CMD_THREAD_SUSPEND)
+    import sys
+    pydevd.settrace(
+        suspend=True,
+        trace_only_current_thread=True,
```
DonJayamanne: Why is `trace_only_current_thread=True`
Why can't and shouldn't we be tracing other threads?

fabioz: `break_into_debugger` will only make the pause for the current thread. The method `enable_attach` is the one responsible for setting the tracing.

PR #513, opened 2018-06-21
GH-512: Fix single-frame stack trace handling.

### LINKED ISSUE #512: on_stackTrace() returns 0 frames when there is exactly 1.
In `on_pydevd_thread_suspend()` we store the current stack frames for later use in `on_stackTrace()`.  The actual object we store is the one we got by parsing the XML response.  It is an XML wrapper that *mostly* acts like a list.  However, if it is a single item then some list-like properties do not work right (in particular, `len(xframes)` returns 0).

The fact that `len(xframes)` (in `on_stackTrace()`) returns 0 is a problem.  Later in `on_stackTrace()` we track stack depth by that length.  So if there is only one frame (e.g. due to a breakpoint at the top-level of a script) then `on_stackTrace()` will response with 0 frames instead of 1.

FWIW, this may impact other code where we expect a list from XML results.

I'll have a PR up momentarily that fixes this.

ericsnowcurrently:
(fixes #512)

The XML library we are using treats a single-item list as a single item instead of a list.  This breaks `on_stackTrace()`.  This PR fixes the problem by converting `xframes` to a list before checking its size.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -1491,6 +1491,8 @@ class VSCodeMessageProcessor(VSCLifecycleMsgProcessor):
                 # This means the stack was requested before the
                 # thread was suspended
                 xframes = []
+            else:
+                xframes = list(xframes)
         totalFrames = len(xframes)
 
         if levels == 0:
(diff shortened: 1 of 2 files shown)


ptvsd/wrapper.py
```diff
@@ -1491,6 +1491,8 @@ def on_stackTrace(self, request, args):
                 # This means the stack was requested before the
                 # thread was suspended
                 xframes = []
+            else:
+                xframes = list(xframes)
```
karthiknadig: This should be done in the suspend handler. So it will be resolved for all current and future uses of the stack frames list.

ericsnowcurrently: fixed

PR #528, opened 2018-06-27
Add missing 'import os'.

fabioz:
Note: this broke with the latest logpoint integration (this path is not actually active right now because frame evaluation needs work before it can be made active).


fabioz:
Note that I've provided a fix for the flaky test failing at https://github.com/Microsoft/ptvsd/pull/529.

Issue #473, opened 2018-06-12
Unit tests in VSC for path translation for remote debugging fails on Linux


DonJayamanne:
https://github.com/Microsoft/vscode-python/issues/1939

I'll try to identify the root cause soon, but want to create this issue in PTVSD first.
This impacts remote debugging between linux and Windows OS.

DonJayamanne:
Narrowed down the issue further, seems to be a problem with Pydevd.
Looks like path translations are not taking place in pydevd.
Note: version of pydevd in 4.1.1a1 works, where as pydevd in master of PVTSD doesn't.

DonJayamanne:
@fabioz this started failing after we merged upstream PyDevD changes.
I.e. some change in PyDevD broke this, looks like path translations for remote debugging isn't working as expected.

fabioz:
@DonJayamanne I'm currently working in path translations as a part of https://github.com/Microsoft/ptvsd/issues/333, so, hopefully that'll fix this too.

fabioz:
@DonJayamanne can you check if https://github.com/Microsoft/ptvsd/pull/533 fixes it? If it doesn't, it'd be nice if you could print the paths received in `pydevd_file_utils.setup_client_server_paths` and also provide the output gotten by setting `pydevd_file_utils.DEBUG_CLIENT_SERVER_TRANSLATION` to `True` to help checking where the error may be on `pydevd`.

karthiknadig:
@DonJayamanne Did you get a chance to look at this?

DonJayamanne:
Will check this out and get back some time later today or first thing tomorrow.

DonJayamanne:
Cannot test as remote debugging is broken.

Issue #333, opened 2018-04-10
Wrong path separator in source path when working cross platform


karthiknadig:
Client is VS connecting to ptvsd running in linux:
Actual:
```
{"id": 2, "name": "\\usr\\local\\lib\\python3.5\\dist-packages\\django\\template\\base.render : 991", "source": {"path": "\\usr\\local\\lib\\python3.5\\dist-packages\\django\\template\\base.py", "sourceReference": 0}, "line": 991, "column": 1}
```

Expected:
```
{"id": 2, "name": "/usr/local/lib/python3.5/dist-packages/django/template/base.render : 991", "source": {"path": "/usr/local/lib/python3.5/dist-packages/django/template/base.py", "sourceReference": 0}, "line": 991, "column": 1}
```

DonJayamanne:
This may have something to do with the options WindowsClient=True.
Here PyDev is replacing all path separators with the windows path seprator `\` (pydevd_fileutils).

DonJayamanne:
I think its important to mention the scenario under which you're getting this issue (i.e. this is a `launch` scenario isn't it)

karthiknadig:
I don't see how launch or attach should have a difference in the stack trace source path.

DonJayamanne:
You might want to test attach, cuz I can't replicate this in VSC.

karthiknadig:
Tested with attach. When using path mappings the path separator for linux files is wrong. The issue occurs in both launch and attach.

int19h:
Looking at pydevd_file_utils.py, it seems that it tries to derive the path separator for the client (i.e. the IDE) from path mappings:

https://github.com/fabioz/PyDev.Debugger/blob/62b77e45ac173ec16d42130d6bde293e316bda56/pydevd_file_utils.py#L303

MikhailArkhipov:
@int19h  - link to pydevd?

fabioz:
Note: I'll start to take a look at this issue on Monday.

fabioz:
@karthiknadig I'm taking a look here and as @int19h mentioned, this is really expected right now.

I.e.: it's expected that the path translation fixes paths to what the client expects and there are translations which convert to the debugger when it receives a filename and it converts back when it sends the filename (this is in `pydevd_file_utils.norm_file_to_client` and `pydevd_file_utils.norm_file_to_server`).

[...]

PR #533, opened 2018-06-27
Fix issues related to path translations and case of files seen on client.

fabioz:
This patch makes improvements in the path handling in pydevd.

It doesn't completely fix https://github.com/Microsoft/ptvsd/issues/333 because the behavior there is expected (the path separators are always sent to be compatible with the client), still, it fixes the `CMD_LOAD_SOURCE` so that translation happens there properly and the case of files should now be correct on windows.

Hopefully it also fixes https://github.com/Microsoft/ptvsd/issues/473 (although I haven't actually tested that it does so).


fabioz:
Note that the test failing on travis seems to be a flaky test (already saw it fail other times and it doesn't seem related to this pull request).

PR #560, opened 2018-07-02
System tests test for listing and evaluation of variables

DonJayamanne:


tests/system_tests/test_variables.py
```diff
@@ -0,0 +1,182 @@
...
+
+            stacktrace = session.send_request("stackTrace", threadId=tid)
+            stacktrace.wait()
+            frame_id = stacktrace.resp.body["stackFrames"][0]["id"]
+            scopes = session.send_request('scopes', frameId=frame_id)
+            scopes.wait()
+            variables_reference = scopes.resp.body["scopes"][0]["variablesReference"] # noqa
+            variables = session.send_request('variables', variablesReference=variables_reference) # noqa
+            variables.wait()
+
```
karthiknadig: Can you ensure that the variables are sorted? We also need to ensure that sort order. Like a, b, c, before _a,_b,_c before __a,__b,__c before __a__,__b__,__c__.

DonJayamanne: Will create a separate test for that.
Else this test gets overloaded and difficult to maintain.

DonJayamanne: @karthiknadig 
Apart from that, is this OK

PR #591, opened 2018-07-04
Fixes deadlock when creating thread.

fabioz:
The issue is that there were some local imports during tracing which under some circumstances could deadlock the debugger (it was reproducible by doing: `from scapy import all as scapy `, albeit but in theory it could deadlock whenever any thread was created -- although I didn't find a way to consistently reproduce it as happened when using `scapy`).

This commit also moves the `pydevd` test files to a `resources` folder and adds a test related to path translation to `pydevd`.


karthiknadig:
I will merge this by EOD of there are no more comments.

Issue #603, opened 2018-07-06
Teminate event should not be send by PTVSD when VSC is restarting the debugger


DonJayamanne:
When restarting the debugger in VSC (using the `Restart` button), VSC sends the disconnect request with the args `restart=true`.

In such situations, PTVSD should not send the `terminate` event. 
Doing so, indicates to PTVSD that debugger has completely terminated.

Issue #530, opened 2018-06-27
PTVSD needs to wait for a little while before closing the socket after sending the last message


DonJayamanne:
## Environment data

- PTVSD version: Master
- OS and version: Mac
- Python version (& distribution if applicable, e.g. Anaconda): N/A
- Using VS Code or Visual Studio: VSC

Related to VSC issue https://github.com/Microsoft/vscode-python/issues/2064

## Actual behavior

* PTVSD shuts down even before VSC receives the entire packet

## Expected  behavior

PTVSD Should wait for some random time waiting for VSC to receive the message before terminating the socket.

## Possible solutions
Note: VSC has a slight delay of 100ms for such situations (e.g. if PTVSD dies, or user kills the PTVSD process manually, etc). In such cases VSC sends the terminated event.

**Solution 1:** We could do the same here, if disconnect request is received with a `restart` flag, PTVSD can wait for 100ms before killing the socket.
**Solution2:** 
* First send the response, then invoke `notify_disconnecting`
* This seems to work. Would prefer to still add the delay based on the `restart` flag in the disconnect request. A delay of 100ms.
```python
        self.send_response(request)
        self._notify_disconnecting()
        self._set_disconnected()
```
/c @ericsnowcurrently 

## Steps to reproduce:
1. Start debugging
1. Add a breakpoint
1. Click the `Restart` button in VSC
1. PTVSD receives the `disconnect` request
1. PTVSD sends the response and immediately disconnects

DonJayamanne:
@karthiknadig @ericsnowcurrently 
This is yet another reason to work on https://github.com/Microsoft/vscode-python/issues/1367

karthiknadig:
For solution 2 is it a custom message? If so then this won't work for VS.

DonJayamanne:
> For solution 2 is it a custom message? If so then this won't work for VS.

Does VS support `restart`?

karthiknadig:
No. It does not. I have to check with the Adapter Host folks about this. 
```
1> [DebugAdapter] --> C (disconnect-19): {"command":"disconnect","arguments":{"terminateDebuggee":true},"seq":19,"type":"request"}
```
On restart, they terminate the debuggee (see above message) and start another session.

ericsnowcurrently:
FWIW, I've landed a few PRs that improve the situation.  Apparently they do not quite solve the problem yet, though.  Per Don, the sys.exit() call blocks indefinitely.  I have not seen this issue in our tests, but who knows.

[...]

PR #604, opened 2018-07-06
Do not send terminated event when restarting

DonJayamanne:
Fixes https://github.com/Microsoft/ptvsd/issues/603

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -1099,7 +1099,6 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
         self._notify_ready()
 
     def on_disconnect(self, request, args):
-        args = request.get('arguments', {})
         self._restart_debugger = args.get('restart', False)
 
         # TODO: docstring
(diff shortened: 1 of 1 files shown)


DonJayamanne:
@karthiknadig

tests/system_tests/test_restart_vsc.py
```diff
@@ -0,0 +1,48 @@
...
+        cwd = os.path.dirname(filename)
+        debug_info = DebugInfo(filename=filename, cwd=cwd)
+
+        with self.start_debugging(debug_info) as dbg:
+            (_, req_launch, _, _, _, _) = lifecycle_handshake(
+                dbg.session, debug_info.starttype)
+
+            req_launch.wait()
+
+            dbg.session.send_request('disconnect', restart=False)
```
karthiknadig: Both cases restart = `False`?

tests/resources/system_tests/test_forever/launch_forever.py
```diff
@@ -0,0 +1,5 @@
+import time
+
+for i in range(10000):
```
karthiknadig: I feel we should make this loop forever. I don't want test to rely on this being 100 seconds long for cleanup.

tests/system_tests/test_restart_vsc.py
DonJayamanne: lol, bug, no way to test the tests, its disabled cuz of #530

Issue #620, opened 2018-07-09
Mix of unicode and str in filename conversion.


fabioz:
Will submit a fix shortly.

DonJayamanne:
Fixed

PR #621, opened 2018-07-09
Fix for #620: Mix of unicode and str in filename conversion.

fabioz:


karthiknadig:
Will merge this EOD if there are no other comments.

DonJayamanne:
* What are we trying to fix here?
The issue isn't very clear.
@karthiknadig I thought we decided not to merge any PyDevD changes unless absolutely necessary? Right now I don't see how this meets that criteria?

karthiknadig:
Ok. I will leave it open for now.

fabioz:
@DonJayamanne @karthiknadig 

The actual error this fixes is in the screenshot below (I think it may also fail if the user has filenames with non ascii chars) -- I got this from a user directly from e-mail (note that this is only on Python 2).

error cython str unicode

fabioz:
@DonJayamanne @karthiknadig just to complement, I introduced this bug in https://github.com/Microsoft/ptvsd/pull/533, and I think it's a critical fix for Python 2 (so, if https://github.com/Microsoft/ptvsd/pull/533 will be kept, it's critical that this is also applied).

PR #626, opened 2018-07-09
Switch from Awaitable to SentRequest.

ericsnowcurrently:
In tests we often need to wait for events and responses.  Recently we switch from using context managers for this to a new Awaitable class.  However, this makes the code a bit harder to follow.  This PR brings things back to a middle ground by:

* returning a new SentRequest object from `DebugSession.send_request()` instead of a dedicated "awaitable"
* adding `DebugSession.send_request_and_wait()`
* adding `t.h.debugsession.wait_all()` to replace Awaitable.wait_all()
* fixing up the tests to use `send_request_and_wait()`

### DIFF
--- a/tests/system_tests/__init__.py
+++ b/tests/system_tests/__init__.py
@@ -146,23 +146,23 @@ def lifecycle_handshake(session, command='launch', options=None,
                         breakpoints=None, excbreakpoints=None,
                         threads=False):
     with session.wait_for_event('initialized'):
-        req_initialize = session.send_request(
+        req_initialize = session.send_request_and_wait(
             'initialize',
             adapterID='spam',
         )
-    req_command = session.send_request(command, **options or {})
-    req_threads = session.send_request('threads') if threads else None
+    req_command = session.send_request_and_wait(command, **options or {})
+    req_threads = session.send_request_and_wait('threads') if threads else None
 
     reqs_bps = []
     reqs_exc = []
     for req in breakpoints or ():
         reqs_bps.append(
-            session.send_request('setBreakpoints', **req))
+            session.send_request_and_wait('setBreakpoints', **req))
     for req in excbreakpoints or ():
         reqs_bps.append(
-            session.send_request('setExceptionBreakpoints', **req))
+            session.send_request_and_wait('setExceptionBreakpoints', **req))
 
-    req_done = session.send_request('configurationDone')
+    req_done = session.send_request_and_wait('configurationDone')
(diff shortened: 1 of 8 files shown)


tests/system_tests/test_basic.py
```diff
@@ -50,19 +49,10 @@ def run_test_arguments(self, debug_info, expected_args):
...
-             ) = lifecycle_handshake(dbg.session, debug_info.starttype,
-                                     threads=True)
-
-            Awaitable.wait_all(req_launch,
-                               session.get_awaiter_for_event('thread'))
-            disconnect = session.send_request('disconnect')
-
-            Awaitable.wait_all(exited, terminated, disconnect)
+            with self.wait_for_exit(session):
+                lifecycle_handshake(session, debug_info.starttype,
```
DonJayamanne: Why are we not waiting for `exited`, `terminated`?

tests/system_tests/__init__.py
```diff
@@ -146,23 +146,23 @@ def lifecycle_handshake(session, command='launch', options=None,
...
-    req_threads = session.send_request('threads') if threads else None
+    req_command = session.send_request_and_wait(command, **options or {})
+    req_threads = session.send_request_and_wait('threads') if threads else None
 
     reqs_bps = []
     reqs_exc = []
     for req in breakpoints or ():
         reqs_bps.append(
-            session.send_request('setBreakpoints', **req))
+            session.send_request_and_wait('setBreakpoints', **req))
```
DonJayamanne: Why are we changing the behavior to be synchronous?
System tests should replicate real world, send multiple requests and receive then as they arrive, instead of send one request, wait for that to be processed. 
If we don't do that, I personally see no value in system tests.

DonJayamanne:
I'll review later

tests/system_tests/test_basic.py
ericsnowcurrently: That's what `self.wait_for_exit()` does.

MikhailArkhipov:
Sorry, test changes should wait until we get back to refactoring and test coverage. I am going to close the PR, we can discuss and reconsider later.

ericsnowcurrently:
While I will not merge this prior to EuroPython, I do plan on merging it.

BTW, it would probably be a good idea for us to create a release branch and cherry-pick commits, rather than block master.

karthiknadig:
Closing this PR, since there are several conflicts.

PR #627, opened 2018-07-10
Do not wait for settrace() to finish until wait_for_attach().

### LINKED ISSUE #545: ptvsd.enable_attach hangs until a debugger connects
## Environment data

- PTVSD version: Master
- OS and version: Mac

## Actual behavior

## Expected  behavior

Should work as it used to.

## Steps to reproduce:
* Start the program

ericsnowcurrently:
(fixes #545)

We move the `settrace()` into a thread, but make sure to wait for the thread to finish in `wait_for_attach()`.

Note that this PR is bare-bones and includes no tests.  For a more complete PR see #628.

### DIFF
--- a/ptvsd/_remote.py
+++ b/ptvsd/_remote.py
@@ -1,5 +1,6 @@
 import pydevd
 
+from ptvsd._util import new_hidden_thread
 from ptvsd.pydevd_hooks import install, start_server
 from ptvsd.socket import Address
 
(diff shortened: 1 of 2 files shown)


ericsnowcurrently:
Mikhail said to not create tests.

MikhailArkhipov:
We _can_ create tests but changing existing test bed, refactoring etc is not recommended unless trivial.

DonJayamanne:
> etc is not recommended unless trivial.

It is trivial, I can do this.

ericsnowcurrently:
Let's go with PR #628.

PR #628, opened 2018-07-10
Do not wait for settrace() to finish until wait_for_attach().

### LINKED ISSUE #545: ptvsd.enable_attach hangs until a debugger connects

ericsnowcurrently:
(helps #545)

We move the settrace() into a thread, but make sure to wait for the thread to finish in wait_for_attach().

My only concern with the solution is that the current thread will not get debugged if `enable_attach()` is called but `wait_for_attach()` is not (and no attach happens right away).  I'm going to address this in a follow-up PR by monkey-patching `sys.settrace` temporarily.

### DIFF
--- a/ptvsd/_remote.py
+++ b/ptvsd/_remote.py
@@ -13,6 +13,7 @@ def _pydevd_settrace(redirect_output=None, _pydevd=pydevd, **kwargs):
     # thread and all future threads.  PyDevd is not enabled for
     # existing threads (other than the current one).  Consequently,
     # pydevd.settrace() must be called ASAP in the current thread.
+    # See issue #509.
     #
     # This is tricky, however, because settrace() will block until
     # it receives a CMD_RUN message.  You can't just call it in a
--- a/ptvsd/_remote.py
+++ b/ptvsd/_remote.py
@@ -63,6 +63,10 @@ def enable_attach(address,
     t = new_hidden_thread('start-pydevd', start_pydevd)
     t.start()
 
+    def wait(timeout=None):
+        t.join(timeout)
+        return not t.is_alive()
+
     def debug_current_thread(suspend=False, **kwargs):
         # Make sure that pydevd has finished starting before enabling
         # in the current thread.
@@ -81,4 +85,4 @@ def enable_attach(address,
             **kwargs
         )
         debug('pydevd enabled (current thread)')
-    return daemon, t.join, debug_current_thread
+    return daemon, wait, debug_current_thread
--- a/ptvsd/_remote.py
+++ b/ptvsd/_remote.py
@@ -1,5 +1,6 @@
 import pydevd
 
+from ptvsd._util import debug
 from ptvsd.pydevd_hooks import install, start_server
 from ptvsd.socket import Address
 
@@ -16,6 +17,7 @@ def enable_attach(address,
                   _pydevd=pydevd, _install=install,
                   **kwargs):
     addr = Address.as_server(*address)
+    debug('installing ptvsd as server')
     # pydevd.settrace() forces a "client" connection, so we trick it
     # by setting start_client to start_server..
     daemon = _install(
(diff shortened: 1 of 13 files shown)


ericsnowcurrently:
It isn't clear how this worked before.  I expect it's been a while that `enable_attach()` has been an unexpected blocking operation (when waiting for the attach).  We just didn't have test coverage for it.

DonJayamanne:
I'll find the commit that broke it. 
I would like to know what broke it, i.e understand the issue and the code better. Did it work before by accident, or by design. If by design, then how did we end up adding a lock in a later change?

DonJayamanne:
@ericsnowcurrently 
Found the commit that introduced this bug, it is 68f3688e1d0ab81bc1efcee7d4f6fb687f50fc5a [Add breakpoints before CMD_RUN is issued. (#521)](https://github.com/Microsoft/ptvsd/pull/521),
please try to identify the root cause of this bug (introduced in the above PR)

ericsnowcurrently:
I'm going to hold off merging this until I've sort out the enable-attach-in-current-thread issue.  That situation is problematic (and not much of an improvement over enable-attach-blocks-until-attach).

DonJayamanne:
@ericsnowcurrently 
Can we try reverting the commit that broke this and work on a different solution for the original issue ([Add breakpoints before CMD_RUN is issued. (#521)](https://github.com/Microsoft/ptvsd/pull/521)).

fabioz:
@DonJayamanne @ericsnowcurrently I've a question: why is it important that there are 2 phases?

i.e.: `enable_attach()` and `wait_for_attach()`

I think most of the problems in the integration happen because this doesn't map well to how `pydevd` works (which has a single `settrace` which is meant to be used blocking the current thread until connected, get all breakpoints, set the tracing and only then resume to run the user code -- potentially adding a programmatic breakpoint).

Anyways, if this is needed, I think the solution is still what I had commented in https://github.com/Microsoft/ptvsd/issues/448#issuecomment-399412798: `wait_for_attach` shouldn't wait for the threading event and should instead ask for a command and wait for its confirmation after making sure all breakpoints and other initialization state is sent previously (i.e.: issue a CMD_LIST_THREADS and wait for it to be completed) and then do some experiments on sending the 'CMD_RUN' before breakpoints are ready (only the `pydevd` public API should be needed this).

Moving the tracing to be set in another thread is not good as it won't really stop in breakpoints in existing threads as it is.

fabioz:
> Whey can't we do this, attach the debugger on the current thread some how and then mimic the user continuing the debugger.



> I've a question: why is it important that there are 2 phases?
>     i.e.: enable_attach() and wait_for_attach()
> 
>     enable_attach: This spins up the debugger in the user code, but does not block. E.g. assume starting PTVSD in a flask app. The flask app can be running and a debug client is not attached to the user code.
>     Once user a debug client attaches, then yay, we're debugging. They can detach the debugger and user code will continue once again, and this can be done over and over.

Ok, but in this state, is the user expected to have breakpoints in place (or partial breakpoints or some parts of the debugger)? It's IMHO hard for the user to rationalize what's expected to be the state here until the `wait_for_attach` finishes.

> 
>     wait_for_attach: This needs to be called after spinning the debugger (after enable_attach and blocks user code until a debugger is attached.

fabioz:
>> Moving the tracing to be set in another thread is not good as it won't really stop in breakpoints in existing threads as it is now.
> 
> Whey can't we do this, attach the debugger on the current thread some how and then mimic the user continuing the debugger.

It should be possible (hopefully in https://github.com/Microsoft/ptvsd/pull/628#issuecomment-404243406 I explained how to do this issuing an early `CMD_RUN`, as long as there's proper bookkeeping on whether other needed commands have been sent before).

PR #632, opened 2018-07-10
WIP - Try unique port

DonJayamanne:
We have tests that seem to be holding on to port 9876.

I created this PR just to test whether we get any connection timeouts when using a different port (other than 9876). If we do not get any timeouts, then this proves the hypothesis that a process is created for testing is still holding onto port 9876.

Note: Previously the test framework for system tests, didn't properly dispose the resources that create socket servers, I have resolved that since then. It is most likely we have similar un-disposed resources in other tests (non-system tests).

Right now, its much easier to just use another port than identify what test is causing this issue, i know it isn't a system test.

### DIFF
--- a/tests/resources/system_tests/test_exceptions/handled_exceptions_attach.py
+++ b/tests/resources/system_tests/test_exceptions/handled_exceptions_attach.py
@@ -1,6 +1,6 @@
 import sys
 import ptvsd
-ptvsd.enable_attach((('localhost', 9876)))
+ptvsd.enable_attach((('localhost', 9879)))
 ptvsd.wait_for_attach()
 
 try:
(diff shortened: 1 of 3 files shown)


DonJayamanne:
Closing and reopening to run tests again

DonJayamanne:
@karthiknadig 
Please review and approve, my hunch was right, the other tests are holding onto port 9876 and take a while to die out. This is a temporary work around.

ericsnowcurrently:
Did something change?  This wasn't a problem before, as far as I know.

ericsnowcurrently:
The system test machinery is designed to clean up resources, so if something isn't getting cleaned up we out to figure out why.

PR #634, opened 2018-07-11
Revert "Use sys.exit() instead of os.kill(). (#538)"

### LINKED ISSUE #618: stdout output not displayed in terminal
## Environment data

- PTVSD version: Master
- OS and version: Mac OS (but could be on other OS)
- Using VS Code or Visual Studio:N/A

## Actual behavior

Data written into `stdout` not displayed in Terminal

## Expected  behavior

Output should be displayed before program exits

## Steps to reproduce:
1. Python program:
```python
import sys

sys.stdout.write("This will not be printed")
```
2. Debug the above program in a terminal window, and it will not be displayed in the Terminal.

Commit introducing the bug bd390322a8e31dcdb318da95e108c82f7f3a4f97

### LINKED ISSUE #633: Running a program to completion displays `Terminated 15` in terminal
## Environment data

- PTVSD version: Master
- OS and version: Mac

## Actual behavior

Running program to completion displays `Terminated 15` in terminal window

## Expected  behavior

This should not happen

## Steps to reproduce:
1. Just run a program to completion

## Notes: 
* This is an indication of the fact that the program is actually being terminated by PTVSD instead of letting it die.
* Commit introducing the bug bd39032

DonJayamanne:
This reverts commit bd390322a8e31dcdb318da95e108c82f7f3a4f97.

Fixes #633 
Fixes #618 

@ericsnowcurrently 
I'm revering the above commit, as it didn't fix anything and it was introduced with an assumption that it was the right way to do things. There was no change to any tests, hence there was no difference, the original bug #530 was not resolved, and we have a two new issues as a result. Hence reverting it.

### DIFF
--- a/ptvsd/daemon.py
+++ b/ptvsd/daemon.py
@@ -5,7 +5,9 @@ import threading
 from ptvsd import wrapper
 from ptvsd.socket import (
     close_socket, create_server, create_client, connect, Address)
-from .exit_handlers import ExitHandlers, UnsupportedSignalError
+from .exit_handlers import (
+    ExitHandlers, UnsupportedSignalError,
+    kill_current_proc)
 from .session import PyDevdDebugSession
 from ._util import (
     ClosedError, NotRunningError, ignore_errors, debug, lock_wait)
@@ -82,7 +84,7 @@ class DaemonBase(object):
         self._wait_for_user = wait_for_user
         self._killonclose = killonclose
 
-        self._exiting = False
+        self._exiting_via_atexit_handler = False
 
         self._exithandlers = ExitHandlers()
         if addhandlers:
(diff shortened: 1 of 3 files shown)


ericsnowcurrently:
What issues were caused by this?  Also, if I remember correctly, using os.kill() means the atexit handler is not triggered, which means disconnect won't work right (since it is triggered via at exit).

DonJayamanne:
@ericsnowcurrently 

> is not triggered, which means disconnect won't work right (since it is triggered via at exit).

Well, all of the tests pass, hence I'm confortable reverting this change. This change introduced two other bugs.

> IIRC, using os.kill will break things.

It was broken to begin with, my suggestion. Fix the OS.kill separately and write tests to fix that. Right now there's absolutely no way to validate your claim, or there are no test to confirm that.

I'll be merging this, but will create an issue to address your concerns. Right now, this is more critical at least from what I can tell.

/cc @karthiknadig

DonJayamanne:
> What issues were caused by this?

Not sure what you mean, PR #530 is the culprit.
What's the root cause, no idea, all i know is #530 did not fix anything (no tests to prove it), it just introduced two new bugs #633, #618

ericsnowcurrently:
Right, #530 is not fixed yet.  The PR you are reverting was the first part of the fix.  I was interrupted before I could finish.  Reverting is fine, but I'll un-revert when I get back to #530 (taking care to address #618 and #633).

DonJayamanne:
Sorry, I've confused everyone, the PR was #538

DonJayamanne:
> I was interrupted before I could finish

How was that? I think we wanted this fixed all along last week & earlier.

ericsnowcurrently:
I was pulled onto #545 (which I'm still working on) since it is P0.  See https://github.com/Microsoft/ptvsd/issues/530#issuecomment-403152945.

DonJayamanne:
Agreed, that's after we had spent a week on this issue and I decided to go with something much quicker, a hack in VS Code. This way we make progress with other issues rather than getting bogged down on a single with no real progress.

PR #646, opened 2018-07-12
Increase connection timeouts in tests

### LINKED ISSUE #631: Failing test on Mac Agent on VSTS
These are random failures.
Related https://github.com/Microsoft/ptvsd/issues/597

screen shot 2018-07-10 at 3 34 39 pm

DonJayamanne:
Fixes #631
* PTVSD takes a while to load, hence having a timeout of 1.0 is not sufficient.
* Increased time to 5.0 as that'll give it plenty of time to load.
* why 5s instead of 2s, simple, why not 5s instead of 2s.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,4 @@ ENV/
 # PyDev
 .project
 .pydevproject
+.settings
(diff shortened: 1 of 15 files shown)


DonJayamanne:
@karthiknadig please review
Simple problem and simple fix to our connection woes.

Issue #617, opened 2018-07-09
Incorrect arguments passed into Python module (debugging using 'python -m <module>')


DonJayamanne:
## Environment data

- PTVSD version: Master
- OS and version: N/A
- Python version (& distribution if applicable, e.g. Anaconda): N/A
- Using VS Code or Visual Studio:N/A

## Actual behavior

Arguments passed into module is incorrect

## Steps to reproduce:
1. Create python module, e.g. name 'xyz'
2. Execute python module from cmd line with args `arg1`, `arg2`, `arg3`
Print `sys.argv` passed into module and value printed is `['-m', 'arg1', 'arg2']`
3. Now, debug the above from VS Code, the output of `sys.argv` is as follows:
`['xyz:', 'arg1', 'arg2']`

PR #655, opened 2018-07-13
Add file path to sys.argv when running as module

karthiknadig:
Addresses #617 

@fabioz I am not sure if this is the right fix for the problem. Let me know if I need to make changes in other places.

### DIFF
--- a/ptvsd/_vendored/pydevd/pydevd.py
+++ b/ptvsd/_vendored/pydevd/pydevd.py
@@ -43,7 +43,7 @@ from _pydevd_frame_eval.pydevd_frame_eval_main import frame_eval_func, stop_fram
 from _pydevd_bundle.pydevd_utils import save_main_module
 from pydevd_concurrency_analyser.pydevd_concurrency_logger import ThreadingLogger, AsyncioLogger, send_message, cur_time
 from pydevd_concurrency_analyser.pydevd_thread_wrappers import wrap_threads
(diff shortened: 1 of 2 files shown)


fabioz:
Seems like a good fix for me... can you just check what's expected correct when running a directory (containing a `__main__.py`)?

i.e.: just want to know if that assign should be a few lines below or not.

karthiknadig:
@fabioz I had to make a bunch of changes to handle the directory case. This is **NOT** final.
There is a slight difference in behavior depending on if `__init__.py` is present or not.

ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -974,11 +974,22 @@ def run(self, file, globals=None, locals=None, is_module=False, set_trace=True):
...
+                    if filename is None:
+                        sys.stderr.write("No module named %s\n" % file)
+                        return
+                    else:
+                        file = filename
             else:
                 file = filename
-            sys.argv[0] = file
+                if filename.endswith('__init__.pyc') or filename.endswith('__init__.py'):
+                    filename = filename.replace('__init__.py', '__main__.py')
```
DonJayamanne: If __main__.pyc exists and not the .py file, then we need to use the pyc counterpart.

karthiknadig:
@fabioz Can you review this?

karthiknadig:
@DonJayamanne Review this please?

Issue #663, opened 2018-07-16
Create test for issue #661


karthiknadig:
Test for #661

PR #668, opened 2018-07-17
Add thread count tests

karthiknadig:
#663


tests/system_tests/test_misc.py
```diff
@@ -22,3 +22,45 @@ def test_with_no_output(self):
...
+            }]
+        }]
+        with self.start_debugging(debug_info) as dbg:
+            session = dbg.session
+            with session.wait_for_event('stopped') as result:
+                lifecycle_handshake(
+                                    session, debug_info.starttype,
+                                    breakpoints=breakpoints,
+                                    threads=True)
+            event = result['msg']
```
DonJayamanne: You might want to add a delay of 1s, cuz today when you add a break point not all thread stop. So a 1s delay will ensure other thread to have hot the breakpoint and stopped.

PR #666, opened 2018-07-16
Revert "Do not wait for settrace() to finish until wait_for_attach().…

DonJayamanne:
… (#628)"

This reverts commit c816752aff96e89f7cae708d7cde7989fe1eb0e1.

PR #670, opened 2018-07-17
WIP: Revert "Revert "Do not wait for settrace() to finish until wait_for_attach(). (#628)" (#666)"

ericsnowcurrently:
(helps #545)

This reverts commit 80d86b961d2e548303aa59ec55b86801bbb50542.

Note that this still suffers from the caveat that if attach doesn't happen within a second (and `wait_for_attach()` isn't called) then the main thread will never have debugging enabled.  One way or another that will need to be addressed.

### DIFF
--- a/ptvsd/_remote.py
+++ b/ptvsd/_remote.py
@@ -1,7 +1,25 @@
 import pydevd
 
+from ptvsd._util import debug, new_hidden_thread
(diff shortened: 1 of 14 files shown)


DonJayamanne:
All tests are passing, yay.
Will close and re-open to run VSTS tests once again.

PR #673, opened 2018-07-17
Fix 72 break suspension policy

fabioz:
Note: built upon https://github.com/Microsoft/ptvsd/pull/671 to avoid conflicts (so, that one has to be merged first).

Changes can be seen at commit: https://github.com/Microsoft/ptvsd/commit/564135748e40b39680f42b70077c18ee481d774f (note that it has some test changes along to fix some flakiness on pydevd tests on travis).


fabioz:
Rebased with current master.

Note that this patch only adds the break suspension policy to breakpoints. Working on providing the stack trace of live threads.

DonJayamanne:
> remote debug threads allocate a new port to run

What's this? Please could you explain this change/feature?

fabioz:
> remote debug threads allocate a new port to run
> 
> What's this? Please could you explain this change/feature?

Sorry, I didn't explain it well... it should be:

- Tests which check remote debugging use a new port to run the test.

(i.e.: it's an improvement only for `pydevd` tests).

Issue #231, opened 2018-03-20
Ability to stop on the first line of user code


DonJayamanne:
This is VSC specific

zooba:
It’s also necessary for VS – if you start with the Step In or Over commands then we should break on the first line.

fabioz:
Will start to take a look at this.

The idea is creating a `CMD_STOP_ON_START` command for `pydevd` which has to be issued before the `CMD_RUN`.

fabioz:
Already done in `pydevd` (waiting for https://github.com/Microsoft/ptvsd/pull/692 to be merged before providing a pull request as changes were done on top of the contents in that pull request).

karthiknadig:
With PR #700 pydevd part of this is done. Following items remain:
- [ ] In ptvsd we need to handle the `StopOnEntry` debug option. (I have this ready and will send a PR after pydevd part is merged)
- [ ] VSC should covert the `stopOnEntry` launch json option to debug option for Launch.

PR #700, opened 2018-07-28
Provide a CMD_STOP_ON_START (#231), fix issues on path separator.

fabioz:
Other changes in the pull request:
- refactored `wait_for_breakpoint` in tests.
- by default, if no workspace roots are given, any file not considered to be a library is considered a project.


karthiknadig:
I tested this with VSC. Look good to me.

Issue #718, opened 2018-08-01
Create a whole suite of tests for enable_attach without using wait_for_attach


DonJayamanne:
Currently almost all system tests use `wait_for_attach` due to a bug in `enable_attach` the simulated the same behavior of `wait_for_attach`.

karthiknadig:
#740 adds couple cases for using `enable_attach` without `wait_for_attach`

- `enable_attach` with `is_attached` and `break_into_debugger`
- `enable_attach` and `break_into_debugger`

karthiknadig:
`break_into_debugger` tests and `disconnect` system tests cover this

Issue #711, opened 2018-07-31
ptvsd.break_into_debugger and ptvsd_wait_for_attach only work the first time


DonJayamanne:
These functions to do work during re-attaches of remote debuggers.

PR #726, opened 2018-08-01
Fix blocking enable_attach and breaking on the main thread in remote debugging

### LINKED ISSUE #545: ptvsd.enable_attach hangs until a debugger connects

DonJayamanne:
Fixes #545, #711, #712 
More tests to be written as part of #718 

Summary of solution:
* For remote attach, spin up pydevd (use `install`) 
* In a separate thread wait for connections (within a loop, to support multiples)
* Ensure only one connection is supported at a time
* Use a simple event to ensure we have a single session at a time (we won't accept more than one connection if even is set)
* Use a simple event to detect when a handshake has been successfully been done

image

@karthiknadig 
I've basically reverted the way `enable_attach` works.
This allows breaking into the current thread without any issues.
Not sure what else it breaks, lets see how the tests go.

Lets talk about the changes, hacks/solutions tomorrow.
Easy fixes, just need some simple locks!!


DonJayamanne:
@karthiknadig 
Please approve.

Issue #72, opened 2018-02-13
On stackTrace should provide stack trace for any live thread


karthiknadig:
https://github.com/Microsoft/PTVS/issues/3736

karthiknadig:
This is not done yet. We have following pending items:
1. A way to tell pydevd to suspend all threads on (breakpoint/exception/etc). There is currently a mechanim but that path does not use breakpoint_id's which we rely on.
2. Get a notification from pydev that all threads are suspended. Currently there is only a way to tell if individual threads are suspended. we need a way to tell if all threads are suspended.

fabioz:
The part related to pydevd is on my todo-list (my current item is path mapping issues across platforms and right after this deal with the thread suspension policy).

DonJayamanne:
@karthiknadig @fabioz Would be great if we have separate issues for those and linked them in here. That way we know what the pre-requisites are for this issue to be looked at. I.e. easier to track the status of this.

fabioz:
Starting to take a look at this.

fabioz:
After working a bit on how to provide a message saying that all threads are suspended I got into corner cases where this cannot be reliably done because to actually pause a thread some tracing event is required, but by pausing a thread it's possible that we get into situations where this never happens and waiting for such a message would not work (i.e.: it would effectively deadlock).

I'm posting an example below where this happens:

```
import threading
import time

go_on = threading.Event()
lock = threading.Lock()
counter = 0


def method():
    global counter
    with lock:
        counter += 1
    go_on.wait()
        

threads = [threading.Thread(target=method) for _ in range(3)]

for t in threads:
    t.start()

while counter < 3:
    time.sleep(.1)    

print('breakpoint with suspend_all here')  # Will never be able to suspend all threads if they reached go_on.wait() as no tracing event will be generated for the thread after reaching that point.

go_on.set()
```

So, not sure how to proceed... the feature requested seems to be getting the stack of threads even if they're not paused, so, maybe the best approach would be actually implementing that feature (creating a `GET_THREAD_STACK` command which can be used to get the stack for any running thread by using `sys._current_frames`).

@karthiknadig do you think that's reasonable?

karthiknadig:
@fabioz We will need the `suspend_policy` in `CMD_SET_BREAK` (currently it is only available if we don't use `breakpoint_id`). That along with `GET_THREAD_STACK` should be enough.

fabioz:
@karthiknadig ok, will work on that. 

[...]

PR #731, opened 2018-08-03
Turning suspend all policy and using get stack trace command

karthiknadig:
This change does the following:
1. Enables the suspend all policy
2. Pause/continue will always suspend or resume all threads
3. Stack trace request uses CMD_GET_THREAD_STACK
4. exception info is extracted on request.

This should address most of the concerns with #72.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -2082,7 +2082,7 @@ class VSCodeMessageProcessor(VSCLifecycleMsgProcessor):
                 self.bp_map.remove(pyd_bpid, vsc_bpid)
 
         cmd = pydevd_comm.CMD_SET_BREAK
-        msgfmt = '{}\t{}\t{}\t{}\tNone\t{}\t{}\t{}\t{}'
+        msgfmt = '{}\t{}\t{}\t{}\tNone\t{}\t{}\t{}\t{}\tALL'
         if needs_unicode(path):
             msgfmt = unicode(msgfmt)   # noqa
         for src_bp in src_bps:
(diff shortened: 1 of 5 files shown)


ptvsd/wrapper.py
```diff
@@ -1988,16 +1987,19 @@ def on_pause(self, request, args):
...
     def on_continue(self, request, args):
         # TODO: docstring
-        tid = self.thread_map.to_pydevd(int(args['threadId']))
-        self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, tid)
-        self.send_response(request)
+        # Always continue all threads.
+        # TODO: Replace this with resume all command after #732 is fixed
+        for pyd_tid in self.thread_map.pydevd_ids():
+            self.pydevd_notify(pydevd_comm.CMD_THREAD_RUN, pyd_tid)
+        self.send_response(request, allThreadsContinued=True)
```
DonJayamanne: Can't we have a single command that would tell pydevd to continue ALL threads. Feels a lot more efficient, and code world work better. 

Also should be done for suspend all threads also If not already done. I.e. when we got the issue button we should send a single request to pause ALL threads instead of sending a request for each these.

ptvsd/wrapper.py
```diff
@@ -2392,20 +2405,8 @@ def on_pydevd_thread_run(self, seq, args):
         pyd_tid, _ = args.split('\t')
         pyd_tid = pyd_tid.strip()
 
-        # Stack trace, active exception, all frames, and variables for
+        # All frames, and variables for
         # this thread are now invalid; clear their IDs.
-        with self.stack_traces_lock:
-            try:
```
DonJayamanne: 👌👌👌👌👌👌

ptvsd/wrapper.py
DonJayamanne: For instance pause will fail if a new thread is created in the interim (a race condition resulting in a bug). So technically it should be pydeb that does the necessary plumbing and iteration of threads.

karthiknadig: I have already discussed this with @fabioz see here https://github.com/Microsoft/ptvsd/issues/732. :)

PR #735, opened 2018-08-06
Properly add __exception__ on all frames on caught exceptions (fixes #679)

### LINKED ISSUE #679: After continuing on raised error the error-type received is always BaseException
## Environment data

- PTVSD version: master
- OS and version: Any
- Python version (& distribution if applicable, e.g. Anaconda): Any
- Using VS Code or Visual Studio: Any

## Actual behavior

image

## Expected  behavior

Should show ValueError

## Steps to reproduce:
```
def raise_some_error():
    raise ValueError('bad')

raise_some_error()
```
Debug with following options on:
image

* Debugger will break on line 2, the error will be `ValueError` 
* Continue debugging
* Debugger will break on line 5, the error will be 'BaseException'

/cc @fabioz

fabioz:
Other changes in the pull request:
- Rename: break_on_exceptions_thrown_in_same_context -> skip_on_exceptions_thrown_in_same_context.
- Make sure that __exception__ is removed when resuming.

### DIFF
--- a/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
+++ b/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
@@ -104,8 +104,9 @@ def get_exception_breakpoint(exctype, exceptions):
     return exc
 
 
-def stop_on_unhandled_exception(py_db, thread, additional_info, exctype, value, tb):
+def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
     from _pydevd_bundle.pydevd_frame import handle_breakpoint_condition, handle_breakpoint_expression
+    exctype, value, tb = arg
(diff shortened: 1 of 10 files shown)


karthiknadig:
@fabioz would it be better if there was a CMD_GET_EXCEPTION_DETAILS command that returned exception type, description, and origin stack? I see a potential problem here where we call CMD_GET_THREAD_STACK and that might not be the place of origin for the exception. Especially when we continue after a raised exception.

DonJayamanne:
Your suggestion makes perfect sense. It's better to have pydev do most of the work that would be required by the debugger protocol, rather than ptvsd to do some massaging. Makes it easier for pydev to implement the debugger protocol.

fabioz:
@karthiknadig Just added a `CMD_GET_EXCEPTION_DETAILS`.

Issue #723, opened 2018-08-01
Fix broken tests in test_breaking_into_raised_and_unhandled_exceptions


DonJayamanne:
_From @DonJayamanne on August 1, 2018 5:25_

Currently skipped

_Copied from original issue: Microsoft/vscode-python#2293_

## Repro steps
1. Remove the line skipping this test in `test_exception.py`
2. `set PYTHONPATH=C:\ptvsd`
3. Run test:
```
py -3.7 -m tests -v tests.system_tests.test_exceptions.PTVSDAttachModuleExceptionLifecycleTests.test_breaking_into_raised_and_unhandled_exceptions
```
4. Should pass.

The test currently fails because of the unexpected third thread stopped event.

karthiknadig:
The cause for this issue is same as #679.

karthiknadig:
Looks like there is another issue, exposed after fixing 679.
@fabioz. looks like we get uncaught exception event raised twice in some cases. These are the call stacks for one such case. see above for repro steps.
**First event: on raised (this seems correct)**
```
File "C:\\Python37\\lib\\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
File "C:\\Python37\\lib\\runpy.py", line 85, in _run_code
    exec(code, run_globals)
File "C:\\GIT\\ptvsd\\tests\\resources\\system_tests\\test_exceptions\\mypkg_attach_unhandled\\__main__.py", line 12, in <module>
    main()
File "C:\\GIT\\ptvsd\\tests\\resources\\system_tests\\test_exceptions\\mypkg_attach_unhandled\\__main__.py", line 8, in main
    raise ArithmeticError(\'Hello\')
File "C:\\GIT\\ptvsd\\ptvsd\\_vendored\\pydevd\\_pydevd_bundle\\pydevd_frame.py", line 146, in trace_exception
    self.handle_exception(frame, event, arg)
File "C:\\GIT\\ptvsd\\ptvsd\\_vendored\\pydevd\\_pydevd_bundle\\pydevd_frame.py", line 334, in handle_exception
    self.do_wait_suspend(thread, frame, event, arg)
File "C:\\GIT\\ptvsd\\ptvsd\\_vendored\\pydevd\\_pydevd_bundle\\pydevd_frame.py", line 134, in do_wait_suspend
    self._args[0].do_wait_suspend(*args, **kwargs)
```

**Second event: uncaught (i suspect is the wrong one)**
```
[...]

PR #739, opened 2018-08-08
Handled exceptions should notify only once

### LINKED ISSUE #724: Fix broken test test_breaking_into_raised_exceptions_only
_From @DonJayamanne on August 1, 2018 5:23_

Currently skipped

_Copied from original issue: Microsoft/vscode-python#2292_

karthiknadig:
Fixes #724, #723 
Enables several other tests.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -31,9 +31,9 @@ from xml.sax import SAXParseException
 import _pydevd_bundle.pydevd_comm as pydevd_comm  # noqa
 import _pydevd_bundle.pydevd_extension_api as pydevd_extapi  # noqa
 import _pydevd_bundle.pydevd_extension_utils as pydevd_extutil  # noqa
-import _pydevd_bundle.pydevd_frame as pydevd_frame # noqa
+import _pydevd_bundle.pydevd_frame as pydevd_frame  # noqa
 #from _pydevd_bundle.pydevd_comm import pydevd_log
-from _pydevd_bundle.pydevd_additional_thread_info import PyDBAdditionalThreadInfo # noqa
+from _pydevd_bundle.pydevd_additional_thread_info import PyDBAdditionalThreadInfo  # noqa
 
 from ptvsd import _util
 import ptvsd.ipcjson as ipcjson  # noqa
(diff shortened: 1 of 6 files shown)


ptvsd/wrapper.py
```diff
@@ -461,7 +461,7 @@ def get_break_mode(self, name):
     def add_exception_break(self, exception, break_raised, break_uncaught,
                             skip_stdlib=False):
 
-        notify_on_handled_exceptions = 1 if break_raised else 0
+        notify_on_handled_exceptions = 2 if break_raised else 0
```
DonJayamanne: Please use a constant for these magic numbers `1`, `0`, `2`
Previously i assumed these were booleans. Clearly these numbers have some special meaning.

tests/system_tests/test_web_frameworks.py
```diff
@@ -258,6 +262,9 @@ def run_test_with_handled_exception(self, debug_info, framework,
             web_client_thread.start()
             web_client_thread.join(timeout=1)
 
+            # Give enough time for all events to be received
+            time.sleep(3)
```
DonJayamanne: is there no way to query the events received and wait in a loop.
Might be better to do so, as I've done something similar in the past and have noticed such tests to fail occasionally (possibly when server is under load).

karthiknadig: I will change how we handle this.

DonJayamanne: Not mandatory, but good if we could

karthiknadig: Created this #755 to tackle the problem with web frame work tests.

karthiknadig:
@DonJayamanne I addressed the flaky system tests issue in #760 . Please approve this.

karthiknadig:
Due to a change in VSTS there are multiple `ptvsd-testing-automation` listed. I will remote the required requirement until this is settled. This changed 2 times today:
image

karthiknadig:
@DonJayamanne can you approve this?

karthiknadig:
Waiting for decision on should we use old ptvsd behavior or C# behavior, by default, for raised exceptions. Another item is, should we add a setting (in ptvsd), which allows control of this behavior via `debugOptions`. It may be useful in rare cases. The thing to consider would be should this setting be exposed to VSC/VS users.
@DonJayamanne let me know what gets decided here.

karthiknadig:
Closing this. We can re-visit this again later.

PR #706, opened 2018-07-31
Fix blocking enable_attach and breaking on the main thread in remote debugging

DonJayamanne:
Fixes #545, #711, #712 
More tests to be written as part of #718 

Summary of solution:
* For remote attach, spin up pydevd (use `install`) 
* In a separate thread wait for connections (within a loop, to support multiples)
* Ensure only one connection is supported at a time
* Use a simple event to ensure we have a single session at a time (we won't accept more than one connection if even is set)
* Use a simple event to detect when a handshake has been successfully been done

image

@karthiknadig 
I've basically reverted the way `enable_attach` works.
This allows breaking into the current thread without any issues.
Not sure what else it breaks, lets see how the tests go.

Lets talk about the changes, hacks/solutions tomorrow.
Easy fixes, just need some simple locks!!

PR #740, opened 2018-08-09
break_into_debugger fixes and tests

### LINKED ISSUE #712: ptvsd.break_into_debugger only works when attaching

karthiknadig:
Fixes: #714, #715, #716, #717
Fixes: #712 
Note: #706 had not fixed #712

### DIFF
--- a/ptvsd/attach_server.py
+++ b/ptvsd/attach_server.py
@@ -7,7 +7,7 @@ from ptvsd._local import run_module, run_file  # noqa
 from ptvsd._remote import (
     enable_attach as ptvsd_enable_attach, _pydevd_settrace,
 )
-from ptvsd.wrapper import debugger_attached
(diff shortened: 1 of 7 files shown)


ptvsd/attach_server.py
```diff
@@ -88,8 +88,12 @@ def break_into_debugger():
...
     """
-    if not debugger_attached.isSet() or not _enabled:
-        return
+    if is_launch():
+        if not is_attached():
+            return
+    else:
+        if not is_attached() or not _enabled:
+            return
     import sys
```
DonJayamanne: You can movie is_attached out and reduce the nesting of the conditions by one level.

karthiknadig: I have another PR for Server attach fix where this is addressed where this is addressed,

PR #758, opened 2018-08-16
for VS client only send single stopped event in multi-threaded case

### LINKED ISSUE #752: Incorrect frame for breakpoint
## Environment data

- PTVSD version: today's master
- OS and version: Win10
- Using VS Code or Visual Studio: Visual Studio

## Steps to reproduce:
1. Type this code and set breakpoint on line 9 as in the picture and start debugging.
```
import threading
from queue import Queue
import time

print_lock = threading.Lock()

def get_url(current_url):

    time.sleep(2)
    with print_lock:
        print("\nStarting thread {}".format(threading.current_thread().name))

def process_queue():
    while True:
        current_url = url_queue.get()
        get_url(current_url)
        url_queue.task_done()

url_queue = Queue()

url_list = ["https://www.google.com"]*5

for i in range(5):
    t = threading.Thread(target=process_queue)
    t.daemon = True
    t.start()

start = time.time()

for current_url in url_list:
    url_queue.put(current_url)

url_queue.join()

```

image

## Actual behavior

VS breaks on line 16.

## Expected  behavior

VS breaks on line 9.

karthiknadig:
Fixes #752

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -2340,7 +2340,7 @@ class VSCodeMessageProcessor(VSCLifecycleMsgProcessor):
         self.send_response(request)
 
     def is_vs_client(self):
-        """ Return True if the client is Visual Studio 2017 False otherwise
+        """ Return True if the client is 'Visual Studio' False otherwise
         """
         return self._client_id == 'visualstudio'
 
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -1019,6 +1019,12 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
 
         self._stopped = False
 
+        # These are needed to handle behavioral differences between VS and VSC
+        # https://github.com/Microsoft/VSDebugAdapterHost/wiki/Differences-between-Visual-Studio-Code-and-the-Visual-Studio-Debug-Adapter-Host # noqa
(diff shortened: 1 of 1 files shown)


karthiknadig:
For https://github.com/Microsoft/vscode-python/issues/2498 we could remove the VS restriction in this PR.  @DonJayamanne any suggestions?

ptvsd/wrapper.py
```diff
@@ -2326,6 +2339,11 @@ def on_setDebuggerProperty(self, request, args):
 
         self.send_response(request)
 
+    def is_vs_client(self):
+        """ Return True if the client is Visual Studio 2017 False otherwise
```
MikhailArkhipov: You may want just to say 'Visual Studio' since this code should be working in Dev17+

ptvsd/wrapper.py
```diff
@@ -2434,6 +2452,13 @@ def on_pydevd_thread_suspend(self, seq, args):
             exc_name, exc_desc, _, _  = \
                 self._parse_exception_details(resp_args, include_stack=False)
 
+        if self.is_vs_client():
+            if self._vs_ignore_stopped_events.isSet() and reason == 'pause':
```
int19h: This should be `is_set` (`isSet` is the older deprecated spelling, and I think it's already gone in 3.x?).

PR #760, opened 2018-08-16
Improve web framework test reliability

karthiknadig:


karthiknadig:
Will do a couple of more runs of these.

karthiknadig:
This seems to reduce the frequency of failures.

tests/system_tests/test_web_frameworks.py
```diff
@@ -18,6 +18,30 @@
...
 
 class WebFrameworkTests(LifecycleTestsBase):
+    def _get_url_from_output(self, session):
+        strings = (e.body['output'] for e in
+                   self.find_events(session.received, 'output'))
+        for s in strings:
+            matches = re.findall(re_link, s)
+            if len(matches) > 0 and len(matches[0]) > 0 and \
+                len(matches[0][0].strip()) > 0:
+                return matches[0][0]
```
huguesv: This might work without explicit `len` calls
```
if matches and matches[0] and matches[0][0].strip():
    return matches[0][0]
```

Issue #45, opened 2018-02-06
Tools tip for dictionary shows object hash


karthiknadig:
![image](https://user-images.githubusercontent.com/3840081/35876515-5e08541a-0b27-11e8-941b-1bda759e56e1.png)

fabioz:
As a note, I've started to take a look at this issue.

fabioz:
After investigating it, this will actually be a big change in `pydevd`

The main issue is that the API to get variables only provides users with a way to get a dictionary of key->value which will already be the final representation of the object (and what's shown to the user is later used to get the variable again).

To fix this, both the protocol as well as the internal API will need to be changed to allow for object identifiers to be given to variables so that the what's shown to the user does not need to contain that information (which means that the protocol has to be changed to send and receive such an id to get the variable and that mapping of ids->variables must be kept while the thread is suspended).

Given that this will touch a good number of places, my suggestion would be already fixing the other related issue which is being able to have a different formatter per-request and try to use json instead of xml as a payload for the responses which return variables.

My estimate is around 12 days to have this complete. I'll already start on it, but as it'll take a while until it's actually integrated let me know if there's something I should take a look first given that this will take longer than I anticipated.

On the good side, the solution is closer to what the debug adapter protocol expects, so, the integration in   `ptvsd` will need to do less work and formatters will be more flexible than now.

karthiknadig:
I will go over the issues and let you know.

DonJayamanne:
Though this is a dup of #1114, please note the values are more prominent in VSC, hence the issue is worse in VSC than in VS.
In VS you get the values only as tooltips, where as in VSC they are `always` visible.
The fact that it manifests it self differently in VSC could impact the priority of this issue

karthiknadig:
The thing is this is how `pydevd` formats dictionaries. When that part is fixed, it will fix this in all places. Do let us know if we need to bump the priority on this.

karthiknadig:
Fixed:
![image](https://user-images.githubusercontent.com/3840081/53442685-b84af380-39be-11e9-90c1-3d0fcec13d97.png)

![image](https://user-images.githubusercontent.com/3840081/53442698-bc771100-39be-11e9-8b55-dc2afecc1a84.png)

PR #764, opened 2018-08-21
Create command to enable pydevd to send data using http-procotol instead of by line with urllib quoting.

fabioz:
This is the first step to support json messages (to later address #45).

Will pause working on that to address other issues but it'd be nice to have that already integrated.

Also, this fixes an issue where enabling debug info (with a `PYDEV_DEBUG=True` env var) triggered a recursion of debug messages when the output was redirected (which is always the case for `ptvsd`).

### DIFF
--- a/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -598,6 +598,8 @@ class NetCommand:
         if IS_PY2:
             if isinstance(text, unicode):
                 text = text.encode('utf-8')
+            else:
+                assert isinstance(text, str)
         else:
             assert isinstance(text, str)
 
--- a/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -191,6 +191,7 @@ CMD_SET_PROJECT_ROOTS = 202
 
 CMD_VERSION = 501
 CMD_RETURN = 502
+CMD_SET_PROTOCOL = 503
 CMD_ERROR = 901
 
 ID_TO_MEANING = {
@@ -257,6 +258,7 @@ ID_TO_MEANING = {
 
     '501': 'CMD_VERSION',
     '502': 'CMD_RETURN',
+    '503': 'CMD_SET_PROTOCOL',
     '901': 'CMD_ERROR',
     }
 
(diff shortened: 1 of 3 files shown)


ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -449,16 +451,11 @@ def _on_run(self):
...
         try:
             while True:
                 try:
                     try:
-                        if get_has_timeout:
-                            cmd = self.cmdQueue.get(1, 0.1)
-                        else:
-                            time.sleep(.01)
-                            cmd = self.cmdQueue.get(0)
+                        cmd = self.cmdQueue.get(1, 0.1)
```
int19h: Is this okay for upstream?

ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -587,19 +571,86 @@ class NetCommand:
...
         self.seq = seq
-        self.text = text
-        encoded = quote(to_string(text), '/<>_=" \t')
-        self.outgoing = '%s\t%s\t%s\n' % (id, seq, encoded)
+        
+        if IS_PY2:
+            if isinstance(text, unicode):
+                text = text.encode('utf-8')
+        else:
+            assert isinstance(text, str)
```
int19h: Looks like this assert should be outside of `if`?

ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
karthiknadig: It is already on upstream https://github.com/fabioz/PyDev.Debugger/blob/master/_pydevd_bundle/pydevd_comm.py#L458

@fabioz usually makes the ganges upstream then ports them over here

ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
int19h: Or rather there should be some assert in the above branch that does the equivalent check there (in case `if` doesn't get triggered).

ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
fabioz: Yeap... that code was for some really old versions of Python (2.3 and before), so, now that the minimum is 2.6 it should be safe.

ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
fabioz: Done.

PR #769, opened 2018-08-26
Fix #355: Add 'remote launch' - launch a program for remote debugging but continue running the user code

### LINKED ISSUE #355: Add 'remote launch' - launch a program for remote debugging but continue running the user code
Today when using `python -m ptvsd --server --port 1234 --file xyz.py`, ptvsd does not run the user code until a debugger is connected.

We'd like to change this to continue running the code (similar to having `ptvsd.enable_attach()` at launch time)

Requirement:
Add a flag named `--wait`

int19h:
Add --wait command line argument, and don't block code execution if it is not specified.

### DIFF
--- a/ptvsd/__main__.py
+++ b/ptvsd/__main__.py
@@ -136,7 +136,7 @@ def _group_args(argv):
             if nextarg is not None:
                 supported.append(nextarg)
             skip += 1
-        elif arg in ('--single-session',):
+        elif arg in ('--single-session', '--wait'):
             supported.append(arg)
         elif not arg.startswith('-'):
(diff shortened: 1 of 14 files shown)


int19h:
Travis is passing now. @karthiknadig, can you take a look at the PR itself?

karthiknadig:
VSTS is stuck. Closing and re-opening PR to try and restart it.

int19h:
It doesn't look like it re-queued it - the last build on VSTS for this branch is from 2 hours ago (and it passed, the results just didn't flow through).

Let me add another commit so that it forces a rebuild.

PR #772, opened 2018-08-28
Adds debug console completion

### LINKED ISSUE #454: Add support for auto completion in console window

karthiknadig:
Fixes #454

@DonJayamanne `supportsCompletionsRequest` should be added to https://github.com/Microsoft/vscode-python/blob/master/src/client/debugger/mainV2.ts


ptvsd/wrapper.py
```diff
@@ -2222,6 +2231,43 @@ def on_exceptionInfo(self, request, args):
                      'source': source},
         )
 
+    @async_handler
+    def on_completions(self, request, args):
+        type_look_up = {
```
int19h: This should be outside of the function, so that the dict is not re-created for every new request.

PR #787, opened 2018-09-07
Fix #200: Local attach

### LINKED ISSUE #200: Local attach
Need to investigate how to implement attaching directly to a local running process (i.e. not via remote debugging) with pydevd. We already ship the attach helper DLLs, but it's likely that they won't "just work", because of the way we deploy pydevd inside of our own package.

int19h:
Add --pid option to ptvsd command line. When used, injects the debugger into the process with the specified ID, such that it connects back to the specified host/port.

### DIFF
--- a/ptvsd/__main__.py
+++ b/ptvsd/__main__.py
@@ -6,6 +6,7 @@ import argparse
 import os.path
 import sys
 
+from ptvsd._attach import attach_main
 from ptvsd._local import debug_main, run_main
 from ptvsd.socket import Address
 from ptvsd.version import __version__, __author__  # noqa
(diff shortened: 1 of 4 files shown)


int19h:
There's one catch: the bitness of the interpreter used to run ptvsd --pid must match the bitness of the process being attached. So we can't just grab a random Python interpreter (e.g. the default one) and use it to run the attach code - the process must be probed first. This complicates things considerably on VS side, but I'm not sure whether that's the case for VSC.

karthiknadig:
It is not related to this change.

DonJayamanne:
Oops will check out the API tomorrow

karthiknadig:
@DonJayamanne Let me know after you have looked at this. I want to merge this.

DonJayamanne:
Tried to test, can't really test without some changes in the debugger.
Please proceed with the merge. We'll test this out when we start the work at our end (will require some changes to the DA, to spin up a socket server for `attach`)

PR #797, opened 2018-09-12
Add ability to run without debugging

### LINKED ISSUE #796: Add ability to launch the debugger in non-debug mode
Currently we can only launch the debugger in non-debug mode when using `-m`.
I'd like to have the same feature by importing PTVSD and invoking a function, similar to debugging using the `debug` function in `debugger.py`

Basically this is necessary to launch the debugger in non-debug mode when using a launcher script.

DonJayamanne:
Fixes #796

### DIFF
--- a/ptvsd/debugger.py
+++ b/ptvsd/debugger.py
@@ -4,7 +4,7 @@
 
 import sys
 
-from ptvsd._local import run_module, run_file
+from ptvsd._local import run_module, run_file, run_main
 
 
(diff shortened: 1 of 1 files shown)


DonJayamanne:
@karthiknadig 
Please could you merge this PR.
Azure CI is failing due to some flaky tests (i've restarted CI twice, and it fails in different places each time).
I.e. unrelated to the PR.

PR #822, opened 2018-09-19
[Blocked] Treat any code not in site-library as user code

### LINKED ISSUE #802: Just-my-code: Anything not in site-packages should be treated as user code

karthiknadig:
Fixes #802

Changing the approach for Just-my-code. In this PR we treat any code not in site-library as user code. 
@int19h Should we make this an option?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,8 @@
 [flake8]
-ignore = E24,E121,E123,E125,E126,E221,E226,E266,E704,E265
-exclude = 
+ignore = W,
+    E24,E121,E123,E125,E126,E221,E226,E266,E704,
+    E265,E722,E501,E731,E306,E401,E302,E222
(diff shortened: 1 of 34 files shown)


int19h:
I think we'd want to enable the scenario where someone is authoring a library, and wants to debug it as it is after pip installs it locally. So an option for whether site-library is user code or not is a good one to have. 

Come to think of it, maybe instead of making it a flag, maybe we can reuse path mappings for this? i.e. for every entry there, allow the user to say that this folder is user code, regardless of any other setting.

d3r3kk:
Special Derek VSTS power maneuver go!

karthiknadig:
This is blocked by #854

karthiknadig:
Will reopen this with the correct change.

PR #823, opened 2018-09-20
Added handling of unhandled exception generated when VS Code disconnects

ahaslam:
Detaching from a VSCode debug session via localhost:3000 on Windows 10 using the debug 'disconnect' button results in an unhandled exception in wrapper.py.

The specific unhandled error is errno.ECONNRESET (10054)

When this exception is not handled correctly, the instance of ptvsd becomes unresponsive and does not allow the debugger to connect again.

Notes:
This change fixes this issue but I can see that convert_eof() in socket.py should be handling this error and raising an EOFError exception - but for some reason it's not and a socket.error with errno == 10054 is raised instead.


int19h:
Thank you for the fix! We'll look into the underlying issue.

PR #838, opened 2018-09-25
Add messages to error responses

### LINKED ISSUE #826: Return error messages in the response for a failed debugger protocol request
I've found at least one case where a certain debugger protocol request fails and a failure response is sent back, however there's no additional information on the failure.
Ideally we should return some more information about the cause of the failure, some error message.
This should help diagnosing issues.

For reference see https://github.com/Microsoft/ptvsd/issues/825

karthiknadig:
fixes #826


d3r3kk:
vsts power punch go

Issue #820, opened 2018-09-19
Sub-process: Do not reuse port in child process


karthiknadig:
Currently we allow re-use of ports in ptvsd by default. we should add a switch that turns off this feature.

int19h:
We will still need to use the option to avoid crashing processes hogging up the port. But it looks like it's all more complicated than just allowing reuse, and also varies by platform.

https://docs.microsoft.com/en-us/windows/desktop/winsock/using-so-reuseaddr-and-so-exclusiveaddruse

https://stackoverflow.com/a/14388707/111335

karthiknadig:
Let us use this work item to add the option to fail on address reuse. I will create another work item to refactor and properly set reuse for all supported platforms.

karthiknadig:
This was a issue specific to windows. We decided to use SO_EXCLUSIVEADDRUSE on windows to fix this.

PR #844, opened 2018-09-25
Use SO_EXCLUSIVEADDRUSE on windows while creating socket

karthiknadig:
#820


d3r3kk:
Retry after resetting webhooks to Azure DevOps...

PR #850, opened 2018-09-26
Test & infrastructure cleanup, stage 1

### LINKED ISSUE #831: Refactoring: IPC-JSON
The product itself and the test suite use two completely different implementations of IPC JSON protocol underlying DAP. This is partly because the product implementation is server-side-only, and the test implementation is client-side-only. Even so, it is mostly the same functionality, and there's no reason to implement it twice in such a radically different way. Furthermore, neither of those implementations has proper test coverage.

This item captures the work necessary to converge on a single implementation that is used throughout the product and the high-level test suite, with proper tests of its own.

int19h:
Add new JSON IPC implementation to be shared between the product and the tests.

Add pytest-based test support, and wire it up to setup.py and Travis.

Dial pylint down to complain about important things only.

Various minor fixes exposed by pylint.

Fixes #831.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "python.linting.enabled": true,
-    "python.linting.pylintEnabled": false,
-    "python.linting.flake8Enabled": true,
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.linting.enabled": true,
+    "python.linting.pylintEnabled": false,
+    "python.linting.flake8Enabled": true,
+}
--- a/ptvsd/messaging.py
+++ b/ptvsd/messaging.py
@@ -5,12 +5,10 @@
 from __future__ import print_function, with_statement, absolute_import
 
 import collections
-import io
 import itertools
 import json
 import sys
 import threading
-import traceback
 
 
 class JsonIOStream(object):
@@ -34,7 +32,7 @@ class JsonIOStream(object):
                 import os, msvcrt
                 msvcrt.setmode(stdin.fileno(), os.O_BINARY)
                 msvcrt.setmode(stdout.fileno(), os.O_BINARY)
-        return self(stdin, stdout)
+        return cls(stdin, stdout)
 
     def __init__(self, reader, writer):
         """Creates a new JsonIOStream.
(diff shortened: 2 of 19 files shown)


int19h:
The new IPC-JSON code is not wired to the product yet. The intent is to write functional tests utilizing it first, then use that to regression-test the product changes for migration.

Note that this change establishes a completely separate test suite under pytests/ - the intent is to fill it with functional tests for ptvsd proper in further stages, transferring (and cleaning up) code from the old tests, and eventually retire the old tests entirely.

To run the new suite:
```sh
pythonX.Y setup.py test
```
it will take care of dependencies etc.

int19h:
Also, while reviewing, please give me feedback on the overall readability of the code and clarity of intent. The goal is to not have any code that is unnecessarily convoluted. Where it's not obvious what the code does even after careful reading, let me know so that I can add more comments, or refactor the code to be clearer.

For the tests especially, the goal is to have readability, even if it means extra verbosity.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
karthiknadig: Ignore this file?

int19h: I removed it from gitignore on purpose - user-specific settings should be in user config, and this one is workspace-specific, and sets the appropriate knobs that everyone should be using (e.g. run flake8).

ptvsd/messaging.py
```diff
@@ -0,0 +1,292 @@
...
+    def __init__(self, channel, seq):
+        self.channel = channel
+        self.seq = seq
+        self.response = None
+        self._got_response = threading.Event()
+        self._handler = lambda _: None
+
+    def _handle_response(self, success, command, error_message=None, body=None):
+        assert self.response is None
+        with self.channel._lock:
```
karthiknadig: Should `_lock` be a "public" property?

ptvsd/messaging.py
```diff
@@ -0,0 +1,292 @@
...
+            success = message['success']
+            command = message['command']
+            error_message = message.get('message', None)
+            body = message.get('body', None)
+            self.on_response(seq, request_seq, success, command, error_message, body)
+        else:
+            raise IOError('Incoming message has invalid "type":\n%r' % message)
+
+    def on_request(self, seq, command, arguments):
+        handler_name = '%s_request' % command
```
karthiknadig: Low pri: Should handler decide the name?

int19h: You mean the suffix? I figured so long as there's a clear convention that guarantees no name clashes, it's fine. Just can't think of any practical scenario where the handler class would want to use custom names... but one possibility is to have a tiny handler wrapper that just has `request` and `event`, and then does its own dispatching to the real handler object by whatever rules it wants.

Issue #833, opened 2018-09-24
Test refactoring: functional tests


int19h:
We need more functional tests - i.e. those that test ptvsd as a black box (including pydevd) against the requirements imposed by clients such as VS and VSCode. This is partially reflected in the existing `system_tests` sub-suite, but there are many things missing there that should be ported from the corresponding VS and VSCode test suits.

int19h:
The initial step here is to look at VS and VSC Python test suites, and determine which tests in them don't have a ptvsd equivalent, or such equivalent is substandard. Then separate work items can be created to cover the specific tests that need porting.

karthiknadig:
I will turn this into an epic and add smaller items to this

PR #857, opened 2018-10-01
Testing framework for timeline-based tests. 

### LINKED ISSUE #832: Test refactoring: high-level test framework
We currently have several different micro-frameworks across the test suite, which share some code and duplicate the other. There needs to be a single micro-framework to write tests against ptvsd, with all tests uniformly using that framework. The framework should be pytest-based, and should abstract away lifecycle management in a way that makes it clear what the test is doing by looking solely at the test code itself, but without too much repeated scaffolding. The framework itself should be properly covered by tests.

int19h:
See [pytests/helpers/timeline.md](https://github.com/int19h/ptvsd/blob/tests2/pytests/helpers/timeline.md) for a detailed description.

Fixes #832, and lays the groundwork for #833.


ptvsd/messaging.py
```diff
@@ -34,6 +34,11 @@ def from_stdio(cls):
                 msvcrt.setmode(stdout.fileno(), os.O_BINARY)
         return cls(stdin, stdout)
 
+    @classmethod
+    def from_socket(cls, socket):
+        socket_io = socket.makefile('rwb', 0)
```
karthiknadig: >makefile [](start = 27, length = 8)

Should we check if the socket has timeout set? From 2.7.15 docs for socket.makefile(): "The socket must be in blocking mode (it can not have a timeout)."

pytests/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.fixture
+def daemon():
+    """Provides a factory function for daemon threads. The returned thread is
+    started immediately, and it must not be alive by the time the test returns.
+    """
+
+    daemons = []
+
+    def factory(func):
```
karthiknadig: >factory(func) [](start = 8, length = 13)

Should this take args?

pytests/helpers/timeline.md
```diff
@@ -0,0 +1,313 @@
...
+```py
+('Event', 'stopped', {'reason': 'breakpoint', 'threadId': 2})
+('Event', 'continued', {'threadId': 1})
+('Event', 'thread', {'reason': 'exited', 'threadId': 1})
+```
+This is most commonly used as seen above, with related events where their relative ordering is unspecified (usually also framed by `>>` to narrow it down to a specific request/response pair). It can also be used to concurrently send multiple requests, and wait until they all got their responses:
+```py
+pause1 = debug_session.send_request('pause', {'threadId': '1'})
+pause2 = debug_session.send_request('pause', {'threadId': '2'})
+debug_session.wait_until(Request(pause1) & Request(pause2))
```
karthiknadig: >Request(pause1) & Request(pause2) [](start = 25, length = 33)

Should this be just pause1 & pause2

pytests/conftest.py
int19h: You mean, to pass through  to `Thread` constructor? This is meant for the simplest (but common) case, so I think most of them don't make much sense there. But `name` might be useful, since it is used to identify the thread in stack traces that are dumped if the test times out.

Or better yet, derive name from the function name inside the decorator, but then allow to specify a suffix to distinguish between different daemons produced from the same function.

ptvsd/messaging.py
int19h: Yep, and it does in fact break spectacularly if you try! It's easy to check, too.

pytests/helpers/timeline.md
int19h: It should actually be `Response(pause1) & Response(pause2)`.

pytests/helpers/timeline.md
```diff
@@ -0,0 +1,313 @@
...
+('Event', 'thread', {'reason': 'exited', 'threadId': 1})
+('Event', 'stopped', {'reason': 'breakpoint', 'threadId': 2})
+('Event', 'continued', {'threadId': 1})
+```
+
+It is usually used to concurrently send multiple requests, and wait until the first one gets a response:
+```py
+pause1 = debug_session.send_request('pause', {'threadId': '1'})
+pause2 = debug_session.send_request('pause', {'threadId': '2'})
+pause_response = debug_session.wait_until(pause1 | pause2)
```
karthiknadig: >pause1 | pause2 [](start = 42, length = 15)

Should this be Response(pause1) | Resposne(pause2)?

int19h: Indeed. I'm going to sprinkle asserts somewhat more liberally in the library, since clearly this might be a fairly common mistake, if I'm the one doing it after writing the code :)

karthiknadig:
> 	$(PYTHON) -m pip install pytest

pytest-timeout? Shoudl we change all of thse to `--upgrade`?

---
Refers to: Makefile:19 in ba856a3. [](commit_id = ba856a3bd9a5c4b07327733cc09d8a1b100896b3, deletion_comment = False)

int19h:
It gets installed automatically as part of the run when you do `setup.py test`. In fact, pytest also gets installed automatically. The problem is that Travis specifically already has it installed, but with a much older version (that pytest-timeout doesn't support), and then setup.py doesn't bother checking the installed version number...

Going forward, we should just move to tox, which will solve these kinds of issues by running everything in a virtualenv. For now, treat it as a collection of hacks that are necessary to get things going :)

Issue #845, opened 2018-09-26
UnicodeDecodeError in Python 2.7 prevents debugging


karthiknadig:
## Environment data

- PTVSD version: master
- OS and version: windows
- Python version (& distribution if applicable, e.g. Anaconda): 2.7
- Using VS Code or Visual Studio: VS

## Actual behavior

Debugging fails with this error:
```
ERROR: 'setDebuggerProperty' request failed!

ProtocolException: Traceback (most recent call last):
File "C:\Python27\lib\site-packages\ptvsd\ipcjson.py", line 358, in on_request
target(request, args)
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 1271, in f
fut = m(self, self.loop, *args, **kwargs)
File "C:\Python27\lib\site-packages\ptvsd\futures.py", line 140, in g
it = f(self, *args, **kwargs)
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 2325, in on_setDebuggerProperty
self._apply_code_stepping_settings()
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 1421, in _apply_code_stepping_settings
'\t'.join(project_dirs))
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 1300, in pydevd_request
return self._pydevd_request(self.loop, cmd_id, args)
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 436, in pydevd_request
seq, s = self.make_packet(cmd_id, args)
File "C:\Python27\lib\site-packages\ptvsd\wrapper.py", line 421, in make_packet
s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 3: ordinal not in range(128)
```

## Expected  behavior

Should not fail with `UnicodeDecodeError`

Bug report: https://developercommunity.visualstudio.com/content/problem/332784/vs-python-%E9%A1%B9%E7%9B%AE%E5%88%87%E6%8D%A2%E7%89%88%E6%9C%AC%E6%97%A0%E6%B3%95%E8%BF%90%E8%A1%8C.html

abhinuvpitale:
The input should be checked and there should be exception handling here, I can take this up!

karthiknadig:
Fixed via #884 . this code path is no longer needed.

PR #861, opened 2018-10-02
added unicodeDecodeError exception for #845

abhinuvpitale:
catch the exception, reload sys and set the system variable correctly.

### DIFF
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -420,8 +420,7 @@ class PydevdSocket(object):
             self.seq += 1
         try:
             s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
-        except UnicodeDecodeError, err:
-            import sys
+        except UnicodeDecodeError:
             reload(sys)
             sys.setdefaultencoding('utf8')
             s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
--- a/ptvsd/wrapper.py
+++ b/ptvsd/wrapper.py
@@ -421,8 +421,7 @@ class PydevdSocket(object):
         try:
             s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
         except UnicodeDecodeError:
-            reload(sys)
-            sys.setdefaultencoding('utf8')
+            args = args.encode('ascii','replace')
             s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
         return seq, s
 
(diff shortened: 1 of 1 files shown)


ptvsd/wrapper.py
```diff
@@ -418,7 +418,12 @@ def make_packet(self, cmd_id, args):
         with self.lock:
             seq = self.seq
             self.seq += 1
-        s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
+        try:
+            s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
+        except UnicodeDecodeError:
+            reload(sys)
```
fabioz: This isn't the correct fix... rather, you should make sure that you're not mixing bytes and unicode doing the proper encode/decode calls (you have to be careful because it has to work on py2 and 3 so, this should definitely have a test to make sure it works well on both).

Also, making `reload(sys)` is full of caveats and is definitively not something that the debugger should be doing.

If I can suggest, take a look at https://nedbatchelder.com/text/unipain.html for some explanation on how unicode works.

abhinuvpitale: @fabioz thank you, I will update the PR likewise

ptvsd/wrapper.py
```diff
@@ -418,7 +418,11 @@ def make_packet(self, cmd_id, args):
         with self.lock:
             seq = self.seq
             self.seq += 1
-        s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
+        try:
+            s = u'{}\t{}\t{}\n'.format(cmd_id, seq, args)
+        except UnicodeDecodeError:
+            args = args.encode('ascii','replace')
```
int19h: I don't think this is the right place to do it. The specific call that causes #845 - which is `pydevd_request(...)` in `_apply_code_stepping_settings` - should be the one ensuring that `args` is properly encoded, because only it knows the correct encoding to apply (and "ascii" with "replace" here will just mangle the argument).

In this particular case, since the caller is trying to pass in a path, it should check if it's `bytes` first, and if it is, *de*code it using `sys.getfilesystemencoding()`, so that it's always passing a proper unicode string to `send_request`.

abhinuvpitale:
Thank you for your comments.
@int19h I understand that, but based on what @karthiknadig says, I dont think I should update there.
Instead `args` can be manually checked for bytes and then this can be applied, is that right?
If not, I am not sure whats the right way, could you point me in the right direction here?

karthiknadig:
Closing this. The code path was removed. @abhinuvpitale install ZenHub and look for Icebox/Backlogs for issues that you might like to work on.

abhinuvpitale:
Thanks @karthiknadig

Issue #856, opened 2018-09-28
subprocess: add option to select port from a given range


karthiknadig:
This option is internal and will be used by sub-processes only.

DonJayamanne:
What's the default Port range if not specified?

karthiknadig:
> What's the default Port range if not specified?

@DonJayamanne This is TBD.

int19h:
It should be the [ephemeral port range](https://en.wikipedia.org/wiki/Ephemeral_port#Range).

DonJayamanne:
Here's a use case, that would make this option external (not only internal).
* Use starts debugging in docker
* Ports 5000-5050 are mapped out
* Debugger starts on 5000
* The subprocess should probably start on 5001 (using the previous Port).

If the Port is not predictable, then we cannot attach to the sun process if it's not in a mapped range.
Sure user could map everything, but that's not always advisable. One night not want to expose all ports (connections to secure changes, mongo, etc to outside docker).

The same apples when connecting to remote Port/ip over ssh.

int19h:
I think there was supposed to be a distinction between "range of ports *this* process should pick a random port from" (i.e. a new option that basically replaces `--port` when it's used - this is what this item is for), and "range of ports this process should tell the *child* processes to pick their ports from" (this is the one that's needed in your scenario).

The problem with conflating the two is that when a range is specified, the first *free* port is supposed to be picked. Suppose our new option is called `--port-range`; the user starts the main process as:
```sh
ptvsd --port-range 5000-5050 ...
```
and then tries to attach the IDE to port 5000. If there was nothing else listening on 5000, then this works. But if there was, it will just silently pick 5001, and trying to connect to 5000 will not work because it'll connect to something else. It would be hard to figure out what went wrong.

What we need is two distinct options - one for child processes to specify a range of ports, and one for the parent process to know which range of ports it should specify for child processes. So, user would invoke the main process thus:
```sh
ptvsd --port 5000 --child-port-range 5000-5050 ...
```
This makes the main process listen on 5000, and if it can't do that because that port is already in use, it immediately prints an error and quits - just like today. But for any child process that gets spawned, the main process would then spawn it with:
```sh
ptvsd --port-range 5000-5050 ...
```
which picks the first free port, and only errors out if there were no free ports in the range at all. So `--port` and `--port-range` are mutually exclusive, but `--port` and `--child-port-range` are orthogonal.

[...]

PR #862, opened 2018-10-02
Fix #810: Sub-Process: monkey patch pydevd to start process using ptvsd

### LINKED ISSUE #810: Sub-Process: monkey patch pydevd to start process using ptvsd
Monkey patch `pydev_monkey.py` to start new processes using `ptvsd`. Ensure that the new process uses `--wait` feature.

int19h:


int19h:
This work item doesn't capture the totality of the multiproc work. The next step is #856, which will add another switch that will allow specifying a range for ports instead of a single port. And then the parent process will just pass the range given to `--multiprocess-port-range` to the child process, and the child process will pick the first available free port from it.

int19h:
Although come to think of it, we don't really need a switch, since the bootstrapping code can just do it directly. Let me try this...

PR #866, opened 2018-10-03
Fix #811: new socket server to receive pid and port for subprocesses

### LINKED ISSUE #811: Sub-Process: new socket server to receive pid and port for the new process
Start a socket server in ptvsd to receive process id port. This may also be used to exchange GUID or other identifiers. Al;so, implement custom event that is sent to VSC about the new process.

int19h:


ptvsd/multiproc.py
```diff
@@ -0,0 +1,90 @@
...
+            debug('Child process %d listening on port %d' % (os.getpid(), port))
+            break
+    else:
+        raise Exception('Could not find a free port in range {first_port}-{last_port}')
+
+    conn = create_client()
+    conn.connect(('localhost', parent_port))
+    stream = JsonIOStream.from_socket(conn)
+    channel = JsonMessageChannel(stream)
+    channel.send_event('ptvsd_subprocess', {
```
int19h: @DonJayamanne, this is how the notification will look like from VSC perspective. At this point it should do an attach to the same host as what was used for the main process, using the specified port. 

In the meantime, I tested it by attaching to child processes manually after the notification shows up in debug.log, and things seem to work fine.

ptvsd/multiproc.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _handle_subprocess(n, stream):
+    class Handlers(object):
+        def ptvsd_subprocess_event(self, channel, body):
+            debug('ptvsd_subprocess: %r' % body)
+            subprocess_queue.put(body)
+            channel.close()
+
+    name = 'ptvsd.SubprocessListener%d' % n
```
karthiknadig: >ptvsd. [](start = 12, length = 6)

Remove 'ptvsd.', new_hidden_thread will append 'ptvsd.' to the name.

DonJayamanne:
@int19h #karthiknadig
Please do ping me when this does get merged into master

PR #876, opened 2018-10-04
Allow configuring whether the debugger should stop/print errors when there's an exception evaluating a breakpoint condition. Fixes #853

### LINKED ISSUE #853: Undefined variable error in breakpoint condition on for line
## Environment data

- PTVSD version: 4.1.3
- OS and version: Windows 10

## Actual behavior

Undefined variable error, breakpoint never hit
```
Error while evaluating expression: i>1600 
  File "c:\program files (x86)\microsoft visual studio\2019\preview\common7\ide\extensions\microsoft\python\core\Packages\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_frame.py", line 52, in handle_breakpoint_condition 
NameError: name 'i' is not defined 
The program 'python.exe' has exited with code -1 (0xffffffff).
```

## Expected  behavior

Undefined variable error ignored and breakpoint hit once it is defined. That was the behavior in ptvsd 3.x. 

## Steps to reproduce:
1. Write this code:

```
from math import cos, radians 
def make_dot_string(x): 
    rad = radians(x) 
    numspaces = int(20 * cos(radians(x)) + 20) 
    st = ' ' * numspaces + 'o' 
    return st 
def main(): 
    for i in range(0, 1800, 12): 
        s=make_dot_string(i) 
        print(s) 
main()
```

2. set a conditions breakpoint on the `for i in range...` line, hover on red dot and click breakpoint setting, set condition `i>1600`

3. start debugging

Note: of course if you put the breakpoint inside the for loop instead, that works as expected.

fabioz:
Note that I haven't added tests related to this on `ptvsd`, only on `pydevd` (although I changed `ptvsd` to configure the settings as needed).


karthiknadig:
@fabioz The ptvsd test failures are due to the new command, the sequence numbers are messed up because of that. Can you comment out the lines 1408-1411? I will update these with the test fixes.

fabioz:
> @fabioz The ptvsd test failures are due to the new command, the sequence numbers are messed up because of that. Can you comment out the lines 1408-1411? I will update these with the test fixes.

Done

ptvsd/wrapper.py
```diff
@@ -43,6 +43,7 @@
 from ptvsd.safe_repr import SafeRepr  # noqa
 from ptvsd.version import __version__  # noqa
 from ptvsd.socket import TimeoutError  # noqa
+import json
```
karthiknadig: Please comment all the changes in this file. Looks like linting and some tests are impacted by line and the change `on_pydevd_get_breakpoint_exception` below.

PR #881, opened 2018-10-06
fix json comments syntax highlight

ayush--s:
this fixes the json comments being highlighted as error in README.md


karthiknadig:
Thank you for the fix.

PR #882, opened 2018-10-06
Don't add additional new line with @pyfile

### LINKED ISSUE #875: pyfile adds an additional new line in every line of code
## Environment data

- PTVSD version: master
- OS and version: any
- Using VS Code or Visual Studio:

## Actual behavior

code:
```python
@pyfile
def foo():
    print('one')
    print('two')
```
The file generated by pyfile, `foo.py`:
```python
print('one')

print('two')

```

## Expected  behavior

The file generated by pyfile, `foo.py`:
```python
print('one')
print('two')
```

Having the extra blank lines makes it confusing to set breakpoints in tests.

rozuur:
Fixes #875

### DIFF
--- a/pytests/conftest.py
+++ b/pytests/conftest.py
@@ -84,7 +84,7 @@ def pyfile(request, tmpdir):
         line = source[0]
         indent = len(line) - len(line.lstrip())
         source = [line[indent:] for line in source]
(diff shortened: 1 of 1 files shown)


rozuur:
@karthiknadig Do you have any other issues that I can work on?

PR #893, opened 2018-10-09
Undefined variable error in breakpoint condition

### LINKED ISSUE #853: Undefined variable error in breakpoint condition on for line

karthiknadig:
Fixes #853

### DIFF
--- a/tests/highlevel/test_messages.py
+++ b/tests/highlevel/test_messages.py
@@ -10,6 +10,7 @@ from _pydevd_bundle.pydevd_comm import (
     CMD_EVALUATE_EXPRESSION,
     CMD_EXEC_EXPRESSION,
     CMD_EXIT,
+    CMD_GET_BREAKPOINT_EXCEPTION,
     CMD_GET_FRAME,
     CMD_GET_VARIABLE,
     CMD_LIST_THREADS,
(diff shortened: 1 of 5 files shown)


tests/highlevel/test_messages.py
```diff
@@ -2795,7 +2794,7 @@ def test_basic(self):
 
 class GetExceptionBreakpointEventTests(PyDevdEventTest, unittest.TestCase):
 
-    CMD = CMD_GET_BREAKPOINT_EXCEPTION
+    CMD = 1234
```
fabioz: Is the magic number some leftover?

karthiknadig: The test is wrongly named. It is supposed to test a unsupported command. I need to change this.

---
In reply to: [223824331](https://github.com/Microsoft/ptvsd/pull/893#discussion_r223824331) [](ancestors = 223824331)

karthiknadig: I need to update this. There should be two tests here.  One for the GetBreakpointException and another for unsupported command.

PR #895, opened 2018-10-09
Fix #894: Race condition in JsonMessageChannel

### LINKED ISSUE #894: Race condition in JsonMessageChannel
```py
    def send_request(self, command, arguments=None):
        d = {'command': command}
        if arguments is not None:
            d['arguments'] = arguments
        seq = self._send_message('request', d)
        request = Request(self, seq)
        with self._lock:
            self._requests[seq] = request
        return request
```
Note that the message is sent before the requests dict is updated. If it goes fast enough, the response handler will receive a response to an "unknown" request.

int19h:
Register the request before sending the message that corresponds to it.

### DIFF
--- a/ptvsd/messaging.py
+++ b/ptvsd/messaging.py
@@ -5,6 +5,7 @@
 from __future__ import print_function, with_statement, absolute_import
 
 import collections
+import contextlib
 import itertools
 import json
 import sys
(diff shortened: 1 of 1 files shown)


int19h:
Please hit "Merge" if the change is good - this is causing multiple automation failures (tests are running faster!), so the faster we can get this into master, the better.

PR #897, opened 2018-10-10
Fix #886: ptvsd command line should support the equivalent of python -c

### LINKED ISSUE #886: ptvsd command line should support the equivalent of python -c
Python supports three ways to start running code non-interactively:

- python file.py
- python -m module
- python -c "python code"

ptvsd currently supports the first two, but not the last one. Implementing it will greatly simply the implementation of command line rewriting for #885. It will also make ptvsd more consistent wrt standard Python command line options.

int19h:
Plus some test fixes.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 ignore = W,
     E24,E121,E123,E125,E126,E221,E226,E266,E704,
-    E265,E722,E501,E731,E306,E401,E302,E222
+    E265,E722,E501,E731,E306,E401,E302,E222,E303
 exclude =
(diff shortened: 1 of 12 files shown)


.flake8
```diff
@@ -1,7 +1,7 @@
 [flake8]
 ignore = W,
     E24,E121,E123,E125,E126,E221,E226,E266,E704,
-    E265,E722,E501,E731,E306,E401,E302,E222
+    E265,E722,E501,E731,E306,E401,E302,E222,E303
```
int19h: This is the one that complains about incorrect number of blank lines between things. It dost protest too much...

PR #909, opened 2018-10-12
Fix #896: Changes in CLI arguments between versions

### LINKED ISSUE #896: Changes in CLI arguments between versions
## Environment data

- PTVSD version: 4.1.3
- Using VS Code or Visual Studio: N/A

## Actual behavior

* CLI args for starting apps for remote debugging has changed 
* Now the arg requires `--server-host`, this wasn't the case in the past.
* This is a breaking change.

## Expected  behavior

* We need to revert if possible as existing users will not be able to use PTSVD for remote debugging scenarios.

Reported here https://github.com/Microsoft/vscode-python/issues/2833#issuecomment-428422616

int19h:
Remove --server-host. Server mode is always implied by default now in all circumstances, and --client is used to specify client mode.

### DIFF
--- a/ptvsd/__main__.py
+++ b/ptvsd/__main__.py
@@ -24,8 +24,6 @@ For the PyDevd CLI handling see:
 
 PYDEVD_OPTS = {
     '--file',
-    '--client',
-    #'--port',
     '--vm_type',
 }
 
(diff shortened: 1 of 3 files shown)


int19h:
@DonJayamanne, please take a look to make sure that this is what we want.

DonJayamanne:
I'll test this to confirm it work as expected, we can then publish a new version of PTVSD.

PR #911, opened 2018-10-12
 Add socket reuse tests

### LINKED ISSUE #851: Add test to ensure exclusive socket address and port usage
We use SO_REUSEADDR on linux/mac and SO_EXCLUSIVEADDRUSE on windows. when creating a socket. This test should ensure that on all supported platforms two processes cannot bind to same port and address. 

NOTE: using SO_REUSEADDR on windows allows two process to bind to same address+port.

karthiknadig:
Fixes #851

### DIFF
--- a/pytests/test_socket.py
+++ b/pytests/test_socket.py
@@ -1,0 +1,27 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+import pytest
+from ptvsd.socket import create_server, shut_down
+
+
+class TestSocketServerReuse(object):
+    HOST1 = '127.0.0.1'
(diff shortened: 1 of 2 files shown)


pytests/test_socket.py
```diff
@@ -0,0 +1,27 @@
...
+    HOST1 = '127.0.0.1'
+    HOST2 = '127.0.0.2'
+    PORT1 = 7890
+
+    def test_reuse_same_address_port(self):
+        sock1 = create_server(self.HOST1, self.PORT1)
+        with pytest.raises(Exception):
+            create_server(self.HOST1, self.PORT1)
+        assert sock1.getsockname() == (self.HOST1, self.PORT1)
+        shut_down(sock1)
```
int19h: You might want to do these in `finally`, so that they get closed even if the test fails.

Issue #425, opened 2018-05-21
Auto-attach to sub process (auto-reload of Django, Flask)


DonJayamanne:
pydevd supports this today, all we need to do is just pass a flag.

DonJayamanne:
@karthiknadig Please check whether this is required for VS

zooba:
Isn't this the same as #57?

karthiknadig:
The important distinction is the number of process that the adapter itself is attached to. I think in this case there will be one process that the adapter is expected to attach to or launch, and continue to be so even after template reload results in a new sub-process. Although I am not sure how the template breakpoints will work in this case, we have to test this to see if there are any other limitations.

DonJayamanne:
I've created a separate issue for django, flask debugging due to an item added into the exit criteria of PTVSD. I do not want multi-process debugging not to get implemented simply because the VSC protocol doesn't support it. I.e. not being able to differentiate between the process ids when debugging (step in, step out, what process ids do threads belong to, etc).

In the case of auto-reloading of django templates, we (VSC) don't care whether its mult-process or not. All we care about is breaking into the code and supporting breakpoints (i.e. standard debugging). **I.e. process id doesn't matter**.

jxramos:
This is what got rendered....

```
I00000.027: Darwin-17.5.0-x86_64-i386-64bit x86_64
            CPython 2.7.10 (64-bit)
            ptvsd 4.2.8

I00000.027: main(['/Users/$USER/.vscode/extensions/ms-python.python-2019.4.12954/pythonFiles/ptvsd_launcher.py', '--nodebug', '--client', '--host', 'localhost', '--port', '57251', '/Users/$USER/source_dir/test.py', '--retry', '0', '--udids', 'xxxxxxxxxxx', '--emails', 'name@domain.com', '--tests', '19245179', '--description', 'test_description'])

I00000.027: sys.argv after parsing: ['/Users/$USER/.vscode/extensions/ms-python.python-2019.4.12954/pythonFiles/ptvsd_launcher.py', '--retry', '0', '--udids', 'xxxxxxxxxxx', '--emails', 'name@domain.com', '--tests', '19245179', '--description', 'test_description']

D00000.027: sys.prefix: ('/System/Library/Frameworks/Python.framework/Versions/2.7',)

D00000.027: site.getusersitepackages(): /Users/$USER/Library/Python/2.7/lib/python/site-packages

[...]

PR #916, opened 2018-10-14
#425: Auto-attach to sub process (auto-reload of Django, Flask) 

int19h:
This completes the implementation of multiprocess debugging for all cases.

### DIFF
--- a/ptvsd/multiproc.py
+++ b/ptvsd/multiproc.py
@@ -11,6 +11,7 @@ import re
 import socket
 import sys
 import time
+import traceback
 
 try:
     import queue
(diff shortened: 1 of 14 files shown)


ptvsd/wrapper.py
```diff
@@ -1286,21 +1279,30 @@ def _stop_event_loop(self):
...
     def _subprocess_notifier(self):
         while not self.closed:
             try:
-                subprocess_info = multiproc.subprocess_queue.get(block=False, timeout=0.1)
+                subprocess_request, subprocess_response = multiproc.subprocess_queue.get(block=False, timeout=0.1)
             except queue.Empty:
+                continue
+
+            try:
+                self.send_event('ptvsd_subprocess', **subprocess_request)
```
int19h: Note that this is still send_event, despite our discussion. The reason is that the old IPC JSON implementation does not support sending requests, only handling them. We should revisit this after refactoring.

ptvsd/multiproc.py
```diff
@@ -7,117 +7,252 @@
...
+
+    # Send the notification about ourselves to root, and wait for it to tell us
+    # whether an incoming connection is anticipated. This will be true if root
+    # had successfully propagated the notification to the IDE, and false if it
+    # couldn't do so (e.g. because the IDE is not attached). There's also the
+    # possibility that connection to root will just drop, e.g. if it crashes -
+    # in that case, just exit immediately.
+
+    request = channel.send_request('ptvsd_subprocess', {
+        'parentProcessId': options.subprocess_of,
```
karthiknadig: This should have the launch\attach request info in it?

int19h:
There is an issue with timeline tests where it consistently can't realize an event even though it's in the timeline. This only repros on Travis, not locally, so I'm going to do a bunch of debug commits to debug this. So I'm closing this for now to avoid spamming notifications, and will reopen once this is fixed (and after I clean up all those commits).

Issue #805, opened 2018-09-17
Continue (F5) causes Could not find thread * if multi-threaded


medariox:
## Environment data

- PTVSD version: 4.1.1
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.14
- Using VS Code or Visual Studio: VS Code

## Actual behavior
After the first breakpoint, it's not possible to continue to the next one (Continue (F5)).

## Expected  behavior
Continue to the next breakpoint.

## Steps to reproduce:
After applying fix https://github.com/Microsoft/ptvsd/pull/758 it is possible to break correctly.
Set multiple break points. The first will break correctly, the second will output following message in the debug console:
```
Could not find thread *
Available: ['pid_10212_id_72621808', 'pid_10212_id_72243296', 'pid_10212_id_72621584', 'pid_10212_id_73150360', 'pid_10212_id_72697952', 'pid_10212_id_72541184', 'pid_10212_id_56967464']
```

fabioz:
Ok, I took a look at this and what's happening is the following:

- all threads are paused (first break)
- a continue is issued -- at this point, it will mark all threads to be run, but they aren't really running (they're just marked to run).
- one of those threads starts running, hits a breakpoint and pauses all existing threads -- at this point, other threads didn't get a chance to run at all, so, they don't send a `CMD_THREAD_RUN` nor a `CMD_THREAD_SUSPEND` as they didn't even leave the pause state, yet, the UI is seeing them in the `run` state, even though they haven't sent the related message.

Now, I'm not exactly sure this is the same issue as I couldn't hit that `Could not find thread *` condition, but it seems like it for me...

Steps to reproduce:

Create file with:

```
import threading

numbers = range(3)


def double(number):
    while True:
        doubled = number * 2
        print(doubled)
        import time
        time.sleep(.5)  # breakpoint here
    return doubled

threads = []
for num in numbers:
    thread = threading.Thread(target=double, args=(num,))
    thread.start()
    threads.append(thread)

for thread in threads:
    thread.join()
```

put breakpoint on the `# breakpoint here`, launch program, then after it hits the breakpoint hit continue a few times (at each time the `main thread` should always be running as it'll be effectively deadlocked in the `join`, but all the other ones should always be `suspended` -- or maybe temporarily running while on `time.sleep()` and the shortly thereafter should be `suspended`).

[...]

PR #922, opened 2018-10-16
Provide a single notification when a breakpoint is hit. #805

fabioz:
Note this still needs additional work in `ptvsd` to be actually integrated.

Things that need to be fixed:

- Timing issues because the timeout in `PyDBCommandThread` changed a bit (from 0.5 to 0.3 and no longer sleeps more initially -- in `pydevd.PyDBCommandThread._on_run`)

- When a manual breakpoint is added, instead of `CMD_THREAD_SUSPEND`, a `CMD_SET_BREAK` reason is given (in `pydevd._locked_settrace`)

- `CMD_SUSPEND_ON_BREAKPOINT_EXCEPTION` was renamed to `CMD_PYDEVD_JSON_CONFIG` as it now accepts json for additional configurations (i.e.: to say that it should generate a single notification for multiple threads).

- 2 new events are generated and should be used to detect that all threads are stopped or all threads are running:
   `CMD_THREAD_SUSPEND_SINGLE_NOTIFICATION`
   `CMD_THREAD_RESUME_SINGLE_NOTIFICATION`
  the existing events (`CMD_THREAD_SUSPEND` and `CMD_THREAD_RUN`) that notify about particular threads being run or being paused are still generated and should be ignored by `ptvsd`.


fabioz:
@karthiknadig just wanted to note that I started to take a look to integrate and fix failing tests on `ptvsd`.

PR #930, opened 2018-10-17
pytests for django and flask

karthiknadig:
Fixes: #891

### DIFF
--- a/pytests/func/test_django.py
+++ b/pytests/func/test_django.py
@@ -6,7 +6,6 @@ from __future__ import print_function, with_statement, absolute_import
 
 import os.path
 import pytest
-import sys
 
 from ..helpers.pattern import ANY
 from ..helpers.session import DebugSession
--- a/pytests/func/test_django.py
+++ b/pytests/func/test_django.py
@@ -8,9 +8,10 @@ import os.path
 import pytest
 
 from ..helpers.pattern import ANY
+from ..helpers.session import DebugSession
 from ..helpers.timeline import Event
-from .testfiles.testroots import get_test_root
-from ..helpers.webhelper import get_web_content, wait_for_connection
+from ..helpers.pathutils import get_test_root
+from ..helpers.webhelper import get_url_from_str, get_web_content, wait_for_connection
 
 DJANGO1_ROOT = get_test_root('django1')
 DJANGO1_MANAGE = os.path.join(DJANGO1_ROOT, 'app.py')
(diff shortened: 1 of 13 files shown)


pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+
+DJANGO1_ROOT = get_test_root('django1')
+DJANGO1_MANAGE = os.path.join(DJANGO1_ROOT, 'app.py')
+DJANGO1_TEMPLATE = os.path.join(DJANGO1_ROOT, 'templates', 'hello.html')
+
+def _wait_for_django_link(debug_session):
+    o = None
+    while True:
+        if o:
+            o = debug_session.wait_for_next(o >> Event('output'))
```
int19h: You shouldn't need to do `o >> ...` here if you call `session.proceed()` after `wait_for_next` returns. That will specify the cut-off point for the next `wait_for_next` call, below which it won't even look for events.

(There was a bug about this not working right, but it should be fixed in master.)

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+
+    bp_var_content = 'Django-Django-Test'
+    debug_session.send_request('setBreakpoints', arguments={
+        'source': {'path': bp_file},
+        'breakpoints': [{'line': bp_line}, ],
+    }).wait_for_response()
+
+    debug_session.start_debugging()
+
+    link = _wait_for_django_link(debug_session)
```
int19h: I think this is best done with a single wait and `ANY.such_that`. Something like:
```py
output_with_link = debug_session.wait_for_next(Event('output', ANY.dict_with({
    'output': ANY.such_that(lambda s: get_url_from_str(s) is not None)
}))
```

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+        assert thread_stopped.body['threadId'] is not None
+
+        tid = thread_stopped.body['threadId']
+
+        resp_stacktrace = debug_session.send_request('stackTrace', arguments={
+            'threadId': tid,
+        }).wait_for_response()
+        assert resp_stacktrace.body['totalFrames'] > 1
+        frames = resp_stacktrace.body['stackFrames']
+        assert frames[0] == {
```
int19h: Looks like we need `ANY.list_starting_with` :)

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+            'id': 1,
```
int19h: I don't think the protocol guarantees sequential breakpoint IDs, so we should probably not enforce that in the tests - i.e. `ANY.int` rather than specific value.

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+        resp_scopes = debug_session.send_request('scopes', arguments={
+            'frameId': fid
+        }).wait_for_response()
+        scopes = resp_scopes.body['scopes']
+        assert len(scopes) > 0
+
+        resp_variables = debug_session.send_request('variables', arguments={
+            'variablesReference': scopes[0]['variablesReference']
+        }).wait_for_response()
+        variables = list(v for v in resp_variables.body['variables'] if v['name'] == 'content')
```
int19h: `[... for ...]` rather than `list(... for ...)`

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+            if link is not None:
+                return link
+
+
+def _django_no_multiproc_common(debug_session):
+    debug_session.multiprocess = False
+    debug_session.cli_args = ['runserver', '--noreload', '--nothreading']
+
+    debug_session.ignore_unobserved += [
+        # The queue module can spawn helper background threads, depending on Python version
```
int19h: The comment is probably wrong here. Might just say that Django starts a lot of helper threads, and the test is too high-level to bother with them.

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+
+    debug_session.start_debugging()
+
+    base_link = _wait_for_django_link(debug_session)
+    assert base_link is not None
+
+    link = base_link + ex_type if base_link.endswith('/') else ('/' + ex_type)
+    with get_web_content(link, {}):
+        thread_stopped = debug_session.wait_for_next(Event('stopped', ANY.dict_with({'reason': 'exception'})))
+        assert thread_stopped == Event('stopped', ANY.dict_with({
```
int19h: Does it need `ANY` here - i.e. is there's anything else in the dict?

pytests/func/test_django.py
```diff
@@ -0,0 +1,190 @@
...
+        debug_session.wait_for_next(Event('continued'))
+
+    # shutdown to web server
+    link = base_link + 'exit' if base_link.endswith('/') else '/exit'
+    with get_web_content(link):
+        pass
+
+
+def test_django_handled_exception_no_multiproc(debug_session):
+    _django_exception_no_multiproc(debug_session, 'handled', 50)
```
int19h: For this sort of parametrization, consider using `@pytest.parametrize` instead of helpers. You can parametrize on groups of arguments instead of single ones:
```py
@pytest.mark.parametrize('ex_type, ex_line', [
  ('handled', 50),
  ('unhandled', 64),
])
def test_django_no_multiproc(debug_session, ex_type, ex_line):
   ...
```

pytests/func/test_flask.py
```diff
@@ -0,0 +1,321 @@
...
+from ..helpers.webhelper import get_url_from_str, get_web_content
+from .testfiles.testroots import get_test_root
+
+
+FLASK1_ROOT = get_test_root('flask1')
+FLASK1_APP = os.path.join(FLASK1_ROOT, 'app.py')
+FLASK1_TEMPLATE = os.path.join(FLASK1_ROOT, 'templates', 'hello.html')
+
+
+def _wait_for_flask_link(debug_session):
```
int19h: This looks identical to the Django version, so perhaps it should be in helpers as well?

(and then the same points apply here - can be done with a single `wait_for_next`)

pytests/func/test_flask.py
```diff
@@ -0,0 +1,321 @@
...
+        Event('thread', ANY.dict_with({'reason': 'started'})),
+        Event('module'),
+        Event('stopped'),
+        Event('continued')
+    ]
+
+    debug_session.debug_options = ['RedirectOutput', 'Jinja']
+    debug_session.cwd = FLASK1_ROOT
+    debug_session.env.update(env)
+    debug_session.expected_returncode = ANY  # No clean way to kill Django server
```
int19h: *Flask

pytests/func/test_django.py
karthiknadig: yes, there is text, description, they can have value depending on the situation.

int19h: But it does explicitly check for text and description.

karthiknadig: sorry that would be `preserveFocusHint`, this can be True or False, also `allThreadsStopped` can be true or false. After taking fabio's change we will have a consistent value for these.

int19h: Ah, I see.

Even after the change, we probably don't want this to check everything - after all, it isn't really meant to be a test of exceptions, and we'll have a separate test for those that does actually check every entry in the dict. For this test, we really only care that an exception happened with a particular message, and that's exactly what the body checks right now, so it's good!

pytests/func/test_flask.py
karthiknadig: will switch all of them to connection wait

pytests/helpers/session.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, method='launch', ptvsd_port=None, pid=None):
         self.debug_options = ['RedirectOutput']
         self.env = os.environ.copy()
         self.env['PYTHONPATH'] = PTVSD_SYS_PATH
+        self.cwd = None
+        self.expected_returncode = 0
+        self.cli_args = ()
```
int19h: Let's call this `program_args`, so that it's clear that these are what gets passed to the script/module.

int19h: And also make it an empty list, so that the tests can `+=` things to it.

pytests/helpers/session.py
```diff
@@ -204,7 +210,7 @@ def wait_for_disconnect(self, close=True):
         if close:
             self.timeline.close()
 
-    def wait_for_termination(self, expected_returncode=0):
+    def wait_for_termination(self, expected_returncode):
```
int19h: Since we now have `expected_returncode` as a property, I think it should be removed from the arguments for all these functions here that have it currently - tests should just set the property as needed. I don't think there are any other tests that pass it explicitly at the moment, so it should be an easy change.

pytests/helpers/webhelper.py
```diff
@@ -0,0 +1,42 @@
...
+
+@contextmanager
+def get_web_content(link, web_result=None, timeout=1):
+    web_client_thread = threading.Thread(
+        target=get_web_string_no_error,
+        args=(link, web_result),
+        name='test.webClient'
+    )
+    web_client_thread.start()
+    yield web_result
```
int19h: I think this helper looks somewhat confusing as a context manager - on the first glance, it feels like `web_result` is usable inside the with-statement, but in practice you can only get 'content' after the background thread has ended its execution, i.e. after with-statement. 

What this actually reminds me most of is `send_request` / `wait_for_response` - the part before `yield` is send, and the part after is `wait`. Maybe make it an object that works like that? i.e. something like:

```py
def get_web_content(link, web_result=None, timeout=1):
  class WebResponse(object):
    def __init__(self):
       self.content = None

    def wait_for_response(self):
       web_client_thread.join()
       return self.content

  response = WebResponse()
  web_client_thread = threading.Thread(
        target=get_web_string_no_error,
        args=(link, response),
        name='test.webClient'
  )
  return response
```
and change `get_web_string` so that it does `obj.content = ...` instead of `obj['content'] = ...'.

Then the test would do:
```py
web_request = get_web_content(...)
# Wait for breakpoint being hit etc. 
...
content = web_request.wait_for_response()
# Do asserts based on content
```

karthiknadig: That is better will change.

Issue #932, opened 2018-10-18
Switch tests to run on Tox


karthiknadig:

int19h:
There's still some work to do here to have Travis etc run everything via tox. They're directly invoking pytest currently.

int19h:
Fixed by #1031

PR #934, opened 2018-10-18
Automate pytest using tox

moulikcipherX:
fixed issue #932


test_requirements.txt
```diff
@@ -1,10 +1,29 @@
-setuptools
+atomicwrites
```
int19h: This list looks excessive. We certainly don't use most of them directly in the tests, and for dependencies of  dependencies, pip should take care of installing those (especially since they will change over time).

tox.ini
```diff
@@ -0,0 +1,7 @@
+[tox]
+envlist=py27,py35,py36,py37
+
+[testenv]
+changedir=tests
```
karthiknadig: should this be pytests?

test_requirements.txt
moulikcipherX: I have python 3.6 so all requirements is downlaoded itself in virtualenv so I run `pip freeze > test_requirements.txt`. I am not aware of these dependencies of dependencies.

tox.ini
moulikcipherX: I looked in `CONTROBUTING.md` for the directory of pytests so I found tests was the one.
`py -3.7 -m tests -v`.

moulikcipherX:
@karthiknadig guide me so that I can make this right ASAP

karthiknadig:
@moulikcipherX Change the directory to `pytests`, Ensure that `test_requirements.txt`  contains only the dependencies needed by the tests.

moulikcipherX:
Hello,
I have made changes that you wanted.

tox.ini
```diff
@@ -0,0 +1,7 @@
+[tox]
+envlist=py27,py35,py36,py37
```
karthiknadig: Needs python 3.4 here as well.

moulikcipherX:
Added python 3.4 testing in tox.ini @karthiknadig

PR #972, opened 2018-10-30
Fix #966: Sub process debugging not working with PTVSD

### LINKED ISSUE #966: Sub process debugging not working with PTVSD
## Environment data

- PTVSD version: Master
- OS and version: All
- Python version (& distribution if applicable, e.g. Anaconda): All

## Actual behavior

* child process is not spawned

## Expected  behavior

* Child process should be spawned and attached

## Steps to reproduce:
1. Install the latest VSC extension (dev)
2. Start debugging flask
3. Child process does not spawn

**Problem:**
* PTVSD is assumed to be in current path, when it isn't.
* PTSVD needs to use a launcher script instead of launching as a module.

int19h:
Allow launching ptvsd as script rather than as a module.

Use script mode in multiproc implementation, and in tests, to avoid explicit PYTHONPATH manipulation.


int19h:
@DonJayamanne, take a look at this. This should be sufficient to allow you to retire the launcher script on VSC side, since now you can just do:
```sh
python /foo/bar/ptvsd --client ...
```
i.e. specify the ptvsd directory by itself as the Python script, instead of using `-m`. It will then take care of patching `sys.path` as needed to load itself as a module and continue from there.

PR #974, opened 2018-10-30
Integrate single suspend event change to master

karthiknadig:


pytests/func/test_completions.py
```diff
@@ -1,3 +1,93 @@
...
+
+    debug_session.send_request('continue').wait_for_response()
+
+    targets.sort(key=lambda t: t['label'])
+    expected.sort(key=lambda t: t['label'])
+    assert targets == expected
+
+    debug_session.wait_for_exit()
+
+def test_outermost_scope(debug_session, pyfile):
```
int19h: These all look like they could be done as parametrized test, e.g.:
```py
expected_at_line = {
    11: [
        {'label': 'SomeClass', 'type': 'class'},
        {'label': 'someFunction', 'type': 'function'},
    ],
    ...
}

@pytest.mark.parametrize('bp_line', expected_at_line.keys())
def test_completion(debug_session, pyfile, bp_line):
    expected = expected_at_line[bp_line]
    ...
```

karthiknadig: will update

Issue #768, opened 2018-08-23
Re-enable frame eval mode in debugger


fabioz:
In a previous fix, the debugger frame eval mode was disabled, but it should be re-enabled later on.

-- see: https://github.com/Microsoft/ptvsd/issues/358#issuecomment-390966462 for more info on when the frame eval mode can be used and its limitations.

-- in the scenario where it works well as it has a very good performance (so, its use is limited but under the proper circumstances its overhead can be near zero -- i.e.: when entering a function call, the overhead to determine tracing is comparable to the tracing debugger, but afterwards, when inside of a function with a breakpoint, the regular mode needs to actually trace the function whereas the frame eval has zero overhead and acts as a programmatic breakpoint).

Note that to re-enable it, the root cause of the performance degradation when both the frame eval mode and the regular mode needs to be fixed (so, there should be a performance test to make sure there are no performance regressions after breakpoints are changed).

fabioz:
Will start taking a look at this one again (@karthiknadig, can you assign to me?)

karthiknadig:
Via #989

PR #989, opened 2018-11-06
Re-enable frame eval mode in the debugger. #768

fabioz:


fabioz:
Just to note, I'm in the process of updating this pull request given the latest changes and fixing errors on ci (with the latest version it seems there are tests failing on flask and completions, so, investigating it).

fabioz:
@karthiknadig this should be ready to integrate now.

fabioz:
Actually, a critical bug was found in this, so, please wait a bit more before integrating.

fabioz:
@karthiknadig the remaining items related to the frame eval should be fixed, so, this can be merged now.

PR #999, opened 2018-11-09
Decouple debug session and start method

### LINKED ISSUE #1001: Decouple debug session from launch method

karthiknadig:
Closes #1001

### DIFF
--- a/pytests/conftest.py
+++ b/pytests/conftest.py
@@ -129,7 +129,7 @@ if os.environ.get('PTVSD_SIMPLE_TESTS', '').lower() in ('1', 'true'):
     # Setting PTVSD_SIMPLE_TESTS locally is useful to not have to run
     # all the test permutations while developing.
     _ATTACH_PARAMS = [
-        'launch',
+        ('launch',),
     ]
 
     _RUN_AS_PARAMS = [
@@ -137,9 +137,10 @@ if os.environ.get('PTVSD_SIMPLE_TESTS', '').lower() in ('1', 'true'):
     ]
 else:
     _ATTACH_PARAMS = [
-        'launch',
-        'attach_socket',
-        # 'attach_pid',
+        ('launch',),
+        ('attach', 'socket', 'cmdline'),
+        ('attach', 'socket', 'import'),
+        ('attach', 'pid'),
     ]
 
     _RUN_AS_PARAMS = [
@@ -148,9 +149,9 @@ else:
     ]
 
 
-@pytest.fixture(params=_ATTACH_PARAMS)
+@pytest.fixture
 def debug_session(request):
-    session = DebugSession(request.param)
+    session = DebugSession()
     try:
         yield session
         try:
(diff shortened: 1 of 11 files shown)


pytests/conftest.py
```diff
@@ -172,35 +173,9 @@ def _run_as(request):
...
-
-    debug_session.send_request('continue').wait_for_response()
-    debug_session.wait_for_next(Event('continued'))
-
-    debug_session.wait_for_exit()
+@pytest.fixture(
+    name='starttype',
+    params=_ATTACH_PARAMS
+)
+def _start_method(request):
```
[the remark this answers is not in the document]
int19h: Or maybe better yet `start_method`, for consistency with `DebugSession`. And maybe rename `method` to `start_method` there as well?

pytests/helpers/session.py
```diff
@@ -461,20 +467,26 @@ def _wait_for_remaining_output(self):
         for thread in self._output_capture_threads:
             thread.join()
 
-    def common_setup(self, path, run_as, breakpoints=()):
+    def common_setup(self, path, starttype, run_as, breakpoints=(), backchannel=False):
```
int19h: I'm thinking that at this point there doesn't appear to be a clear distinction between `prepare_to_run` and `common_setup`. Also, we have a lot of inconsistency here in that some things are passed as arguments, while others (like `expected_returncode`) are object attributes.

I suggest replacing both with a single method - let's call it something like `initialize()` - that takes no arguments at all. All session parameters are configured consistently via object attributes. So, you'd do something like this:
```py
debug_session.start_method = ('launch',)
debug_session.target = 'script.py'
debug_session.run_as = 'file'
debug_session.breakpoints = [...]
debug_session.initialize()
```
or:
```py
debug_session.start_method = ('attach', 'socket')
debug_session.target = 'module_name'
debug_session.run_as = 'module'
debug_session.backchannel = True
debug_session.initialize()
```
As a shortcut, we can make it `initialize(**kwargs)`, which would then be treated simply as setting the corresponding attributes. Thus:
```py
debug_session.initialize(
     start_method = ('attach', 'socket'),
     run_as = 'module',
     target = 'module_name',
     backchannel = True,
)
```

int19h: Another idea is to combine `target` and `run_as` in the same manner as `start_method`. E.g.:
```py
debug_session.target = ('file', 'script.py')
debug_session.target = ('module', 'foo.bar')
debug_session.target = ('code', 'print("ok")')
```
and allow the second item in the tuple to be a function instead of a str, in which case `@pyfile` is automatically applied to it, and then the module name is derived from the filename, or raw code extracted from the file, depending on the first item. This would make it possible to write tests like `test_run` in a more generic way, so instead of this:
```py
    @pyfile
    def code_to_debug():
       ...

    if run_as == 'file':
        debug_session.prepare_to_run(filename=code_to_debug, backchannel=True)
    elif run_as == 'module':
        debug_session.add_file_to_pythonpath(code_to_debug)
        debug_session.prepare_to_run(module='code_to_debug', backchannel=True)
    elif run_as == 'code':
        with open(code_to_debug, 'r') as f:
            code = f.read()
        debug_session.prepare_to_run(code=code, backchannel=True)
    else:
        pytest.fail()
```
you can simply do:
```py
    def code_to_debug():
        ...
    debug_session.initialize(target=(run_as, code_to_debug))
```
and then `initialize` will contain the equivalent of the conditional above.

PR #1004, opened 2018-11-10
Add start method import

karthiknadig:


ptvsd/wrapper.py
```diff
@@ -2427,6 +2427,8 @@ def on_pydevd_thread_create(self, seq, args):
         # as well.
         with self.is_process_created_lock:
             if not self.is_process_created:
+                if not debugger_attached.isSet():
+                    return
```
karthiknadig: @int19h Need your input on this. Basically i was running into a race in tests where we would get `process` message before we even send `initialize` request. The process message came from this location. I think we should not send this event unless debugger attached.

int19h: Agreed.

pytests/func/test_run.py
```diff
@@ -12,24 +12,37 @@
...
-@pytest.mark.parametrize('start_method', [START_METHOD_LAUNCH, START_METHOD_CMDLINE])
+@pytest.mark.parametrize('start_method', [START_METHOD_LAUNCH, START_METHOD_CMDLINE, START_METHOD_IMPORT])
 def test_run(debug_session, pyfile, run_as, start_method):
     @pyfile
     def code_to_debug():
         import os
         import sys
         import backchannel
 
+        if os.getenv('PTVSD_ENABLE_ATTACH', False):
```
int19h: This should probably go into `DebugSession.initialize()`, so we can run other tests with import too without repeating it every time. The session can generate another temp script with `@pyfile` that does the import, and then invokes `target[1]` via the `runpy` module if `target[0]` is "file" or "module". For "code", we can just prepend "import ptvsd; ...;" directly to `target[1]`.

karthiknadig: That is a good idea, will update.

pytests/helpers/session.py
```diff
@@ -128,33 +128,39 @@ def stop(self):
...
-            argv += ['--wait']
-        else:
-            self._listen()
-            argv += ['--client']
+    def _get_argv_for_launch(self):
+        argv = [sys.executable]
+        argv += [ptvsd.__main__.__file__]
+        argv += ['--client']
         argv += ['--host', 'localhost', '--port', str(self.ptvsd_port)]
+        if self.multiprocess_port_range:
```
int19h: We don't have this in the product at the moment (it was there for a while as an artifact of an older multiproc implementation, but it's not needed for the current one).

pytests/helpers/session.py
```diff
@@ -178,10 +184,41 @@ def prepare_to_run(self):
...
+        print('Preparing debug session with method %r' % str(self.start_method))
+        argv = []
+        if self.start_method == START_METHOD_LAUNCH:
+            self._listen()
+            argv += self._get_argv_for_launch()
+        elif self.start_method == START_METHOD_CMDLINE:
+            argv += self._get_argv_for_attach_using_cmdline()
+        elif self.start_method == START_METHOD_IMPORT:
+            argv += self._get_argv_for_attach_using_import()
+            ptvsd_path = os.path.dirname(os.path.dirname(ptvsd.__main__.__file__))
```
int19h: Can you add a TODO comment here saying that we need to get rid of this after we fully switch to Tox? With Tox, ptvsd will be installed in the test environment as a proper package, so no PYTHONPATH patching should ever be necessary).

pytests/func/test_breakpoints.py
```diff
@@ -8,16 +8,16 @@
...
+from pytests.helpers.timeline import Event, Response
 from pytests.helpers.pathutils import get_test_root, compare_path
-from pytests.helpers.session import START_METHOD_LAUNCH, START_METHOD_CMDLINE
+from pytests.helpers.session import START_METHOD_LAUNCH, START_METHOD_CMDLINE, START_METHOD_IMPORT
 
 BP_TEST_ROOT = get_test_root('bp')
 
 
-@pytest.mark.parametrize('start_method', [START_METHOD_LAUNCH, START_METHOD_CMDLINE])
+@pytest.mark.parametrize('start_method', [START_METHOD_LAUNCH, START_METHOD_CMDLINE, START_METHOD_IMPORT])
```
int19h: This is going to be the new default for all tests, right? (i.e. unless a test has a good reason to exclude some method, it should cover all of them). If so, let's make the `start_method` fixture return this by default, and drop `@parametrize` on all tests that use the default.

karthiknadig: Agreed. Will make the change.

karthiknadig:
This looks like it keeps hitting the NoneType issue when importing ptvsd. I can't merge this until that issue is fixed.

int19h:
I was actually hoping to use this to repro it! Can you disable that start method for all tests, but merge the rest?

PR #1030, opened 2018-11-18
Migrate functional tests: part 1

### LINKED ISSUE #1032: Migrate to pytests: part 1
* Output tests
* Break point tests
* Thread tests
* Return value tests

karthiknadig:
Fixes #1032

### DIFF
--- a/pytests/func/test_breakpoints.py
+++ b/pytests/func/test_breakpoints.py
@@ -12,6 +12,7 @@ import sys
 from pytests.helpers.pathutils import get_test_root, compare_path
 from pytests.helpers.session import DebugSession
 from pytests.helpers.timeline import Event
+from pytests.helpers.pattern import ANY
 
 
 BP_TEST_ROOT = get_test_root('bp')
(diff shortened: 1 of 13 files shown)


pytests/func/test_breakpoints.py
```diff
@@ -57,3 +58,67 @@ def test_path_with_unicode(run_as, start_method):
 
         session.send_request('continue').wait_for_response(freeze=False)
         session.wait_for_exit()
+
+
+@pytest.mark.parametrize('condition_type, condition, value, hits', [
```
int19h: Can you make this get dependent params from a dict? (so that they don't show up in the test name needlessly)

pytests/func/test_breakpoints.py
```diff
@@ -57,3 +58,67 @@ def test_path_with_unicode(run_as, start_method):
...
+    with DebugSession() as session:
+        session.initialize(
+            target=(run_as, code_to_debug),
+            start_method=start_method,
+            ignore_unobserved=[Event('continued')],
+        )
+        session.send_request('setBreakpoints', arguments={
+                'source': {'path': code_to_debug},
+                'breakpoints': [{'line': bp_line, condition_type: condition}],
+            }).wait_for_response()
```
int19h: Indentation.

pytests/func/test_breakpoints.py
```diff
@@ -57,3 +58,67 @@ def test_path_with_unicode(run_as, start_method):
...
+        }).wait_for_response()
+        scopes = resp_scopes.body['scopes']
+        assert len(scopes) > 0
+
+        resp_variables = session.send_request('variables', arguments={
+            'variablesReference': scopes[0]['variablesReference']
+        }).wait_for_response()
+        variables = list(v for v in resp_variables.body['variables']
+                         if v['name'] == 'i')
+        assert len(variables) == 1
```
int19h: You can combine these into a single assert if you compare list to list:
```py
assert variables = Pattern([
  ANY.dict_with({ ... })
])
```py

File: pytests/func/test_breakpoints.py [hunk 4]
```diff
@@ -57,3 +58,67 @@ def test_path_with_unicode(run_as, start_method):
...
+        assert variables[0] == ANY.dict_with({
+            'name': 'i',
+            'type': 'int',
+            'value': value,
+            'evaluateName': 'i',
+        })
+
+        session.send_request('continue').wait_for_response(freeze=False)
+        i = 1
+        while i < hits:
```
int19h:
```py
for i in range(1, hits):
```

pytests/func/test_evaluate.py
```diff
@@ -235,3 +235,80 @@ def code_to_debug():
...
+        assert len(scopes) > 0
+
+        resp_variables = session.send_request('variables', arguments={
+            'variablesReference': scopes[0]['variablesReference']
+        }).wait_for_response()
+        variables = list(v for v in resp_variables.body['variables']
+                         if v['name'].startswith('(return)'))
+
+        assert len(variables) == 1
+        assert variables[0] == expected1
```
int19h:
```py
assert variables == Pattern([expected1])
```

int19h: (etc for other similar cases)

pytests/func/test_output.py
```diff
@@ -0,0 +1,36 @@
...
+    def code_to_debug():
+        from dbgimporter import import_and_enable_debugger
+        import_and_enable_debugger()
+        print('Hello\tWorld')
+
+    with DebugSession() as session:
+        session.initialize(target=(run_as, code_to_debug), start_method=start_method)
+        session.start_debugging()
+        session.wait_for_exit()
+        data = session.get_stdout_as_string() + session.get_stderr_as_string()
```
int19h: We probably want to check either channel separately. You can do `print(file=sys.stderr)` in the test to exercise stderr.

Or is there some outstanding issue where they get mixed up?

pytests/func/test_threads.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+from pytests.helpers.timeline import Event
+from pytests.helpers.session import DebugSession
+
+
+@pytest.mark.parametrize('args, count', [
+    ('single', 1),
+    ('multi', 3),
+])
+def test_multi_thread(pyfile, run_as, start_method, args, count):
```
int19h: `count` is a dependent arg, so it should go into a dict.

Also, these names are a bit confusing at a glance, as in "multi-threaded single thread". :) Maybe "one" and "many"?

pytests/func/test_output.py
karthiknadig: There is no mixup. Just checking stdout should be enough.

int19h: Shouldn't we check stderr too tho, just to make sure it gets captured as well?

karthiknadig: Actually this is wrong. This should check the output event.

karthiknadig: This test is specifically for tab character. There was an odd behavior with that one. I will add more output tests and cover stderr output there

PR #1049, opened 2018-11-21
Tests and fixes

### LINKED ISSUE #1047: Support both expression and log breakpoints
CTI found a way to set both but the conditional breakpoint seems to end up being ignored.

karthiknadig:
Fixes #1047


karthiknadig:
This is broken.

PR #1057, opened 2018-11-29
More tests

karthiknadig:


pytests/func/test_disconnect.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.parametrize('start_method', ['launch'])
+@pytest.mark.skip(reason='Bug #1052')
+def test_exit_on_disconnect_for_launch(pyfile, run_as, start_method):
+    @pyfile
+    def code_to_debug():
+        from dbgimporter import import_and_enable_debugger
+        import_and_enable_debugger()
+        import backchannel
+        import atexit
+        def onexit():
```
int19h: I'm not sure we want to test this. It does currently call `sys.exit` on disconnect in this scenario, but I think it's not really the right thing - in other contexts, when process is terminated by the debugger, it's an immediate termination that does not attempt to do any cleanup, and `sys.exit` is supposed to be reserved for normal exits. Consequently, in the refactor branch, disconnect does `os._exit` now instead, and `atexit` callbacks aren't called.

pytests/func/test_vs_specific.py
```diff
@@ -0,0 +1,97 @@
...
+            ignore_unobserved=[Event('stopped'), Event('continued')],
+        )
+        session.set_breakpoints(module2, [bp_line])
+        session.start_debugging()
+
+        session.wait_for_thread_stopped()
+        modules = session.all_occurrences_of(Event('module'))
+        modules = [(m.body['module']['name'], m.body['module']['path']) for m in modules]
+        assert modules[:3] == [
+            ('module2', ANY.such_that(lambda s: compare_path(module2, s))),
```
int19h: Looks like it's time for `ANY.path`!

pytests/func/test_disconnect.py
karthiknadig: I will test this in a different way. All we want to do is ensure that the debugger exits after disconnect in launch scenario.

Issue #1000, opened 2018-11-09
Debug console autocompletion not working


DonJayamanne:
_From @darrickyee on November 9, 2018 16:21_

## Environment data

- VS Code version: 1.28.2
- Extension version (available under the Extensions sidebar): 2018.10.0
- OS and version: Windows 7
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.0
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A

## Actual behavior

Autocompletion in the debug console console doesn't seem to work on anything except the first name typed.

Autocomplete works when an object is the first thing typed:
![debug1](https://user-images.githubusercontent.com/18172544/48274399-3768eb80-e411-11e8-8de2-efb22f9b8fc1.jpg)

No autocompletion if the same object is on the other side of an assignment, passed as a parameter, or even just enclosed in brackets:
![debug2](https://user-images.githubusercontent.com/18172544/48274484-6a12e400-e411-11e8-9559-50606887aba0.jpg)

Is this the intended behavior?

_Copied from original issue: Microsoft/vscode-python#3282_

karthiknadig:
I don't think this is a regression. The completions were always pretty simple (i.e., the first case). I don't think it ever handled the second case. 
/cc @fabioz

fabioz:
I'll investigate it here (i.e.: it seems to work for me in plain pydev/pydevd, so, it may be some integration issue).

fabioz:
The problem is that the completion command only expects the token for the completion and not the full text entered (that part was expected to be handled by the IDE)... 

I'm in the process of changing the related message to use the DAP json format so that it has the full info on the text/column so that it can do what's needed properly. 

As a note, I'm doing it on top of the changes in the pull request: https://github.com/Microsoft/ptvsd/pull/1070 (so, please don't rebase when merging).

PR #1070, opened 2018-12-13
Provide json as the output of messages from pydevd -- initial proof of concept dealing with only a couple of messages (#1039).

### LINKED ISSUE #1039: VSCode hangs when debugging with step into (F11) import statement
Original issue can be found here https://github.com/Microsoft/vscode-python/issues/3069

fabioz:
This is the initial change which allows pydevd to send json messages.

Right now only 2 messages are changed to send json, but if the approach is Ok I'll start to port other messages.

Some notes:
- The json format is close to the debug adapter protocol but not exactly the same... for instance, I'm passing a `pydevd_cmd_id` on json messages so that the dispatch code in ptvsd doesn't have to change.
- To generate the json messages I'm using code I had around some time ago which generates code from the debug adapter json specification.
- The plan is converting the messages sent from pydevd, not the received messages (which is what's needed for fix #1039).


fabioz:
Note: I'm doing the fix for https://github.com/Microsoft/ptvsd/issues/1000 on top of this pull request, so, please don't rebase this pull request when merging.

PR #1084, opened 2019-01-03
Attempt to fix exceptions thrown while debugging

jeffrimko:
This PR includes two attempts at fixing exceptions I have seen while debugging a Python CLI application in VS Code. 

The change to `_local.py` is for instances when `ex.code` is not an integer. According to the [docs here](https://docs.python.org/3/library/exceptions.html#SystemExit.code), the value might be an error message. Some libraries do return the usage message such as [Docopt](https://github.com/docopt/docopt/blob/master/docopt.py#L29).

The change to `pydevd_io.py` is for an `OSError: [WinError 87] The parameter is incorrect` exception that is thrown when trying to debug code that includes code line `print("", end="", flush=True)`. I don't fully understand what causes this exception to be thrown. The try-catch prevents the exception from interrupting the debug session but further investigation may be warranted. I am running Python 3.6.2 on Windows 10 Home version 1803.

### DIFF
--- a/src/ptvsd/_local.py
+++ b/src/ptvsd/_local.py
@@ -139,5 +139,8 @@ def _run(argv, addr, _pydevd=pydevd, _install=install, **kwargs):
     try:
         _pydevd.main()
     except SystemExit as ex:
-        daemon.exitcode = 0 if ex.code is None else int(ex.code)
+        try:
+            daemon.exitcode = 0 if ex.code is None else int(ex.code)
+        except ValueError:
+            daemon.exitcode = 1
         raise
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_io.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_io.py
@@ -34,7 +34,10 @@ class IORedirector:
 
     def flush(self):
         for r in self._redirect_to:
-            r.flush()
+            try:
+                r.flush()
+            except OSError:
+                pass
 
     def __getattr__(self, name):
         for r in self._redirect_to:
@@ -56,7 +59,7 @@ class IOBuf:
         b = self.buflist
         self.buflist = []  # clear it
         return ''.join(b)  # bytes on py2, str on py3.
-    
+
     def write(self, s):
         if not IS_PY3K:
             if isinstance(s, unicode):


jeffrimko:
Started looking into why the "ptvsd-testing-automation" step failed but after reviewing some of the older builds, looks like it wasn't due to any of my changes.

fabioz:
Hi @jeffrimko, thanks for the pull request...

Now, regarding the exception in `pydevd_io.py`, that seems tricky as it may be masking real errors... is it possible for you to provide code which reproduces that?

i.e.: do you get that if you run that `print` on a clean module or do you do other things? Is it possible that some setup somewhere else in your code may be making things go awry? Maybe a `reload(sys)` or something else which deals with sys and its streams?

Regarding the exitcode, the docs (https://docs.python.org/2/library/exceptions.html#exceptions.SystemExit / https://docs.python.org/3/library/exceptions.html) say that it can actually be anything (so, not only a string/int)... so, the proper fix would be checking if it's an `int` (or `int/long` on `PY2`) and if it's not, always set the exit code to 1 (so, getting the ValueError is not enough as other errors may be thrown on `int(some_object)` -- if it's a string it shouldn't do the `int` as it's doing now and should just set the exit code to 1).

jeffrimko:
For recreating the `pydevd_io.py`, I can recreate by debugging a Python script containing only `print("", end="", flush=True)`, see following screenshot:

recreate-pydevd_io

After some quick testing, it seems due to the print statement not actually printing anything. For example, if I change the line to `print(" ", end="", flush=True)` then I don't see the issue. Weird.

Ultimately, this exception was found due to importing [this library](https://github.com/jeffrimko/Qprompt/blob/master/lib/qprompt.py#L240) but maybe there is a better way to do the print() check. I can look into this a bit more later along with cleaning up the `SystemExit` int checking.

fabioz:
As a note, I'm able to reliably reproduce the `print` error in Python 3.6.2 on Windows but not Python 3.6.5 (so, I think we're working around a problem which is already fixed in Python itself).

jeffrimko:
@fabioz That's for the info! I will check on my machine as well later today.

jeffrimko:
@fabioz Confirmed on my machine that updated versions of Python fix the print/flush issue. Reverted those changes. Updated the SystemExit code checking a logic per your suggestions.

PR #1088, opened 2019-01-07
Removing old test code

karthiknadig:


int19h:
While we're at it, can you also rename `pytests` to `tests`? That's the standard convention for the test suite, and the only reason why it was different is to accommodate the old tests without interfering with them. 

(Will probably need a bunch of find/replace fixes in `import` statements, and a one-line change in pytest.ini, but shouldn't affect anything else.)

PR #1091, opened 2019-01-09
Add support for step over my code / step return my code to pydevd. Fixes #477.

### LINKED ISSUE #477: pydevd: Step  over and step return should support just-my-code

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
```diff
@@ -410,10 +411,13 @@ def trace_dispatch(self, frame, event, arg):
 
             if is_exception_event:
                 breakpoints_for_file = None
-                # CMD_STEP_OVER = 108
-                if stop_frame and stop_frame is not frame and step_cmd == 108 and \
+                # CMD_STEP_OVER = 108, CMD_STEP_OVER_MY_CODE = 159
+                if stop_frame and stop_frame is not frame and step_cmd in (108, 159) and \
```
karthiknadig: >108, 159 [](start = 75, length = 8)

Why numeric values here, instead of CMD_*?

karthiknadig:
@fabioz Looks good for the most part. I had one comment.

fabioz:
This is a micro-optimization... this is is the most executed path on the
debugger when stepping is enabled... although the case for the exception is
probably a stretch it did have an impact for line events.

PR #1100, opened 2019-01-15
Ignore any file starting with '<' in the debugger. Fixes #1071

### LINKED ISSUE #1071: Stepping over fails in debugger with message Unable to open '&lt;string&gt;' FIle not found
@malmaud commented on [Wed Nov 21 2018](https://github.com/Microsoft/vscode-python/issues/3426)

Getting this error when debugging this file:

```python
import typing
from dataclasses import dataclass, field


@dataclass
class Thing:
    a: typing.Any
    b: typing.Any
    c: typing.Any = field(init=False, repr=False)

    def __post_init__(self):
        self.c = self.a + self.b


Thing(a=1, b=2)
```

If I set a breakpoint at the `self.c = ` line and then click 'step over', I see 

```
Unable to open '&lt;string&gt;': File not found (file:///Users/malmaud/Dropbox (MIT)/Desktop/fun/&lt;string&gt;).
```

I guess this is related to https://github.com/Microsoft/vscode-python/issues/3063.

VSCode 1.29.1
Python 3.6.7 with `dataclasses` package  installed via ` pip install dataclasses`
Extension 2018.10.1

Screenshot:
screen shot 2018-11-21 at 11 45 30 am

---

@DonJayamanne commented on [Wed Nov 21 2018](https://github.com/Microsoft/vscode-python/issues/3426#issuecomment-440897012)

What's the name of the file you are debugging

---

@malmaud commented on [Wed Nov 21 2018](https://github.com/Microsoft/vscode-python/issues/3426#issuecomment-440897509)

`/Users/malmaud/Dropbox (MIT)/Desktop/fun/debug.py`

fabioz:
This didn't work out well because tests will launch files with `<string>` paths.

Apparently this is because when calling Python with `-c` Python itself will use `<string>` to run code.

i.e.: 
```
>>> python -c "import sys;print(sys._getframe().f_code.co_filename)"
<string>
```

So, I'm not entirely sure on the best way to proceed -- note that in that case, the user wouldn't be able to get to the `<string>` contents either and it'd fail with the same issue reported on #1071... a way out could be create a temporary file for the tests and run that temporary file instead of relying on `-c` which would result in the filename mapping to `<string>`.

Suggestions @int19h @karthiknadig ?

-- note: I got this issue because I thought it was very simple to fix... if it's going to be much more work it may need to be postponed.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/pydevd.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd.py
@@ -534,9 +534,10 @@ class PyDB(object):
                 If it's a regular user file which should be traced.
         '''
         basename = abs_real_path_and_basename[-1]
-        if basename.startswith('<frozen '):
+        if basename.startswith('<'):
             # In Python 3.7 "<frozen ..." appear multiple times during import and should be
-            # ignored for the user.
+            # ignored for the user and <string> is commonly used for dummy files, so, consider
+            # any file starting with '<' as invisible for the debugger.
             return self.PYDEV_FILE
         return self._dont_trace_get_file_type(basename)
 
--- a/src/ptvsd/_vendored/pydevd/tests_python/debug_constants.py
+++ b/src/ptvsd/_vendored/pydevd/tests_python/debug_constants.py
@@ -7,6 +7,7 @@ TEST_JYTHON = os.getenv('PYDEVD_TEST_JYTHON', None) == 'YES'
 
 IS_PY3K = sys.version_info[0] >= 3
 IS_PY36_OR_GREATER = sys.version_info[0:2] >= (3, 6)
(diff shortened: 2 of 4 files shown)


karthiknadig:
@int19h and I had discussed this issue a while ago. One possible way is the change `compile` so that it always has the file field populated. This should be done in a way such that when the code is suspended, we known from the path field that this is `-c` code. We could then use source reference to show this to user as a "file". This is a common problem when executing code in a notebook environment. Although, I am not sure of the feasibility of this solution, and i think it is currently beyond the scope of the PR.

fabioz:
Ok, will close this pull request until that's sorted out.

As a note, I think that the compile could always create a temp file with the contents and then compile with that file (and where `-c` is used, don't actually use `-c`, but create a temp file and execute it).

PR #1104, opened 2019-01-16
Restrict wait on exit tests to launch only

karthiknadig:


karthiknadig:
@int19h I want to merge this. It should help with the macOS failures.

PR #1105, opened 2019-01-17
Port refactored command line parser to master

int19h:


tests/func/test_args.py
```diff
@@ -13,6 +13,7 @@ def code_to_debug():
         import sys
         from dbgimporter import import_and_enable_debugger
         import_and_enable_debugger()
+        print(sys.argv)
```
karthiknadig: print for debug?

int19h: Sorry, didn't notice this before. Yes, it's diagnostic, but since it's a test, it doesn't hurt to have it here - this way we get the whole thing in the log if it fails in an automated run.

Issue #997, opened 2018-11-08
Excluding files from being debugged


DonJayamanne:
## Environment data

- PTVSD version: Master
- OS and version: N/A
- Python version (& distribution if applicable, e.g. Anaconda): N/A
- Using VS Code or Visual Studio:VSC

## Actual behavior

Unable to prevent files from being excluded from the debugger.

## Expected  behavior

Should be able to exclude files from being debugged.

## Steps to reproduce:
Currently VSC is using the following API to exclude a file from being debugged:
```
import ptvsd
ptvsd.debugger.DONT_DEBUG.append(os.path.normcase(__file__))
```

karthiknadig:
Was this ever supported?

DonJayamanne:
Yes it was

karthiknadig:
@int19h I don't see `DONT_DEBUG` being used anywhere? was this supported in 3.* but never in 4.*?

DonJayamanne:
Can we increase the priority of this issue, thanks.
Or does the label `feature-parity` have an implicit priority attached to it.

qubitron:
This is important for VS Code so adding back P1

fabioz:
I'll take a look at this.

The suggestion is being able to configure through the launch config:

```
{
    "name": "Attach (Remote Debug)",
    "type": "python",
    "request": "attach",
    "port": 5678,
    "host": "localhost",
     "exclude": ["*bar.py", "*dir3*"].
     "includeDirs": ["/home/user/dir2"].
     "excludeDirs": ["/home/user/dir2/dir1"].
}
```

Some notes:
`exclude` can be any name (being directories or not) and has precedence over anything (so, if it matches it's always excluded).

`includeDirs` and `excludeDirs` need to map to directories and it's possible that one is inside the other (so, the one which matches the biggest path wins).

Does that sound reasonable? Is a Python API required too?

karthiknadig:
That sound good. We will need python API as well.

qubitron:
My understanding at this time is that we only need the python API at this time.

We have another proposal for more specific include/exclude rules here:
https://github.com/Microsoft/vscode-python/issues/2087#issuecomment-425528613

Copying:
```
justMyCode.excludeModules: list of modules to exclude from my code
justMyCode.excludePaths: list of paths to exclude from my code (defaults to a $sitepackages variable, which would need to be added)
justMyCode.includeModules: list of modules to include as my code
justMyCode.includePaths: list of folders to include as my code
[...]

PR #1119, opened 2019-01-23
Exclude files from being debugged. #997

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -421,8 +424,8 @@ def remove_python_exception_breakpoint(self, py_db, exception):
         except:
             pydev_log.debug("Error while removing exception %s" % sys.exc_info()[0])
 
-        py_db.on_breakpoints_changed(remove=True)
-    
+        py_db.on_breakpoints_changed(removed=True)
```
karthiknadig: Can you make this a separate PR? We are planning on making a release today and would like to take this.

fabioz:
Hi @karthiknadig, unfortunatelt I won't be able to do it today, only
tomorrow...

I think you can cherry-pick only the related commit if you need the fix
today, otherwise I can make it tomorrow.

karthiknadig:
>  I think you can cherry-pick only the related commit if you need the fix today, otherwise I can make it tomorrow.

Will do no worries.

fabioz:
Just as a note, I did one more bugfix in
https://github.com/Microsoft/ptvsd/pull/1121 which may be nice having in
the release if you still haven't done it...

fabioz:
Note: rebased and did an additional fix on step return when filtering, so, should be ready to review.

PR #1120, opened 2019-01-24
Pass 'removed' instead of 'remove' to on_breakpoints_changed. #1118

### LINKED ISSUE #1118: TypeError: on_breakpoints_changed() got an unexpected keyword argument 'remove'
Happens whenever an exception breakpoint is added or removed with the current bits. Full stack:
```
  File "/home/int19h/.local/lib/python3.7/site-packages/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command.py", line 47, in process_net_command
    cmd = on_command(py_db, cmd_id, seq, text)
  File "/home/int19h/.local/lib/python3.7/site-packages/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command.py", line 461, in cmd_remove_exception_break
    self.api.remove_python_exception_breakpoint(py_db, exception)
  File "/home/int19h/.local/lib/python3.7/site-packages/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py", line 424, in remove_python_exception_breakpoint
    py_db.on_breakpoints_changed(remove=True)
TypeError: on_breakpoints_changed() got an unexpected keyword argument 'remove'
```
Unclear why, since `on_breakpoints_changed` does have that argument declared. 

Happens on both Python 2 and Python 3.

karthiknadig:
Fixes #1118

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
@@ -421,8 +421,8 @@ class PyDevdAPI(object):
         except:
             pydev_log.debug("Error while removing exception %s" % sys.exc_info()[0])
 
-        py_db.on_breakpoints_changed(remove=True)
-    
+        py_db.on_breakpoints_changed(removed=True)
+
     def remove_plugins_exception_breakpoint(self, py_db, exception_type, exception):
         # I.e.: no need to initialize lazy (if we didn't have it in the first place, we can't remove
         # anything from it anyways).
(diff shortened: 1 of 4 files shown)


karthiknadig:
This seems to be causing a hang in MacOS 2.7 tests. but if I take the bigger change it seems to be fine. Is there any change that I must get along with this?

fabioz:
Not really, just this change should be Ok and I think the issue is some case which is already flaky on master (the given change only affects removing breakpoints and as far as I can tell the test which is timing out doesn't remove any breakpoint).

PR #1127, opened 2019-01-28
Remove breakpoints from all mappings. Fixes #1126

### LINKED ISSUE #1126: Removing breakpoints in remote debugging does not work
@DonJayamanne commented on [Mon Jan 28 2019](https://github.com/Microsoft/vscode-python/issues/4176)

Launch python code for debugging:
* Mac, Python3.6
* Ptvsd version 4.2.2
```python
import ptvsd
ptvsd.enable_attach(('localhost', 3000))

import time
for i in range(1000):
    time.sleep(1)
    print(i)
```

* Launch program
* Attach debugger
* Add a breakpoint to the `print` line
* Breakpoint hits
* Remove breakpoint
* Continue
* Program stops at `print` line again, even when there are no breakpoints!!!
* Disconnect debugger, program is stuck!!

**Two problems:**
* Program stops at `print` line again, even when there are no breakpoints!!!
* Disconnect debugger, program is stuck!!

fabioz:
@karthiknadig this should fix the issue, but unfortunately I won't have time to do new tests for this today... (I'll leave it up to you on whether you want to put as is, include tests yourself or wait for me to do it tomorrow).

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
@@ -289,8 +289,18 @@ class PyDevdAPI(object):
         Removes all the breakpoints from a given file or from all files if filename == '*'.
         '''
         changed = False
-        for file_to_id_to_breakpoint in [
-            py_db.file_to_id_to_line_breakpoint, py_db.file_to_id_to_plugin_breakpoint]:
+        lst = [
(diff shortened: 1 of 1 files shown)


fabioz:
Ok, did some more code-review myself after the initial fix... should be ready now ;)

karthiknadig:
Add the tests as well. I will take this with the tests.

karthiknadig:
We decided to take this first. We will take the tests in a followup PR.

PR #1132, opened 2019-01-29
Fix #1130: New debugger displays warnings about `speeds using cython…not found`

### LINKED ISSUE #1130: New debugger displays warnings about `speeds using cython not found`
@DonJayamanne commented on [Mon Jan 28 2019](https://github.com/Microsoft/vscode-python/issues/4190)

Old version of PTVSD didn't display this warning, new version of PTVDS does.

Here's the old version:
screen shot 2019-01-28 at 4 46 12 pm

Here's the output from the new version:
screen shot 2019-01-28 at 4 46 54 pm

I consider this a P0, as we're displaying warnings that we didn't (removed) in previous versions. This will end up confusing users.

int19h:
Comment out the prints in pydevd.


int19h:
For the time being, I'll fix this directly in pydevd code to avoid printing these out. But going forward, we need to come up with some consistent way to disable these messages on all code paths inside pydevd with a single flag or similar. (The flag that we were using before is not the right one for this anymore, now that we build with Cython modules.)

PR #1133, opened 2019-01-29
Handle subscripting a potentially None object.

karthiknadig:
@fabioz This change seems to mitigate the issue [here](https://github.com/Microsoft/ptvsd/issues/1126#issuecomment-458400958). If this isn't the right fix, go ahead a provide a PR.


fabioz:
Hi @karthiknadig, I did the fix along with the tests in https://github.com/Microsoft/ptvsd/pull/1134 (so, closing this one).

PR #1137, opened 2019-01-31
Fix #1093: breakpoint are not hit when debugger is attached to remote target

### LINKED ISSUE #1093: breakpoint are not hit when debugger is attached to remote target
## Environment data

- PTVSD version: 4.2.0
- OS and version: Linux, Fedora 29
- Python version: Python 2.7.15
- Using VS Code or Visual Studio:
Version: 1.30.2
Commit: 61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8
Date: 2019-01-07T22:49:48.319Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
OS: Linux x64 4.19.13-300.fc29.x86_64

Target machine:
- OS and version: Linux, Ubuntu 18.04LTS
- PTVSD version: 4.2.0
- Python 2.7.15rc1

## Actual behavior
I try to debug remotely python script with vscode. In my launch.json i have proper configuration including pathMappings.

I have try following cases:
1. Run script on target machine like:
python -m ptvsd --single-session --host 10.3.72.215 --port 3000 --wait --multiprocess script-debug
On vscode i put break point and attach to target. As result script-debug is executed without hitting break point and on target machine i see message:
pydev debugger: warning: trying to add breakpoint to file that does not exist: "settings from localRoot /script-debug"
 Some times i can manage to start debugging - i.e. debugger stops at BP, but this is less than 1% of all attempts.

2. Run script on target but modified:
`_script-debug_`
```py
import os
import sys
import ptvsd

# # Allow other computers to attach to ptvsd at this IP address and port.
ptvsd.enable_attach(address=('0.0.0.0', 3000), redirect_output=True)

# # Pause the program until a remote debugger is attached
ptvsd.wait_for_attach()
```

Same behavior like p.1 

3. Run script on target but more modified:
`_script-debug_`
```py
import os
import sys
import ptvsd

# # Allow other computers to attach to ptvsd at this IP address and port.
ptvsd.enable_attach(address=('0.0.0.0', 3000), redirect_output=True)

# # Pause the program until a remote debugger is attached
ptvsd.wait_for_attach()
ptvsd.break_into_debugger()
```

In this case 100% of attempts as successful if i remove/disable all BP from vscode and add/enable them once debugger is attached.

If i start debugger with breakpoints, that are set before attachment to remote target, again see message on remote site:
pydev debugger: warning: trying to add breakpoint to file that does not exist: "settings from localRoot /script-debug"
So each time before debug session i have to disable BP, attach to target and re-enable them in order to use debugger.

Looks like some kind of race condition when debugger is attached and pathMappings, and BP are set to remote side.

## Expected  behavior
To be possible to run target like 'python -m ptvsd --single-session --host 10.3.72.215 --port 3000 --wait --multiprocess script-debug" and not to be forced to use ptvsd.break_into_debugger() in order to set BP

Already configured BP to work without needs to be re-enabled.

int19h:
Do not allow 'setBreakpoints' request until either 'launch' or 'attach' request is received.

Wait for 'launch' or 'attach' request to process path mappings before handling 'setBreakpoints'.

### DIFF
--- a/src/ptvsd/wrapper.py
+++ b/src/ptvsd/wrapper.py
@@ -1213,7 +1213,7 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
         multiproc.root_start_request = request
         self.start_reason = 'attach'
         self._set_debug_options(args)
-        self._handle_attach(request, args)
+        self._handle_launch_or_attach(request, args)
         self.send_response(request)
 
     def on_launch(self, request, args):
@@ -1221,7 +1221,7 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
         self.start_reason = 'launch'
         self._set_debug_options(args)
         self._notify_launch()
-        self._handle_launch(request, args)
+        self._handle_launch_or_attach(request, args)
         self.send_response(request)
 
     def on_disconnect(self, request, args):
@@ -1302,10 +1302,7 @@ class VSCLifecycleMsgProcessor(VSCodeMessageProcessorBase):
     def _handle_configurationDone(self, request, args):
         pass
 
-    def _handle_attach(self, request, args):
-        pass
-
-    def _handle_launch(self, request, args):
+    def _handle_launch_or_attach(self, request, args):
         pass
 
     def _handle_detach(self):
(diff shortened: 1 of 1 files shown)


src/ptvsd/wrapper.py
```diff
@@ -2276,6 +2268,14 @@ def _get_bp_type(self, path):
 
     @async_handler
     def on_setBreakpoints(self, request, args):
+        if not self._path_mappings_received:
+            self.send_error_response(request, "'setBreakpoints' request must be issued after 'launch' or 'attach' request.")
+
```
karthiknadig: missing return?

int19h: Indeed!

PR #1140, opened 2019-02-05
Add some tests for recent fixes

karthiknadig:


tests/func/test_breakpoints.py
```diff
@@ -346,3 +345,36 @@ def test_package_launch():
...
+        hit = session.wait_for_thread_stopped()
+        frames = hit.stacktrace.body['stackFrames']
+        assert bp_line == frames[0]['line']
+
+        # remove breakpoints in file
+        session.set_breakpoints(code_to_debug, [])
+        session.send_request('continue').wait_for_response(freeze=False)
+        session.wait_for_exit()
+
+        # NOTE: Due to https://github.com/Microsoft/ptvsd/issues/1028 we may not get
```
int19h: I think you can take care of this by using a backchannel to sync between test and debuggee - it just needs to do `read_json` before exit, and the test can then wait until the last output event shows up (by checking `output == '10'`).

karthiknadig: Will try. If this works, I will enable this for some of the other test cases.

karthiknadig:
This runs locally seems to crash on

Issue #1142, opened 2019-02-06
Log points treated as breakpoints


karthiknadig:
## Environment data

- PTVSD version: master
- OS and version: any
- Python version (& distribution if applicable, e.g. Anaconda): any
- Using VS Code or Visual Studio: VSC

## Actual behavior

Debugger breaks at a log point.

## Expected  behavior

Should not break at log points.

## Steps to reproduce:
1. use following code:
```py
a = 10
for i in range(1, a):
    print('value: %d' % i)
```
2. Set log point is VSC at line 3.
![image](https://user-images.githubusercontent.com/3840081/52374840-27ae7280-2a13-11e9-9f26-31d9b4a9cb9b.png)

3. Hit Run.

DonJayamanne:
Ok great, this is a regression.
We have users reporting this issue in our extension.
Curious how the tests failed!

fabioz:
I'll take a look at this.

fabioz:
Ok, just provided a pull request... now, I have a question regarding log points:

What's the proper behavior if there's a log point with a condition?

Should it:
`1.` Show log messages every time the line is executed and stop on the conditional breakpoint.
`2.` Just show a log when the given condition is matched and and never stop on the breakpoint.
`3.` Something else.

The specification (https://microsoft.github.io/debug-adapter-protocol/specification) says the following regarding `logMessages`:

`If this attribute exists and is non-empty, the backend must not 'break' (stop) but log the message instead. Expressions within {} are interpolated.`

Right now `1.` is implemented because there was a test for this condition but the docs aren't really clear on the interaction with conditions and I was inclined to think it should be `2.`, so, just double-checking (note: tests were skipped due to something else, which is why they haven't been able to get the regression -- the pull request re-enables the tests and should fix their flakiness).

DonJayamanne:
> What's the proper behavior if there's a log point with a condition?

It is my understanding that this isn't possible.

fabioz:
This is possible if you add the breakpoint and edit it later...

![breakpoint_edit](https://user-images.githubusercontent.com/117621/52429780-e50c9900-2aeb-11e9-8c0f-dc4f2a8d81a5.png)

fabioz:
Given that this is pretty obscure I think we can leave as is implemented now (we can come back to it if someone complains about it or if the spec is rectified).

DonJayamanne:
@fabioz 	I tried that with node, and it doesn't work.
[...]

PR #1143, opened 2019-02-07
Fix: #1142: Log points treated as breakpoints.

fabioz:


karthiknadig:
I am seeing a NoneType exception that I had not seen in the previous runs. It might be worth looking at.
```
Exception in thread ptvsd.EventLoop (most likely raised during interpreter shutdown):
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/threading.py", line 801, in __bootstrap_inner
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/threading.py", line 754, in run
  File "/home/vsts/work/1/s/.tox/py27/lib/python2.7/site-packages/ptvsd/futures.py", line 121, in run_forever
<type \'exceptions.TypeError\'>: \'NoneType\' object is not callable
```

fabioz:
I have actually seen this many times already. The issue is that the ptvsd
daemon threads are not stopped  before interpreter shutdown, so, this is
kind of expected (they are being executed while python is setting all the
module variables to None).

Just catching the exception and ignoring it would be ok too.

PR #1147, opened 2019-02-12
 Fix #841: BreakOnSystemExitZero debug option is not respected

### LINKED ISSUE #841: BreakOnSystemExitZero debug option is not respected
## Environment data

- PTVSD version: 4.1.3
- OS and version: Windows 10

## Actual behavior

Debugger breaks on `sys.exit(0)` even when the option to break on zero is disabled:

image

## Expected  behavior

Debugger only breaks on `sys.exit(0)` if the option is enabled.

## Steps to reproduce:
1. In tools, options, python, debugging, uncheck "Break on SystemExit exception with exit code of zero"
2. Debug this code:
```
sys.exit(0)
```

Edit: FYI all Python exceptions were unchecked:
image

int19h:


tests/conftest.py
```diff
@@ -89,6 +96,16 @@ def script_file():
...
+
+    Specific lines can be marked with a comment like so:
+
+        print('fizz') #@fizz
+
+    The returned string value has an additional attribute 'lines', which is
+    a dict of all marked lines - the part after #@ becomes the key, and the
+    corresponding line number is the value. This can be used to set breakpoints
+    or to assert line numbers for exceptions and stack traces in a way that
+    does not require updating them as debugged code changes.
```
int19h: @karthiknadig, check this out, I think this'll come in handy in quite a few tests.

int19h: Just realized that this approach won't work for tests that use named files. Better to have it as a separate function that opens a file by name and extracts that info, then it's fully reusable.

PR #1148, opened 2019-02-12
Fix #1009: Provide a public API to register threads not created by Python for debugging

### LINKED ISSUE #1009: Provide a public API to register threads not created by Python for debugging
This can be done by simply wrapping `pydevd.settrace(suspend=False)`

int19h:


tests/func/test_threads.py
```diff
@@ -49,3 +51,53 @@ def worker(tid, offset):
...
+            assert windll.kernel32.CreateThread(c_void_p(0), c_size_t(0), thread_func, c_void_p(0), c_uint32(0), c_void_p(0))
+        elif platform.system() == 'Linux' or platform.system() == 'Darwin':
+            from ctypes import CDLL, CFUNCTYPE, byref, c_void_p, c_ulong
+            from ctypes.util import find_library
+            libpthread = CDLL(find_library('libpthread'))
+            thread_func_p = CFUNCTYPE(c_void_p, c_void_p)
+            thread_func = thread_func_p(foo) # must hold a reference to wrapper during the call
+            assert not libpthread.pthread_create(byref(c_ulong(0)), c_void_p(0), thread_func, c_void_p(0))
+        else:
+            pytest.skip('Not supported on ' + platform.system())
```
karthiknadig: pytest is not imported in this block.

int19h: Oh, in fact this wouldn't work even if it's imported, since this isn't running in the test process. I'll move the check outside.

PR #1150, opened 2019-02-14
Fix #206: Validate breakpoint targets

### LINKED ISSUE #206: Validate breakpoint targets
It is currently possible to set breakpoints on lines where valid breakpoints aren't possible (e.g. blank lines, or lines in the middle of a multi-line statement). When that happens, there's no error or warning, and the breakpoint doesn't get hit.

We should either automatically move the breakpoint to the nearest valid line, or display an error message in that case.

int19h:
When setting breakpoints, validate line numbers against dis.findlinestarts() for the given file.

### DIFF
--- a/src/ptvsd/_util.py
+++ b/src/ptvsd/_util.py
@@ -8,6 +8,7 @@ import contextlib
 import os
 import threading
 import time
+import types
 import sys
 
 
@@ -413,3 +414,18 @@ def _allow_debug_break(enabled=True):
 
 def _is_debug_break_allowed():
     return _enable_debug_break
+
+
+try:
+    import dis
+except ImportError:
+    def get_code_lines(code):
+        return None
+else:
+    def get_code_lines(code):
+        for _, lineno in dis.findlinestarts(code):
+            yield lineno
+        for const in code.co_consts:
+            if isinstance(const, types.CodeType) and const.co_filename == code.co_filename:
+                for lineno in get_code_lines(const):
+                    yield lineno
--- a/src/ptvsd/_util.py
+++ b/src/ptvsd/_util.py
@@ -423,8 +423,14 @@ except ImportError:
         return None
 else:
     def get_code_lines(code):
+        # First, get all line starts for this code object. This does not include
+        # bodies of nested class and function definitions, as they have their
+        # own objects.
         for _, lineno in dis.findlinestarts(code):
             yield lineno
+
+        # For nested class and function definitions, their respective code objects
+        # are constants referenced by this object.
         for const in code.co_consts:
             if isinstance(const, types.CodeType) and const.co_filename == code.co_filename:
                 for lineno in get_code_lines(const):
--- a/src/ptvsd/wrapper.py
+++ b/src/ptvsd/wrapper.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function, absolute_import, unicode_literals
 
+import bisect
 import contextlib
 import copy
 import errno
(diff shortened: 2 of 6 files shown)


tests/func/test_breakpoints.py
```diff
@@ -398,3 +399,84 @@ def code_to_debug():
         output = session.all_occurrences_of(Event('output', ANY.dict_with({'category': 'stdout'})))
         output = sorted(int(o.body['output'].strip()) for o in output if len(o.body['output'].strip()) > 0)
         assert list(range(0, 10)) == output
+
+
+def test_invalid_breakpoints(pyfile, run_as, start_method):
+    @pyfile
+    def code_to_debug():
```
int19h: @qubitron, can you take a look at this (just the body of `code_to_debug`) to see that it captures various scenarios correctly? The intent is expressed in comments - bpN-requested is where the user might set it, and bpN-expected is where it should end up.

If you can think of any other such cases, let me know, so that I can add them here.

src/ptvsd/wrapper.py
```diff
@@ -2282,6 +2283,36 @@ def on_setBreakpoints(self, request, args):
 
         pydevd_request = copy.deepcopy(request)
         del pydevd_request['seq']  # A new seq should be created for pydevd.
+
+        # Validate breakpoints and adjust their positions.
+        args = pydevd_request['arguments']
+        if 'breakpoints' in args and 'path' in args['source']:
+            path = args['source']['path']
+            try:
+                with open(path) as f:
```
karthiknadig: For `attach` scenarios with path mappings this will not find the file.

fabioz: I think that this code should be in `_pydevd_bundle.pydevd_process_net_command_json._PyDevJsonCommandProcessor.on_setbreakpoints_request` (as @karthiknadig pointed, it requires internal `pydevd` apis to work properly).

src/ptvsd/_util.py
```diff
@@ -413,3 +414,18 @@ def _allow_debug_break(enabled=True):
 
 def _is_debug_break_allowed():
     return _enable_debug_break
+
```
fabioz: This could be moved to `_pydevd_bundle.pydevd_utils`.

src/ptvsd/_util.py
```diff
@@ -413,3 +414,18 @@ def _allow_debug_break(enabled=True):
...
+    import dis
+except ImportError:
+    def get_code_lines(code):
+        return None
+else:
+    def get_code_lines(code):
+        for _, lineno in dis.findlinestarts(code):
+            yield lineno
+        for const in code.co_consts:
+            if isinstance(const, types.CodeType) and const.co_filename == code.co_filename:
```
fabioz: It'd be nice to have a comment explaining why it's recursively checking the `CodeType.co_consts` to get lines.

Is it possible to have some code created on the fly that could recurse and have a stack overflow error here? (I think that's very improbable, but CPython itself stops the tracing and breaks the debugger when a stack overflow error is raised, so, just double-checking).

src/ptvsd/wrapper.py
karthiknadig: @fabioz I had talked about this with @int19h. Basically there are 2 parts to this change, validation and adjustment. I think validation part is well defined. The adjustment part is not yet clear about the possible set of settings we might need since it is subjective to user. So for now we wanted get something simple built and try it out. I definitely think that the validation part should be in pydevd. As for the adjustment part this may have to be done in ptvsd depending on how the settings will turnout for this. 

To keep it simple and avoid making the split when we are trying it out we decided to do it entirely here. Once we have some feedback from the internal loop we will move this into pydevd. At that point we can decide how we want to handle the settings.

src/ptvsd/_util.py
int19h: Good point (TL;DR: it's because nested definitions compile into their own code objects, which are then referenced via LOAD_CONST to execute the definition).

I don't believe recursion is possible here, since it's basically a tree, not a graph.

src/ptvsd/wrapper.py
int19h: @karthiknadig, I have updated the PR to handle path mapping.

Issue #1092, opened 2019-01-09
DAP in pydevd


fabioz:

karthiknadig:
Since we have individual work items we should track using those.

PR #1157, opened 2019-02-20
Support scopes, stackTrace, setVariable, evaluate, setExpression and modules DAP messages in pydevd. #1092

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_xml.py
```diff
@@ -152,6 +152,30 @@ def make_thread_killed_message(self, id):
...
+        while frame is not None:
+            if frame.f_code is None:
+                continue  # IronPython sometimes does not have it!
+
+            method_name = frame.f_code.co_name  # method name (if in method) or ? if global
+            if method_name is None:
+                continue  # IronPython sometimes does not have it!
+
+            abs_path_real_path_and_base = get_abs_path_real_path_and_base_from_frame(frame)
+            if py_db.get_file_type(abs_path_real_path_and_base) == py_db.PYDEV_FILE:
```
karthiknadig: This should also skip ptvsd files. is there a way to configure that?

fabioz: `ptvsd` files are already handled there (done at: https://github.com/Microsoft/ptvsd/blob/master/src/ptvsd/wrapper.py#L183)

fabioz:
Note: this pull request fixes https://github.com/Microsoft/ptvsd/issues/45

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_xml.py
karthiknadig: Ah! right.

karthiknadig:
Wait for pavels review before merging.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
```diff
@@ -1,13 +1,68 @@
...
+
+            module = Module(module_id, module_name, filename_in_utf8)
+            if version:
+                module.version = version
+
+            if package_name:
+                # Note: package doesn't appear in the docs but seems to be expected?
+                module.kwargs['package'] = package_name
+
+            module_event = ModuleEvent(ModuleEventBody('new', module))
```
int19h: Since path to module is known here, you might want to add `path` to the module event body as well.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -268,5 +272,126 @@ def on_setbreakpoints_request(self, py_db, request):
...
+    def on_source_request(self, py_db, request):
+        '''
+        :param SourceRequest request:
+        '''
+        source_reference = request.arguments.sourceReference
+        content = ''
+        if source_reference != 0:
+            server_filename = pydevd_file_utils.get_server_filename_from_source_reference(source_reference)
+            if os.path.exists(server_filename):
+                with open(server_filename, 'r') as stream:
```
int19h: This needs to handle any I/O or permission errors, and a potential race against the file being deleted right after the check above.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
```diff
@@ -212,24 +235,51 @@ def resolve(self, dict, key):
...
-        if isinstance(key, str):
-            return '%r' % key
-        else:
-            if not pydevd_constants.IS_PY3K:
-                if isinstance(key, unicode):
-                    return "u'%s'" % key
-            return key
+        if not pydevd_constants.IS_PY3K:
+            if isinstance(key, unicode):
+                return "u'%s'" % key
```
int19h: This looks like it'd have issues with quotes, backslashes, and other stuff that might need escaping.

int19h: Also, what if the key is a tuple - won't that just take the first item's repr? I think this needs to be `(key,)` to ensure it works correctly in all cases, and same for the line below.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
```diff
@@ -260,6 +310,37 @@ def resolve(self, var, attribute):
...
+        embed the reference in the key).
+
+        Note that the return should be ordered.
+
+        :return list(tuple(name:str, value:object, evaluateName:str))
+        '''
+        l = len(lst)
+        ret = []
+
+        format_str = '%0' + str(int(len(str(l)))) + 'd'
```
int19h: This looks like it needs to be `l-1` (i.e. index of the last element, rather than the length).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
```diff
@@ -260,6 +310,37 @@ def resolve(self, var, attribute):
...
+
+        i = 0
+        for item in lst:
```
int19h: enumerate()

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
int19h: Also, this is always using `%d`, but repr for keys is expected to respect hex mode as well.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -166,7 +166,10 @@ def _on_run(self):
                 self._py_db_command_thread_event.clear()
                 self._py_db_command_thread_event.wait(0.3)
         except:
-            pydev_log.debug(sys.exc_info()[0])
+            try:
+                pydev_log.debug(sys.exc_info()[0])
+            except:
+                pass  # In interpreter shutdown sys may be None already.
```
int19h: This comment can be turned into an assert to ensure that there aren't any random exceptions getting swallowed here unexpectedly.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
```diff
@@ -212,24 +235,51 @@ def resolve(self, dict, key):
...
+        i = 0
+        for key, val in dict_iter_items(dct):
+            i += 1
+            key_as_str = self.key_to_str(key)
+            ret.append((key_as_str, val, '[%s]' % (key_as_str,)))
+            if i > MAX_ITEMS_TO_HANDLE:
+                ret.append((TOO_LARGE_ATTR, TOO_LARGE_MSG))
+                break
+
+        ret.append(('__len__', len(dct), '.__len__()'))
```
int19h: There's a disparity here in that the value it reports is `len()`, but it's presented as the result of calling `__len__()`. These can be different in some corner cases, since `len()` does a bunch of type conversions on what `__len__` returns.

In a similar case elsewhere, we used synthetic members that are presented as function calls - i.e. instead of `d.__len__(x)`, it would show up as a member `len(d)` in the IDE, and marked differently from attributes. Might be something to consider here.

tests/func/test_set_expression.py
```diff
@@ -0,0 +1,62 @@
...
+            'frameId': hit.frame_id
+        }).wait_for_response()
+        scopes = resp_scopes.body['scopes']
+        assert len(scopes) > 0
+
+        resp_variables = session.send_request('variables', arguments={
+            'variablesReference': scopes[0]['variablesReference']
+        }).wait_for_response()
+        variables = list(v for v in resp_variables.body['variables'] if v['name'] == 'a')
+        assert len(variables) == 1
```
int19h: You can combine these two in a single assert by asserting directly against the list, i.e. `assert variables = [{...}]`.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
fabioz: I've created https://github.com/Microsoft/ptvsd/issues/1172 to address this (trying to keep the features in this pull request as close as possible to what was already available).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
fabioz: Agreed. This should be properly tested too to return proper errors on those cases, so, created https://github.com/Microsoft/ptvsd/issues/1173 to address this.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
fabioz: Agreed (fixed -- always using `'%r' % (key,)`).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
fabioz: Fixed the `l-1` (with proper tests).

Created https://github.com/Microsoft/ptvsd/issues/1174 to deal with the hex formatting.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
fabioz: Done.

src/ptvsd/_vendored/pydevd/pydevd.py
fabioz: Rectified the comment to:
```
# In interpreter shutdown many things can go wrong (any module variables may
# be None, streams can be closed, etc).
```

It's not really just `sys` which may be `None`, many other things could go haywire at that point. 

Note that I agree this should be handled better, but we should first have a way to identify that we're indeed in interpreter shutdown (created https://github.com/Microsoft/ptvsd/issues/1175 to address that as it's really out of the scope of this pull request -- this was mostly because of a test failure at interpreter shutdown).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
fabioz: Agreed. Created https://github.com/Microsoft/ptvsd/issues/1176 to address that (ptvsd was giving `d.__len__`, before, so, it shouldn't be a blocker -- and I think the changes needed for that will be clearer in a new pull request).

tests/func/test_set_expression.py
fabioz: Done

fabioz:
Ok, fixed some of the issues from the review and created other items (which I plan to tackle shortly in separate pull requests as this one is already pretty big) for other items (@karthiknadig please let me know if something with more priority comes along).

Also rebased with master, so, if it passes I plan on merging tomorrow.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
fabioz: Actually, `filename_in_utf8` is passed as the path already, so, closing #1172.

Issue #1159, opened 2019-02-20
stopOnEntry not stopping on user code


karthiknadig:
## Environment data

- PTVSD version: master
- OS and version: any
- Python version (& distribution if applicable, e.g. Anaconda): 3.7
- Using VS Code or Visual Studio: VSC

## Actual behavior

Debugger pauses, but does not show where it is broken.

## Expected  behavior

Should break on first line of user code. And show that it is broken there.

## Logs
```js
{"type": "event", "seq": 10, "event": "stopped", "body": {"reason": "step", "threadId": 1, "text": null, "description": null, "preserveFocusHint": false, "allThreadsStopped": true}}

{"command":"threads","type":"request","seq":8}


{"type": "response", "seq": 11, "request_seq": 8, "success": true, "command": "threads", "message": "", "body": {"threads": [{"id": 1, "name": "MainThread"}]}}


{"command":"stackTrace","arguments":{"threadId":1,"startFrame":0,"levels":20},"type":"request","seq":9}


{"type": "response", "seq": 12, "request_seq": 9, "success": true, "command": "stackTrace", "message": "", "body": {"stackFrames": [], "totalFrames": 0}}
```

As you can see there are no frames returned.

fabioz:
I fixed an issue where it was not properly identifying if a file was from ptvsd or not, but in theory it'd only affect windows as it was comparing a normalized path to a non-normalized path... this report says any OS, can you confirm that? (if that's the case there may be some other issue -- I added a test for this in ptvsd now and it seems to be passing in other OSes).

PR #1160, opened 2019-02-21
Fixes issue identifying ptvsd files (fixes stop on entry). #1159

fabioz:


fabioz:
Errors from the Azure ci don't seem related to this change.

Issue #923, opened 2018-10-16
AttributeError: 'thread' has no attribute 'frame'


huguesv:
## Environment data

- PTVSD version: 4.1.3, also tested master branch
- OS and version: Windows
- Python version (& distribution if applicable, e.g. Anaconda): Python 2.7, django 1.11.6
- Using VS Code or Visual Studio: VS

## Actual behavior

Does stop break on template breakpoint
Console output:
```
Traceback (most recent call last):
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_comm.py", line 801, in make_thread_stack_str
    abs_path_real_path_and_base = get_abs_path_real_path_and_base_from_frame(curr_frame)
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\_vendored\pydevd\pydevd_file_utils.py", line 546, in get_abs_path_real_path_and_base_from_frame
    ret = get_abs_path_real_path_and_base_from_file(f)
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\_vendored\pydevd\pydevd_file_utils.py", line 524, in get_abs_path_real_path_and_base_from_file
    abs_path, real_path = _NormPaths(f)
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\_vendored\pydevd\pydevd_file_utils.py", line 229, in _NormPaths
    raise AssertionError('Paths passed to _NormPaths must be str. Found: %s (%s)' % (filename, type(filename)))
AssertionError: Paths passed to _NormPaths must be str. Found: c:\users\huvalo\source\repos\djangowebproject10\djangowebproject10\app\templates\app\contact.html (<type 'unicode'>)
Traceback (most recent call last):
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\wrapper.py", line 1327, in done
    fut.result()
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\futures.py", line 40, in result
    reraise(self._exc_info)
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\futures.py", line 149, in callback
    x = next(it)
  File "C:\Projects\GitHub\microsoft\ptvsd\ptvsd\wrapper.py", line 2472, in on_pydevd_thread_suspend
    xframes = list(xml.thread.frame)
[...]

PR #1179, opened 2019-02-26
[WIP] Ensure template exceptions are not ignored

### LINKED ISSUE #694: Debugger does not break on template exceptions
## Environment data

- PTVSD version: master
- OS and version: windows
- Python version (& distribution if applicable, e.g. Anaconda): 3.7
- Packages: Django

## Actual behavior

Does not break on template exceptions when 'raised' is selected. Trace back is written to the console.

## Expected  behavior

Should break on template exceptions

## Steps to reproduce:

1. Create a minimal django app.
2. Use following template:
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8" />
        <title>Django test</title>
    </head>
    <body>
        {% doesnotexist %}
    </body>
</html>
```

karthiknadig:
Fixes #694

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
@@ -128,7 +128,9 @@ class PyDBFrame:
                             return False, frame
 
                     if exception_breakpoint.ignore_libraries:
-                        if not main_debugger.is_exception_trace_in_project_scope(trace):
+                        if not main_debugger.is_exception_trace_in_project_scope(trace) and \
+                            main_debugger.plugin is not None and \
+                            not main_debugger.plugin.has_exception_breaks():
                             pydev_log.debug("Ignore exception %s in library %s -- (%s)" % (exception, frame.f_code.co_filename, frame.f_code.co_name))
                             return False, frame
 
--- a/src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
@@ -229,15 +229,14 @@ def _NormPaths(filename):
     try:
         return NORM_PATHS_CONTAINER[filename]
     except KeyError:
-        if filename.__class__ == str or (sys.version_info < (3, 0) and filename.__class__ == unicode):
-            abs_path = _NormPath(filename, os.path.abspath)
-            real_path = _NormPath(filename, rPath)
-
-            # cache it for fast access later
-            NORM_PATHS_CONTAINER[filename] = abs_path, real_path
-            return abs_path, real_path
-        else:
+        if filename.__class__ != str:
             raise AssertionError('Paths passed to _NormPaths must be str. Found: %s (%s)' % (filename, type(filename)))
+        abs_path = _NormPath(filename, os.path.abspath)
+        real_path = _NormPath(filename, rPath)
+
+        # cache it for fast access later
+        NORM_PATHS_CONTAINER[filename] = abs_path, real_path
+        return abs_path, real_path
 
 
 def _NormPath(filename, normpath):
--- a/src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
@@ -229,14 +229,15 @@ def _NormPaths(filename):
     try:
         return NORM_PATHS_CONTAINER[filename]
     except KeyError:
-        if filename.__class__ != str:
-            raise AssertionError('Paths passed to _NormPaths must be str. Found: %s (%s)' % (filename, type(filename)))
-        abs_path = _NormPath(filename, os.path.abspath)
-        real_path = _NormPath(filename, rPath)
+        if filename.__class__ == str or (sys.version_info < (3, 0) and filename.__class__ == unicode):
+            abs_path = _NormPath(filename, os.path.abspath)
+            real_path = _NormPath(filename, rPath)
 
-        # cache it for fast access later
-        NORM_PATHS_CONTAINER[filename] = abs_path, real_path
-        return abs_path, real_path
+            # cache it for fast access later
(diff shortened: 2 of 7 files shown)


src/ptvsd/wrapper.py
```diff
@@ -541,11 +541,15 @@ def __init__(self, proc):
         self.exceptions = {}
         self.lock = threading.Lock()
 
+    def remove_exception_break(self, ex_type=None, exception='BaseException'):
```
int19h: Might as well just make it `ex_type='python'` and skip the check inside, since strings are immutable.

tests/func/test_django.py
```diff
@@ -105,7 +105,7 @@ def test_django_breakpoint_no_multiproc(bp_target, start_method):
 
 
 @pytest.mark.parametrize('start_method', ['launch', 'attach_socket_cmdline'])
-@pytest.mark.skip(reason='Bug #694')
+@pytest.mark.skipif(sys.version_info < (3, 0), reason='Bug #923')
```
int19h: #923 is closed no repro - should it be reopened?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
```diff
@@ -128,7 +128,9 @@ def should_stop_on_exception(self, frame, event, arg):
                             return False, frame
 
                     if exception_breakpoint.ignore_libraries:
-                        if not main_debugger.is_exception_trace_in_project_scope(trace):
+                        if not main_debugger.is_exception_trace_in_project_scope(trace) and \
+                            main_debugger.plugin is not None and \
```
fabioz: I think this would fail in the following case:
- add django exception breakpoint
- mark to ignore exceptions in libraries
- non-django exception is raised
- exception would be shown to the user

The current problem is the following: if there's an exception breakpoint it always takes priority over plugin breakpoints (i.e.: `if exception_breakpoint is not None...`). 

This is backwards, I think that the proper fix is checking if there are exception breakpoints and if there are, ask the plugins to handle it (i.e.: the `main_debugger.plugin.exception_break` currently in the `else` block) and if it's not handled by the plugins then go on to the current code.

So, we should have a test-case for this case (not sure if there is already) and the fix is something as:

```
# Plugins should have priority when handling exceptions so that an exception is not
# ignored if a plugin would handle it.
should_stop = False
if main_debugger.plugin is not None and main_debugger.plugin.has_exception_breaks():
    result = main_debugger.plugin.exception_break(main_debugger, self, frame, self._args, arg)
    if result:
        should_stop, frame = result

if not should_stop:
    exception_breakpoint = main_debugger.get_exception_breakpoint(
        exception, main_debugger.break_on_caught_exceptions)
    if exception_breakpoint is not None:
        ... current code to handle exception_breakpoint ...
```

tests/func/test_django.py
fabioz: This doesn't seem right to me... that exception `AttributeError: 'thread' has no attribute 'frame'` means a real problem (it means that the debugger is stopping in a file inside the debugger -- this shouldn't happen because we should not trace those files at all).

Note that there was a recent fix for this where `ptvsd` files could end up being traced (https://github.com/Microsoft/ptvsd/commit/fd2aea41f3619b2a4e263f3151db39b1a5d29277), but if that's not the case and you have a repro, please reopen the issue and assign it to me as I think this would be a critical bug.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
karthiknadig: I tried this, it does not seem to work right. I see another problem. Template exceptions in django are not reporting in the django exception break context. Also the first frame is app.py in the test case. So there is something broken there.

Also, since it is not the first frame, just_reported is false. so django excpetion_break always returns None.

fabioz: So regular django-exception breakpoints aren't working properly? If something is broken in `pydevd_plugins.django_debug.exception_break` it should probably be fixed first then (do you want me to take a look at it?).

karthiknadig: Yes, please take a look at this. I will close this PR. I will reopen after the django part is fixed.

tests/func/test_django.py
karthiknadig: I wanted to turn it off on 2.7, missed updating the reason. Since it is now closed, i removed that from some of the other tests as well.

src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
```diff
@@ -229,14 +229,15 @@ def _NormPaths(filename):
     try:
         return NORM_PATHS_CONTAINER[filename]
     except KeyError:
-        if filename.__class__ != str:
+        if filename.__class__ == str or (sys.version_info < (3, 0) and filename.__class__ == unicode):
```
karthiknadig: @fabioz is there a better way to do 2.7 compat here?

fabioz: Humm, how did you get to this use case (in general, this would be an error -- it should really be str on both -- i.e.: bytes on py2 and str on py3)

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py
fabioz: Ok, created https://github.com/Microsoft/ptvsd/issues/1181 for that.

src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
karthiknadig: running `test_flask_template_exception_no_multiproc` test on windows  resulted in that assertion.

karthiknadig:
Closing this PR, I will create a separate PR with updated tests. The pydevd parts that need to be fixed have a separate task.

PR #1182, opened 2019-02-26
Enable delayed stack trace loading

### LINKED ISSUE #1146: Add support for Delayed Stack Trace Loading
Enable `supportsDelayedStackTraceLoading` and ensure we support `startFrame` and `levels` arguments of the `StackTraceRequest`.

karthiknadig:
Fixes #1146

### DIFF
--- a/src/ptvsd/wrapper.py
+++ b/src/ptvsd/wrapper.py
@@ -958,6 +958,7 @@ INITIALIZE_RESPONSE = dict(
     supportsConditionalBreakpoints=True,
     supportsConfigurationDoneRequest=True,
     supportsDebuggerProperties=True,
+    supportsDelayedStackTraceLoading=True,
     supportsEvaluateForHovers=True,
     supportsExceptionInfoRequest=True,
     supportsExceptionOptions=True,
@@ -1562,7 +1563,6 @@ class VSCodeMessageProcessor(VSCLifecycleMsgProcessor):
 
     @async_handler
     def on_stackTrace(self, request, args):
-        # TODO: docstring
         vsc_tid = int(args['threadId'])
 
         try:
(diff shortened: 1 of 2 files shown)


src/ptvsd/wrapper.py
```diff
@@ -1582,7 +1582,16 @@ def on_stackTrace(self, request, args):
...
+        levels = int(args.get('levels', 0))
+        if levels == 0:
+            stackFrames = resp_args['body']['stackFrames']
+        else:
+            try:
+                start = int(args.get('startFrame', 0))
+                end = start + levels
+                stackFrames = resp_args['body']['stackFrames'][start:end]
+            except:
+                stackFrames = []
```
int19h: The DAP spec says that levels is the "maximum number of frames to return", so it's an upper bound, not a request for a specific number - but this will return an empty list if `end` is out of bounds. OTOH it will return an empty list instead of reporting an error on invalid `start`, because of the catch-all except.

 I think this needs to be more permissive on `end`, and just allow exception to propagate on invalid `start`; something like:
```py
        stackFrames = resp_args['body']['stackFrames']
        if levels > 0:
                start = int(args.get('startFrame', 0))
                end = min(start + levels, len(stackFrames))
                stackFrames = stackFrames[start:end]
```

tests/func/test_breakpoints.py
```diff
@@ -480,3 +480,54 @@ def code_to_debug():
...
+                'threadId': hit.thread_id,
+                'startFrame': start,
+                'levels': 25
+            }).wait_for_response()
+            assert resp_stacktrace.body['totalFrames'] > 0
+            frames += resp_stacktrace.body['stackFrames']
+            start = len(frames)
+
+        assert len(full_frames) == len(frames)
+        assert full_frames == frames
```
int19h: The second assert implies the first.

karthiknadig: I had that on for debugging purposes. Will remove.

PR #1206, opened 2019-03-06
Fix another logging issue.

int19h:


src/ptvsd/log.py
```diff
@@ -118,8 +118,7 @@ def g(*args, **kwargs):
 def to_file():
     global file
 
-    if ptvsd.options.log_dir:
-        assert not file
+    if ptvsd.options.log_dir and not file:
```
int19h: This function can get invoked twice on some code paths, by design, so the assert should rather be a simple check.

Issue #1202, opened 2019-03-06
Error seen on disconnect from remote machine


karthiknadig:
## Environment data

- PTVSD version: master
- OS and version: Windows local, Linux remote
- Python version (& distribution if applicable, e.g. Anaconda): Conda
- Using VS Code or Visual Studio: VSC

## Actual behavior

```
Traceback (most recent call last):
              File "/azureml-envs/azureml_/lib/python3.6/site-packages/ptvsd/ipcjson.py", line 279, in process_one_message
                msg = self.__message.pop(0)
            IndexError: pop from empty list
            
            During handling of the above exception, another exception occurred:
            
            Traceback (most recent call last):
              File "/azureml-envs/azureml_/lib/python3.6/site-packages/ptvsd/ipcjson.py", line 284, in process_one_message
                msg = self.__message.pop(0)
            IndexError: pop from empty list
            
            During handling of the above exception, another exception occurred:
            
            Traceback (most recent call last):
              File "/azureml-envs/azureml_/lib/python3.6/site-packages/ptvsd/ipcjson.py", line 268, in process_messages
                self.process_one_message()
              File "/azureml-envs/azureml_/lib/python3.6/site-packages/ptvsd/ipcjson.py", line 290, in process_one_message
                raise EOFError('no more messages')
            EOFError: no more messages
```

int19h:
Should be fixed by 4d4a572ca291aa9ae6f4e3c85440b8700ab52199

PR #1210, opened 2019-03-07
Assorted test fixes

int19h:


src/ptvsd/ipcjson.py
```diff
@@ -266,7 +266,7 @@ def process_messages(self):
         while not self.__exit:
             try:
                 self.process_one_message()
-            except AssertionError:
+            except (AssertionError, EOFError):
```
int19h: I think this will also fix https://github.com/Microsoft/ptvsd/issues/1202 - basically the exception was expected, the problem is that it was incorrectly also logged as error below.

tests/conftest.py
```diff
@@ -111,10 +111,12 @@ def factory(source):
...
 
         # Now we need to adjust indentation. Compute how much the first line of
-        # the body is indented by, then dedent all lines by that amount.
+        # the body is indented by, then dedent all lines by that amount. Blank
+        # lines don't matter indentation-wise, and might not be indented to begin
+        # with, so just replace them with a simple newline.
         line = source[0]
         indent = len(line) - len(line.lstrip())
-        source = [line[indent:] for line in source]
+        source = [line[indent:] if line.strip() else '\n' for line in source]
```
int19h: Without this change, we were basically stripping out blank lines from `@pyfile` code, which made line numbers hard to predict. It doesn't matter as much when using `#@` markers, but it can still be confusing in the logs.

PR #1235, opened 2019-03-13
Fixes #1212. Adds json requests and tests for pause, continue, and stepping

### LINKED ISSUE #1212: Migrate pause, continue, next, stepIn, and stepOut

karthiknadig:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -184,6 +186,85 @@ def on_attach_request(self, py_db, request):
...
+        return NetCommand(CMD_RETURN, 0, response.to_dict(), is_json=True)
+
+    def on_next_request(self, py_db, request):
+        '''
+        :param NextRequest request:
+        '''
+        arguments = request.arguments  # : :type arguments: NextArguments
+        thread_id = arguments.threadId
+
+        if thread_id.startswith('*'):
```
int19h: This is going to break once we switch thread IDs to numbers, as the spec requires. Perhaps make it a separate message for step-over-my-code? We could also add a custom field to the request - I think that's allowed.

karthiknadig: I agree. @fabioz, do you have any preference here?

fabioz: I think `pydevd` should respect the `just-my-code` global setting and receive just a regular `next` request with the thread id without the `*` prefix (the settings are already set in `_set_debug_options`... it sets the value at `self.api.set_use_libraries_filter`, so, a `self.api.get_use_libraries_filter` should be done and that setting should be used to decide which internal command should be used).

PR #1244, opened 2019-03-17
Fix #1227: subprocess arguments incorrectly escaped inside the debugger

### LINKED ISSUE #1227: subprocess arguments incorrectly escaped inside the debugger
@Meinersbur commented on [Sun Mar 10 2019](https://github.com/Microsoft/PTVS/issues/5155)

Consider the following program:
```
import sys
import subprocess

python = sys.executable
subprocess.check_call([python, '-c', 'print("A B")'])
subprocess.check_call([python, '-c', 'print("XY")'])
print("Done")
```

When executed with python from the command line this prints:
```
>python3 module1.py
A B
XY
Done

>python3 --version
Python 3.7.2
```

When executed from inside the Visual Studio 2017 debugger (same python 3.7.2, Python installed product 15.9.18254.1 according to Help -> About...), and Visual Studio 2019 preview, I get:
```
A B
NameError: name 'XY' is not defined
```

The issue seems that the command line argument is only escaped when there is a space in it, but not deemed necessary with quotes. I checked the subprocess module, it correctly calls CreateProcess with escaped arguments in both cases. That is, the python debugger seems to hook the CreateProcess call, parses the command line, but re-escapes it insufficiently.

int19h:
Re-implement argument quoting on Windows in accordance with
https://docs.microsoft.com/en-us/windows/desktop/api/shellapi/nf-shellapi-commandlinetoargvw

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
+++ b/src/ptvsd/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
@@ -1,5 +1,6 @@
 # License: EPL
 import os
+import re
 import sys
 import traceback
 from _pydev_imps._pydev_saved_modules import threading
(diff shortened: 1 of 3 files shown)


int19h:
There's some genuine test issue, I'll look into it.

Issue #1232, opened 2019-03-13
Can't initiate remote debugging from an IPython process


alendit:
## Environment data

- PTVSD version: 4.2.4
- OS and version: arch (latest)
- Python version: 3.7.2 (GCC 7.3.0)
- VS Code
- IPython: 7.2.0

## Actual behavior

When trying to connect to a remote jupyter process, the following happens:

```
import ptvsd
ptvsd.enable_attach(address=('127.0.0.1', 3000), redirect_output=True)
ptvsd.wait_for_attach()  # blocks
# attach to this address from a parallelly running vscode instance
# this works fine, call stack window displays the currently running threads
# the ipython execution continues
print("Test")
# stdout is displayed in the debugging console correctly
def test():
    v = 10
    breakpoint()  # alternatively ptvsd.break_into_debugger()
    return v
test()
# the function returns 10 without breaking
# ipython returns to interactive mode
```

## Expected  behavior

Last step should pause on the breakpoint.

Doing the same steps in a separate script produces the expected behaviour.

I guess it has something to do with jupyter threading / interactivity implementation, but I couldn't find what's going wrong from a cursory looks through `pydebd` code. Other debuggers, e.g. `ipdb`, `pudb`, work fine in the same setting.

int19h:
The problem is that we consider any code that has filename like `<...>` to be filtered out as not-user-code for stepping purposes - and `breakpoint()` is considered a kind of step - unless it's `<string>`:

https://github.com/Microsoft/ptvsd/blob/1173a2b913e5bb0ec628038c0b71429cfd0c88d2/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_filtering.py#L225

In case of IPython cells, they have names like `<ipython-input-2-1f2f30391028>`, so they get excluded in their entirety. 

As a workaround, this check can be disabled by setting `'debugStdLib': true` in your attach configuration in launch.json, and then the break works normally. Note that you won't see the source code of the cell in debugger when it breaks for the same reason (it needs a filename to do that). But you can still see the locals and the stack, and do blind stepping.

Still, this shouldn't be treated as non-user-code in the first place. @fabioz, @karthiknadig, what do you guys think? We can add another special case there for `<ipython...>` , similar to how `<string>` is handled, but perhaps it's better to treat `<...>` as user code by default, and only exclude known things like `<frozen...>`?

fabioz:
I think it should be ok to  do that...  with some additional work, it
[...]

PR #1245, opened 2019-03-17
Fix #1173: Gracefully handle I/O errors when processing source request 

### LINKED ISSUE #1173: Gracefully handle I/O errors when processing source request
_pydevd_bundle.pydevd_process_net_command_json._PyDevJsonCommandProcessor.on_source_request should protect against permission errors and a race condition where file is erased after checking that it exists.

int19h:
Also adds `linecache` fallback and a different approach to handling `<...>` as discussed in #1232.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
@@ -1,6 +1,7 @@
 from functools import partial
 import itertools
 import json
+import linecache
 import os
 
 from _pydevd_bundle._debug_adapter import pydevd_base_schema
@@ -460,15 +461,41 @@ class _PyDevJsonCommandProcessor(object):
         :param SourceRequest request:
         '''
         source_reference = request.arguments.sourceReference
-        content = ''
+        server_filename = None
+        content = None
+
         if source_reference != 0:
             server_filename = pydevd_file_utils.get_server_filename_from_source_reference(source_reference)
-            if os.path.exists(server_filename):
-                with open(server_filename, 'r') as stream:
-                    content = stream.read()
+            if server_filename:
+                # Try direct file access first - it's much faster when available.
+                try:
+                    with open(server_filename, 'r') as stream:
+                        content = stream.read()
+                except:
+                    pass
 
-        body = SourceResponseBody(content)
-        response = pydevd_base_schema.build_response(request, kwargs={'body':body})
+                if content is None:
+                    # File might not exist at all, or we might not have a permission to read it,
+                    # but it might also be inside a zipfile, or an IPython cell. In this case,
+                    # linecache might still be able to retrieve the source.
+                    lines = (linecache.getline(server_filename, i) for i in itertools.count(1))
+                    lines = itertools.takewhile(bool, lines)  # empty lines are '\n', EOF is ''
+
+                    # If we didn't get at least one line back, reset it to None so that it's
+                    # reported as error below, and not as an empty file.
+                    content = ''.join(lines) or None
+
+        body = SourceResponseBody(content or '')
+        response_args = {'body': body}
+
+        if content is None:
+            if server_filename:
+                message = 'Unable to retrieve source for %s' % (server_filename,)
+            else:
+                message = 'Invalid sourceReference %d' % (source_reference,)
+            response_args.update({'success': False, 'message': message})
+
+        response = pydevd_base_schema.build_response(request, kwargs=response_args)
         return NetCommand(CMD_RETURN, 0, response.to_dict(), is_json=True)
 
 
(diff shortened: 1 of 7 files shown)


int19h:
And yes, with `linecache` and normalization fixes, IPython cell source fetching works as expected:

image

(Note that there are still many other things not working well or at all when debugging cells, so this doesn't make it a properly supported scenario, just improves the current state of affairs slightly.)

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -460,15 +461,41 @@ def on_source_request(self, py_db, request):
...
+                    with open(server_filename, 'r') as stream:
+                        content = stream.read()
+                except:
+                    pass
+
+                if content is None:
+                    # File might not exist at all, or we might not have a permission to read it,
+                    # but it might also be inside a zipfile, or an IPython cell. In this case,
+                    # linecache might still be able to retrieve the source.
+                    lines = (linecache.getline(server_filename, i) for i in itertools.count(1))
```
fabioz: I think you could use `linecache.getlines` to get the whole file at once instead of getting it line-by-line.

src/ptvsd/wrapper.py
```diff
@@ -1557,8 +1557,10 @@ def on_source(self, request, args):
                 pydevd_request,
                 is_json=True)
 
-            body = resp_args['body']
-            self.send_response(request, **body)
+            if resp_args.get('success', True):
```
fabioz: Since `pydevd` now appears to send a proper error response, why is this check still needed?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -460,15 +461,41 @@ def on_source_request(self, py_db, request):
...
-        content = ''
+        server_filename = None
+        content = None
+
         if source_reference != 0:
             server_filename = pydevd_file_utils.get_server_filename_from_source_reference(source_reference)
-            if os.path.exists(server_filename):
-                with open(server_filename, 'r') as stream:
-                    content = stream.read()
+            if server_filename:
```
fabioz: I think we really should have tests for these cases (error on `open` and using `linecache`).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
int19h: It's not a part of its public API - at least I don't see it there:
https://docs.python.org/3/library/linecache.html

src/ptvsd/wrapper.py
int19h: It's necessary to properly propagate it - `send_response` always sends a success response, and old code just propagated the body. Which in error case was just blank, so on the IDE side, you'd get an empty file opened instead of an error.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
fabioz: Humm, I see it's not in the docs, but it's been there since the first revision up to the latest master (so, I don't think they'd break it).

Anyways not sure... I think I'd go with the `linecache.getlines` -- even not being in the docs as it should be faster and have less issues with threads (as theoretically the cache could be invalidated in a thread while we're reading it) , but I'll leave that up to you whether you think that's better or not.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
int19h: Testing `linecache` proved to be tricky - I tried a few approaches, but eventually had to resort to just adding entries directly to the cache. See the new commit.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
int19h: In practice in most cases we'd just go straight for the disk file anyway, so perf is unlikely to be an issue. For IPython, it hacks the cache so that it's not really a cache anymore (i.e. the entries for cells never expire, and are effectively immutable so long as the cell is alive).

fabioz:
It seems that those changes broke the `test_in_project_roots` test in pydevd (see: https://travis-ci.org/fabioz/PyDev.Debugger/builds/508939501)... can you fix it before merging?

int19h:
Here we go. I changed the paths, since now there shouldn't be names like `/tmp/<string>` floating around anymore, as they are not force-normalized.

PR #1257, opened 2019-03-21
Make sure that DAP ids are int32. Fixes #1251

### LINKED ISSUE #1251: VS error: value was either too large or too small for an Int32
ptvsd master
VS 2017
Python 3.7 64-bit

Code as simple as:
```py
print(3)
```

Set breakpoint and F5:

image

Seems to be a 64-bit specific thing, doesn't repro on 32-bit. 

I suspect that this is actually a VS DAP client bug, where some `number` in the protocol is improperly treated as `int32`, but the actual value is out of range for the latter.

fabioz:
The approach I decided to take is to add the translation support to the schema so that when serializing/deserializing messages the translation is done automatically (the particular places are `NetCommand.__init__` and `_PyDevJsonCommandProcessor.process_net_command_json`).

This was a bit more work than manually doing it at the needed places but in the future when we translate the `threadId` it should use the same recipe.

### DIFF
--- a/tests/helpers/pattern.py
+++ b/tests/helpers/pattern.py
@@ -12,6 +12,7 @@ from tests.helpers.pathutils import compare_path
 
 
 class BasePattern(object):
+
     def __repr__(self):
         raise NotImplementedError()
 
@@ -48,6 +49,7 @@ class Any(BasePattern):
         """
 
         class AnyDictWith(dict):
+
             def __repr__(self):
                 return repr(items)[:-1] + ', ...}'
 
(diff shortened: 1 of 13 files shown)


tests/helpers/pattern.py
```diff
@@ -165,3 +167,9 @@ def __ne__(self, other):
 
 ANY.int = ANY.such_that(lambda x: isinstance(x, numbers.Integral))
 ANY.int.name = 'ANY.int'
+
+# Note: in practice it could be any int32, but as in those cases we expect the number to be
+# incremented sequentially, this should be reasonable for tests.
+ANY.small_int = ANY.such_that(lambda x: isinstance(x, numbers.Integral) and x < 10000)
```
int19h: Since it's signed, this should check the negative bound as well, just in case.

Also, since it's for IDs specifically, maybe make it `ANY.dap_id`?

PR #1260, opened 2019-03-22
Add json support for exception info request. Fixes #1222

### LINKED ISSUE #1222: pydevd json support for exceptionInfo

karthiknadig:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -1076,6 +1076,99 @@ def internal_get_description(dbg, seq, thread_id, frame_id, expression):
         dbg.writer.add_command(cmd)
 
 
+def _get_line_for_traceback(file_path, line_no):
```
int19h: `linecache.getline` is what traceback itself uses for this exact purpose; I think we should do so as well.

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -982,6 +983,33 @@ def test_evaluate(case_setup):
...
+        writer.write_set_protocol('http_json')
+        writer.write_add_exception_breakpoint_with_policy(
+            'IndexError',
+            notify_on_handled_exceptions=2,  # Notify only once
+            notify_on_unhandled_exceptions=0,
+            ignore_libraries=1
+        )
+
+        json_facade.write_make_initial_run()
+        hit = writer.wait_for_breakpoint_hit(REASON_CAUGHT_EXCEPTION)
```
karthiknadig: @fabioz I am having issues with getting this test to hit here. The `ptvsd` test seems to be running fine. But i am not sure why this one does not get hit. do you see anything obvious that i might be missing?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -1076,6 +1077,96 @@ def internal_get_description(dbg, seq, thread_id, frame_id, expression):
...
+
+                info = dbg.suspended_frames_manager.get_topmost_frame_and_frame_id_to_line(thread_id)
+                if info is not None:
+                    topmost_frame, frame_id_to_lineno = info
+
+                if trace_obj is not None:
+                    for frame_id, frame, method_name, filename_in_utf8, lineno in iter_visible_frames_info(
+                            dbg, trace_obj.tb_frame, frame_id_to_lineno):
+ 
+                        if sys.version_info < (3, 5,):
```
fabioz: Why get the line text only when < 3.5?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -126,6 +126,15 @@ def request_stack(self, py_db, seq, thread_id, fmt=None, timeout=.5):
         else:
             py_db.post_internal_command(internal_get_thread_stack, '*')
 
+    def request_exception_info(self, py_db, request, thread_id):
```
fabioz: This should be `request_exception_info_json` as it only handles the json case.

i.e.: The cases where the postfix is not needed are the ones where the message is received by the proper frontend (xml or json) then passed on through to the api (the same call for both xml and json) and the final dispatch is handled by the current `py_db.cmd_factory` (which is either `NetCommandFactory` or `NetCommandFactoryJson`). On the cases where there's very little reuse of code, covering only one of those is fine, but the method should be postfixed accordingly.

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
fabioz: Not sure... I applied your changes here and it seems to break for me (maybe you have outdated `.pyds` locally?).

Repo and results from running those tests:

https://github.com/fabioz/PyDev.Debugger/tree/karthiknadig-issue1222
https://travis-ci.org/fabioz/PyDev.Debugger/builds/510299196

-- note that I only did an additional fix to the test after the break was hit: https://github.com/fabioz/PyDev.Debugger/commit/5c4bb7ac0d1a1c58e09ad61d5b399e3d5a59f204

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
karthiknadig: `traceback.format_list` for `>=3.5` gets the source is line_text is `None`, we don't really need to provide that. Since we are using `linecache`, this check is probably not needed.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -1076,6 +1077,93 @@ def internal_get_description(dbg, seq, thread_id, frame_id, expression):
...
+            )
+        )
+    except:
+        exc = get_exception_traceback_str()
+        response = pydevd_base_schema.build_response(request, kwargs={
+            'success': False,
+            'message': exc,
+            'body':{}
+        })
+    dbg.writer.add_command(NetCommand(CMD_RETURN, 0, response.to_dict(), is_json=True))
```
fabioz: Note: with the latest changes I just merged, this should pass just `response` and not `response.to_dict()`.

PR #1270, opened 2019-03-27
 Use json for goto and gotoTargets request. Fixes #1221

### LINKED ISSUE #1221: pydevd json support for gotoTargets and goto

karthiknadig:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -85,6 +86,8 @@ def __init__(self, from_json):
         self.api = PyDevdAPI()
         self._debug_options = {}
         self._next_breakpoint_id = partial(next, itertools.count(0))
+        self._next_goto_targets_id = partial(next, itertools.count(0))
+        self._goto_targets_map = {}
```
int19h: It looks like we never clean this up, and it's essentially unbounded, because every time a request is done there's a new ID, even if it's a line that has been queried before.

Unfortunately, the DAP spec doesn't define the lifetime for goto targets. We could assume that a `goto` request invalidates all previously obtained `gotoTargets` (whether they were or weren't used in it), but that would break any IDE that decides to cache it to avoid roundtrips.

I think we should file another request for spec clarification. But for the time being, let's play it safe, and assume no invalidation. In this case, we can at least put an upper bound on it by ensuring that the target for the same line always has the same ID - this is what the old code in wrapper.py was doing.

fabioz: I think there could be a helper (like the `IDMap` in `ptvsd`) which allows auto-generating or getting the same id (although I think it needs less functionality `IDMap` has... probably just something as `obtain_id(value)` and `obtain_value(id)`).

PR #1280, opened 2019-03-29
Debugger should not stop on exceptions raised in excluded files. Fixes #1268

### LINKED ISSUE #1268: Debugger should not stop on exceptions raised in excluded files.
See: https://github.com/Microsoft/ptvsd/issues/997#issuecomment-474986271 for more info.

fabioz:


tests/helpers/session.py
```diff
@@ -370,7 +371,8 @@ def wait_for_termination(self):
         self.wait_for_disconnect(close=False)
 
         if Event('exited') in self:
-            self.expect_realized(Event('exited', {'exitCode': self.expected_returncode}))
+            if self.expected_returncode != 'any':
+                self.expect_realized(Event('exited', {'exitCode': self.expected_returncode}))
```
int19h: I would expect it to just work if `expected_returncode` is set to `pattern.ANY` - this shouldn't need any special logic to handle it. Ditto below.

fabioz: Done

PR #1292, opened 2019-04-02
fix incorrect exception raised while processing a list that exceeds M…

baek9:
fix incorrect exception raised while processing a list that exceeds MAX_ITEMS_TO_HANDLE

Issue number is : #1301 

In VS Code, when expanding a list with more than MAX_ITEMS_TO_HANDLE(default is 300) items in the debugging window, an exception(ValueError) occurs and debugging stops.


baek9:
It is in the middle of writing the test case.

tests/func/test_suspended_frames.py
```diff
@@ -0,0 +1,29 @@
+import sys
```
fabioz: The last thing is that a unit-test like this should be on the `pydevd` side of things (so, you can just add a new test to `/ptvsd/src/ptvsd/_vendored/pydevd/tests_python/test_suspended_frames_manager.py` instead of creating a new file -- I'll take care of backporting to `pydevd` so that it runs properly as `pydevd` tests don't run on the `ptvsd` repo).

Otherwise, it should be a functional test which actually uses `ptvsd`.

fabioz:
Just merged. Thank you for the fix ;)

Issue #1239, opened 2019-03-15
Attach to process fails with "RuntimeError: Could not find dll file" on Ubuntu 14.04


sunew:
## Environment data

- PTVSD version: 4.2.4
- OS and version: Ubuntu 14.04.5 LTS
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.6, virtualenv
- Using VS Code or Visual Studio: vscode

## Actual behavior

With virtualenv acitvated:

`python2.7 -m ptvsd --host 10.5.10.5 --port 3000 --pid 27723`

Gives this traceback: 

```

Attaching with arch: i386:x86-64
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/opt/elasticbox/working/elasticbox/lib/python2.7/site-packages/ptvsd/__main__.py", line 369, in <module>
    main(sys.argv)
  File "/opt/elasticbox/working/elasticbox/lib/python2.7/site-packages/ptvsd/__main__.py", line 357, in main
    run()
  File "/opt/elasticbox/working/elasticbox/lib/python2.7/site-packages/ptvsd/__main__.py", line 340, in attach_to_pid
    run_python_code(pid, code, connect_debugger_tracing=True)
  File "/opt/elasticbox/working/elasticbox/local/lib/python2.7/site-packages/ptvsd/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py", line 419, in run_python_code_linux
    raise RuntimeError('Could not find dll file to inject: %s' % target_dll)
RuntimeError: Could not find dll file to inject: /opt/elasticbox/working/elasticbox/local/lib/python2.7/site-packages/ptvsd/_vendored/pydevd/pydevd_attach_to_process/attach_linux_amd64.so

```

## Expected  behavior

A successful attachment, making it possible to connect from vscode on the ip and port.

## Steps to reproduce:
See above.

karthiknadig:
Similar to #863, but it fails in a different area. @sunew I recommend using ptvsd to directly launch the code you want to debug, or use `ptvsd.enable_attach()` in your script. We are currently working on other areas of the debugger and will get to the attach to pid issue later.

sunew:
[...]

Issue #863, opened 2018-10-02
Attach to pid fails on Linux


andyneff:
## Environment data

- PTVSD version: 4.1.3
- OS and version: Fedora 28 and Debian 9
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.15, 3.6.6, 3.7.0
- Using VS Code or Visual Studio: VS Code, but not even connected at this point

## Actual behavior

1. `pipenv run python -m ptvsd --port 5678 --host localhost --pid 25226`

### Python 2.7 Fedora 28 host and Fedora 28 docker

```
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/andy/.local/share/virtualenvs/shm-D-JgbiGU/lib/python2.7/site-packages/ptvsd/__main__.py", line 227, in <module>
    main()
  File "/home/andy/.local/share/virtualenvs/shm-D-JgbiGU/lib/python2.7/site-packages/ptvsd/__main__.py", line 223, in main
    wait=args.wait)
  File "/home/andy/.local/share/virtualenvs/shm-D-JgbiGU/lib/python2.7/site-packages/ptvsd/__main__.py", line 212, in handle_args
    attach_main(addr, name, *extra, **kwargs)
  File "/home/andy/.local/share/virtualenvs/shm-D-JgbiGU/lib/python2.7/site-packages/ptvsd/_attach.py", line 42, in attach_main
    run_python_code(pid, code, connect_debugger_tracing=True)
  File "/home/andy/.local/share/virtualenvs/shm-D-JgbiGU/lib/python2.7/site-packages/ptvsd/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py", line 400, in run_python_code_linux
    assert '\'' not in python_code, 'Having a single quote messes with our command.'
AssertionError: Having a single quote messes with our command.
```

### Python 3.6 on Fedora 28 host and Fedora 28 docker

```
Attaching with arch: i386:x86-64
[...]

PR #1293, opened 2019-04-02
Fix single quote issue during process attach

martinRenou:
This fixes the `AssertionError: Having a single quote messes with our command.` error cited here #863.

I then have a the same traceback as here #1239.

### DIFF
--- a/src/ptvsd/__main__.py
+++ b/src/ptvsd/__main__.py
@@ -349,7 +349,7 @@ ptvsd.options.port = {port}
 
 import ptvsd.log
 ptvsd.log.to_file()
-ptvsd.log.info('Debugger successfully injected')
+ptvsd.log.info("Debugger successfully injected")
 
 if ptvsd.options.client:
     from ptvsd._remote import attach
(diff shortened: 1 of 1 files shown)


src/ptvsd/__main__.py
```diff
@@ -349,7 +349,7 @@ def escape(ch):
 
 import ptvsd.log
 ptvsd.log.to_file()
-ptvsd.log.info('Debugger successfully injected')
+ptvsd.log.info("Debugger successfully injected")
```
int19h: I really should have known better, since I wrote the comment above. Should we add an assert here maybe? Right after `format`.

Also, I think the `print(code)` below is some debugging leftover. It's not a bad idea to log it, but it shouldn't be printed to stdout.

martinRenou: Will do :)

src/ptvsd/__main__.py
```diff
@@ -349,17 +349,17 @@ def escape(ch):
...
     from ptvsd._remote import attach
     attach(({host}, {port}))
 else:
     ptvsd.enable_attach()
 '''.format(**locals())
-    print(code)
+    assert "'" not in code
 
-    ptvsd.log.debug('Injecting debugger into target process.')
+    ptvsd.log.debug('Injecting debugger into target process: \n"""{}\n"""'.format(code))
```
int19h: This needs to be `"... {0} ... "`, because we're using `string.Formatter` rather than `format` for logging, and in Python 2 it doesn't understand unnumbered placeholders like `{}`. Otherwise looks good!

int19h: Also, can you swap the logging and the assert? This way, if the assert gets tripped, we can see what exactly went wrong in the log - e.g. if one of the formatted values ends up introducing a single quote, we want to know which one did that.

martinRenou: Sure, it makes more sense

PR #1309, opened 2019-04-04
Add setting to limit number of lines in stack for exception details. Fixes #582

### LINKED ISSUE #582: Add ability to limit number of lines in stack trace returned for exceptions
## Environment data

- PTVSD version: Master
- OS and version: N/A

## Suggestion
* VSC will pass some setting to PTVSD
* The presence of this setting tells PTVSD to limit number of lines (frames) in stack trace returned with the exception info.

https://github.com/Microsoft/vscode-python/issues/2075#issuecomment-402299091

karthiknadig:
`maxExceptionStackFrames: 5` in launch json should limit the number of frames to 5, lines to ~10. The exact number of lines depends on the stack situation. Default is all frames, this has to be the default to support VS style exception info UI. Since VS did not have this setting and adding a setting there is non trivial.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -1143,10 +1143,7 @@ def internal_get_exception_details_json(dbg, request, thread_id, max_frames, set
             except:
                 pass
 
-        if max_frames == 0:
-            stack_str = ''.join(traceback.format_list(frames))
-        else:
-            stack_str = ''.join(traceback.format_list(frames[-max_frames:]))
+        stack_str = ''.join(traceback.format_list(frames[-max_frames:]))
 
         # This is an extra bit of data used by Visual Studio
         source_path = frames[0][0] if frames else ''
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -1077,7 +1077,7 @@ def internal_get_description(dbg, seq, thread_id, frame_id, expression):
         dbg.writer.add_command(cmd)
 
 
-def internal_get_exception_details_json(dbg, request, thread_id, set_additional_thread_info=None, iter_visible_frames_info=None):
+def internal_get_exception_details_json(dbg, request, thread_id, max_frames, set_additional_thread_info=None, iter_visible_frames_info=None):
     ''' Fetch exception details
     '''
     try:
@@ -1143,7 +1143,10 @@ def internal_get_exception_details_json(dbg, request, thread_id, set_additional_
             except:
                 pass
 
-        stack_str = ''.join(traceback.format_list(frames))
+        if max_frames == 0:
+            stack_str = ''.join(traceback.format_list(frames))
+        else:
+            stack_str = ''.join(traceback.format_list(frames[-max_frames:]))
 
         # This is an extra bit of data used by Visual Studio
         source_path = frames[0][0] if frames else ''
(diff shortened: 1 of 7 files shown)


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -1130,7 +1130,10 @@ def internal_get_exception_details_json(dbg, request, thread_id, set_additional_
             name = 'exception: type unknown'
             description = 'exception: no description'
 
-        stack_str = ''.join(traceback.format_list(frames))
+        if max_frames == 0:
+            stack_str = ''.join(traceback.format_list(frames))
+        else:
+            stack_str = ''.join(traceback.format_list(frames[-max_frames:]))
```
int19h: Doesn't look like you need an `if` here, since `frames[-0:]` will get the whole thing just the same.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -399,8 +401,9 @@ def on_exceptioninfo_request(self, py_db, request):
         '''
         # : :type exception_into_arguments: ExceptionInfoArguments
         exception_into_arguments = request.arguments
-        thread_id = exception_into_arguments.threadId
-        self.api.request_exception_info_json(py_db, request, thread_id)
+        thread_id = exception_into_arguments.threadId            
+        max_frames = int(self._debug_options['args'].get('maxExcpetionStackFrames', 0))
```
int19h: Spelling,

tests/func/test_exception.py
```diff
@@ -208,3 +208,73 @@ def code_to_debug():
...
+        hit = session.wait_for_thread_stopped(reason='exception')
+        frames = hit.stacktrace.body['stackFrames']
+        assert frames[0]['line'] == line_numbers['unhandled']
+
+        resp_exc_info = session.send_request('exceptionInfo', {
+            'threadId': hit.thread_id
+        }).wait_for_response()
+
+        expected = ANY.dict_with({
+            'exceptionId': ANY.such_that(lambda s: s.endswith('ArithmeticError')),
```
int19h: `pattern.Regex` might come in handy here and below - it prints better in the error output if assert fails.

tests/func/test_exception.py
```diff
@@ -208,3 +208,73 @@ def code_to_debug():
...
+                'typeName': ANY.such_that(lambda s: s.endswith('ArithmeticError')),
+                'message': 'bad code',
+                'source': Path(code_to_debug),
+            }),
+        })
+        assert resp_exc_info.body == expected
+        stack_str = resp_exc_info.body['details']['stackTrace']
+        stack_line_count = len(stack_str.split('\n'))
+        assert stack_line_count >= min_expected_lines
+        assert stack_line_count <= max_expected_lines
```
int19h: You can chain these:
```py
assert min_expected_lines <= stack_line_count <= max_expected_lines
```

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -1065,11 +1065,28 @@ def test_evaluate(case_setup):
...
+def test_exception_details(case_setup, max_frames):
+    with case_setup.test_file('_debugger_case_large_exception_stack.py') as writer:
         json_facade = JsonFacade(writer)
 
         writer.write_set_protocol('http_json')
+        if max_frames == 'all':
+            json_facade.write_launch(maxExcpetionStackFrames=0)
+            # trace back compresses repeated text
+            min_expected_lines = 10
+            max_expected_lines = 1000
```
fabioz: Given that the difference is so large, it doesn't seem like this test will fail when it should if something goes awry. 

Maybe you can change the code to have 2 methods recursing (that way I think Python won't compress it as the same function won't repeat over and over in the traceback) -- likewise for the test in `ptvsd`.

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -1065,11 +1065,28 @@ def test_evaluate(case_setup):
...
+        elif max_frames == 'default':
+            json_facade.write_launch()
+            # default is all frames
+            # trace back compresses repeated text
+            min_expected_lines = 10
+            max_expected_lines = 1000
+        else:
+            json_facade.write_launch(maxExcpetionStackFrames=max_frames)
```
fabioz: Note that the `maxExcpetionStackFrames` typo needs to be fixed across the pull request.

Issue #1276, opened 2019-03-28
Debugger crashes when using Django's Development Server with a Proper Launch Profile


DonJayamanne:
@pymike00 commented on [Thu Mar 28 2019](https://github.com/Microsoft/vscode-python/issues/4988)

## Environment data

- VS Code version: 1.32.2-1
- Extension version: 2019.3.6139
- OS and version: Linux Manjaro XFCE x86_64 18.0.4
- Python version: 3.7.2
- Type of virtual environment used: venv 
- Relevant/affected Python packages and their versions: Django 2.1.7

## Expected behaviour

Using the Python Extension on Visual Studio Code to run Django's development server as explained in the docs, [https://code.visualstudio.com/docs/python/tutorial-django#_create-a-debugger-launch-profile](https://code.visualstudio.com/docs/python/tutorial-django#_create-a-debugger-launch-profile), without any problems.

## Actual behaviour

Can't use the debugger to launch Django's development server; As soon as you try to use the site, the debugger pauses on the exception VariableDoesNotExist, even on a fresh Django Project.

The problem does not manifest itself when using a previous version of the Python Extension (such as 2019.2.5558), nor it manifests itself when running Django's development server from a shell, or even from VS Code's Terminal.

## Steps to reproduce:

1. Create a Virtual Environment with VENV and install Django.
2. Create New Project: django-admin startproject NewProject
3. Enter Folder: cd NewProject 
3. Open The Editor: code .
4. Select Python Interpreter to be the one of the previously created VENV
5. Create debugger Launch profile as explained in the docs. Comment out "no threading"
6. Launch dev Server with the newly created Launch profile (terminal will show no problems)
7. Open Chrome, go to localhost:8000. Hello World Page will load, but VS Code will show "Could not load source '<unknown source>': Source unavailable.". Debugger paused on exception <unknown source>
8. Go to  localhost:8000/admin/. Debugger will pause on exception VariableDoesNotExist as shown below.

## Logs
Impossible to get the output as explained from the Output panel in Python: it is empty and the development server doesn't show any problem. **Everything works fine if running Django's development server from a shell or from the terminal**. The log below is the message shown in red (as shown in the screenshot) with the VariableDoesNotExist exception.

```
Exception has occurred: VariableDoesNotExist
[...]

PR #1310, opened 2019-04-04
Django 2.2 tests and fixes for exception breakpoints in Django 1.7. Fixes #1298, #1294

### LINKED ISSUE #1294: Make pydevd tests work with Django 2.2
Right now only 1.7 and 2.1 are tested and 2.2 is already available.

### LINKED ISSUE #1298: Show message when stepping into library code with justMyCode: true
If debugStdLib is false and the user attempts to step in, we should show a message telling the user that the debugger will skip. It should also tell them how to change the setting so that they get the intended behavior.

This is similar to how Just-my-code setting is handled in VS for C#

fabioz:
Fixes #1298, Fixes #1294


fabioz:
@karthiknadig I've added the fix for #1276 to this branch as it was also related to django, so, please re-review ;)

PR #1317, opened 2019-04-04
Fix #1296 SyntaxError for breakpoints with empty condition

### LINKED ISSUE #1296: SyntaxError for breakpoints with empty condition
@xiaofeng0123 commented on [Mon Apr 01 2019](https://github.com/Microsoft/vscode-python/issues/5071)

Visual Studio Code - Insiders: 1.33.0-insider
python-insider: 2019.4.6304-alpha

Repro step:
1. Input below code:
     for i in range(1,500,20):
           print(i)
2. Set "Expression" condition breakpoint such as "i>400"
3. Start to debug make sure it works well
4. Right-click on the breakpoint to select "Edit Breakpoint"
5. Remove the expression in the dialog and then press "Enter"
6. Right-click on the breakpoint again to select "Edit Breakpoint"
7. Set "Hit Count" condition breakpoint such as "20"
8. Start to debug

Actual Result:
Error displays on "TERMINAL" panel
image

Expect Result:
No error displays

**Note:**
If remove the "Expression" breakpoint. Then add "Hit Count" breakpoint, it will no error occurs.
Repro gif: ‪\\bvtsrv2\team\v-xiaofz\PVSC\HitCountConditionBreakpoint.gif

int19h:
Treat breakpoints with `condition: ""` as non-conditional.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
@@ -59,10 +59,10 @@ class LineBreakpoint(object):
 
     @property
     def has_condition(self):
-        return self.condition is not None or self.hit_condition is not None
+        return bool(self.condition) or bool(self.hit_condition)
 
     def handle_hit_condition(self, frame):
-        if self.hit_condition is None:
+        if not self.hit_condition:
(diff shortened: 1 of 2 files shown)


int19h:
It's a temp mitigation - we're going to file this on VSC, since it really shouldn't be doing it this way.

PR #1328, opened 2019-04-09
 Normalize path translation files according to client OS. Fixes #1311

### LINKED ISSUE #1311: Remote debugging failed with warning "pydev debugger: warning: try to add breakpoint to file that does not exist"
@xiaofeng0123 commented on [Tue Apr 02 2019](https://github.com/Microsoft/vscode-python/issues/5094)

Visual Studio Code - Insiders: 1.33.0-insider
python-insider: 2019.4.6365-alpha
ptvsd version: 4.2.6 

Repro step:
1. Create a python file named remote_debug.py with below code:
     import ptvsd
     ptvsd.enable_attach(('IP of unbuntu',5678))
     ptvsd.wait_for_attach()
     for i in range(1,10,2):
            print(i)
     print('demo done')
2. Copy this file to ubuntu which should include ptvsd installed with same version as current machine
3. Execute python file in ubuntu:
    python3 remote_debug.py
4. In VSC, set DEBUG to "Python:Remote Attach", and set the path mappings and host value
image
5. Set breakpoint on current machine and press F5 to start debugging

Actual Result:
Fail to hit breakpoint with warning "pydev debugger: warning: trying to add breakpoint to file that does not exist..."
image

Expect Result:
It should hit breakpoint

fabioz:


src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
```diff
@@ -195,20 +199,38 @@ def set_ide_os(os):
...
+        print('pydev debugger: client OS: %s' % (os,))
+
+    _normcase_from_client = normcase
+    if os == 'WINDOWS':
+
+        # Client in Windows and server in Unix, we need to normalize the case.
+        if not IS_WINDOWS:
+
+            def _normcase_from_client(filename):
+                return filename.lower()
```
int19h: Since we're using the standard `normcase` elsewhere, for consistency, perhaps it's better to import `normcase` from `ntpath` / `posixpath` / `macpath` as needed here, depending on the IDE platform? I believe that's what we did in the old implementation in this scenario.

src/ptvsd/wrapper.py
```diff
@@ -1334,16 +1334,36 @@ def _initialize_path_maps(self, args):
...
+        msg = '1.1\t{}\tID'.format(self._client_os_type)
         return self.pydevd_request(cmd, msg)
 
     @async_handler
     def _handle_launch_or_attach(self, request, args):
         self._path_mappings_received = True
 
+        client_os_type = self.debug_options.get('CLIENT_OS_TYPE', '').upper().strip()
+        if client_os_type and client_os_type not in ('WINDOWS', 'UNIX'):
+            ptvsd.log.warn('Invalid CLIENT_OS_TYPE passed: %s (must be either "WINDOWS" or "UNIX").' % (client_os_type,))
```
int19h: `warn` is a wrapper around `string.Formatter`, so you can do this:
```py
ptvsd.log.warn('Invalid CLIENT_OS_TYPE passed: {0} (must be either "WINDOWS" or "UNIX").', client_os_type)
```
Note that it must be `{0}` rather than `{}`, because `Formatter` doesn't support the latter on Python 2.

fabioz: Done

src/ptvsd/_vendored/pydevd/pydevd_file_utils.py
fabioz: I've done some changes to use the same `normcase` implementation everywhere.

Now, just to note, the actual implementation we use is not just `ntpath.normcase` (on Windows) nor do we use the regular `macpath.normcase` nor `posixpath.normcase` implementations either... I'll try to explain a bit the rationale there:

On `linux` it should always be a `no-op` anyways, so, there isn't a great difference from the default implementation (we just skip some type checks), but on mac, the implementation actually defaults to making it all lower (as on windows) because depending on the filesystem that's being used on `mac` it may be case insensitive or not... 

Now, we don't run into as many problems on mac as on windows because most of the problems in windows are actually due to the drive letter (it's strange, but most of the casing problems aren't really due to the casing of the path which is usually correct, but due to the drive letter which is something that each tool appears to give different, whereas there's no drive letter on mac)... this has always been that way in the debugger -- so, maybe something to keep an eye open for and we could think about changing it in the future, but I wouldn't like to change that in this pull request.

PR #1346, opened 2019-04-11
 Set don't trace ptvsd files via json message. Fixes #1220

### LINKED ISSUE #1220: Replace get_file_type used to ignore debugger files with a json based setting

karthiknadig:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/_debug_adapter/pydevd_custom_schema.py
```diff
@@ -0,0 +1,180 @@
+from .pydevd_base_schema import BaseSchema, register, register_request, register_response
```
fabioz: The schema file is currently generated by `_pydevd_bundle._debug_adapter.__main__pydevd_gen_debug_adapter_protocol` and I think that instead of manually coding the Python custom schema we should create a `debugProtocolCustom.json` and in `gen_debugger_protocol` load both files creating a dict where the `definitions` contains contents of both (if the generator changes to create a different structure in the future we shouldn't have to manually update a separate file).

src/ptvsd/wrapper.py
```diff
@@ -1745,6 +1732,10 @@ def on_setDebuggerProperty(self, request, args):
             jmc = int(args.get('JustMyCodeStepping', 0)) > 0
             self.debug_options['DEBUG_STDLIB'] = not jmc
 
+        pydevd_request = copy.deepcopy(request)
```
fabioz: Could we use `_forward_request_to_pydevd` here? Note that this wouldn't need to be an `async_handler` in this case.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -563,12 +587,18 @@ def get_file_type(self, abs_real_path_and_basename):
             None:
                 If it's a regular user file which should be traced.
         '''
-        basename = abs_real_path_and_basename[-1]
-        if basename.startswith('<frozen '):
-            # In Python 3.7 "<frozen ..." appear multiple times during import and should be
-            # ignored for the user.
-            return self.PYDEV_FILE
-        return self._dont_trace_get_file_type(basename)
+        abs_path = abs_real_path_and_basename[0]
```
fabioz: Really minor optimizations (but this is in the critical path performance-wise for the debugger, so, I think it's worth it):

Start the function as:

```
try:
    return _cache_file_type[abs_real_path_and_basename[0]]  #Don't assign abs_path unless we have a KeyError.
except:  # Notice: catch all exception to avoid having to load the `KeyError` name.
    abs_path = abs_real_path_and_basename[0]
```

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -546,7 +546,31 @@ def handle_breakpoint_expression(self, pybreakpoint, info, new_frame):
...
+            True :
+                If files should NOT be traced.
+
+            False:
+                If files should be traced.
+        '''
+        # By default all external files are traced.
+        return False
+
+    def get_file_type(self, abs_real_path_and_basename, _cache_file_type={}):
```
fabioz: The `_cache_file_type` should to be defined in the module level (you can still assign to the local in the function definition: `_cache_file_type=_CACHE_FILE_TYPE`) and we should check that it's empty when `on_setdebuggerproperty_request` is called as the result of `get_file_type` must always be the same (there are other caches throughout the debugger which rely on having always the same file type, so, `on_setdebuggerproperty_request` cannot change that invariant -- after we started debugging the `on_setdebuggerproperty_request` for that parameter can no longer be changed).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -663,5 +664,28 @@ def on_goto_request(self, py_db, request):
         response = pydevd_base_schema.build_response(request, kwargs={'body': {}})
         return NetCommand(CMD_RETURN, 0, response, is_json=True)
 
+    def on_setdebuggerproperty_request(self, py_db, request):
+        args = request.arguments.kwargs
+        if 'dontTraceStartPatterns' in args and 'dontTraceEndPatterns' in args:
+            start_patterns = tuple(args['dontTraceStartPatterns'])
+            end_patterns = tuple(args['dontTraceEndPatterns'])
+
+            def dont_trace_files(abs_path):
```
fabioz: We should check that after we receive the `dontTraceStartPatterns` / `dontTraceEndPatterns`, if it's called again it's always called with the same parameters and that `get_file_type` hasn't been called before this request (check if something is already cached) as `get_file_type` is not expected to change its return value (if that's needed, more work is needed in the debugger as other caches rely on this).

You can do this all local to this function (maybe rename `dont_trace_files` to `dont_trace_files_property_request` and check the function name and add the `start_patterns` and the `end_patterns` to the function itself to see if they're still equal and show a critical warning in stderr and return a failure response if the invariant is broken -- please create a test case for this specific situation).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -17,6 +17,7 @@
 from _pydevd_bundle.pydevd_net_command import NetCommand
 from _pydevd_bundle.pydevd_utils import convert_dap_log_message_to_expression
 import pydevd_file_utils
+import pydevd
```
fabioz: This import seems unused.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -663,5 +664,48 @@ def on_goto_request(self, py_db, request):
...
+                return True
+        return False
+
+    def on_setdebuggerproperty_request(self, py_db, request):
+        args = request.arguments.kwargs
+        if 'dontTraceStartPatterns' in args and 'dontTraceEndPatterns' in args:
+            start_patterns = tuple(args['dontTraceStartPatterns'])
+            end_patterns = tuple(args['dontTraceEndPatterns'])
+            if self._can_set_dont_trace_pattern(py_db, start_patterns, end_patterns):
+                def dont_trace_files(abs_path):
```
fabioz: I think that this method could be renamed to `dont_trace_files_property_request`.

-- that way when checking `if py_db.dont_trace_external_files.__name__ == 'dont_trace_files'` I think it helps a bit in understanding that it's a patched function.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -663,5 +664,48 @@ def on_goto_request(self, py_db, request):
         response = pydevd_base_schema.build_response(request, kwargs={'body': {}})
         return NetCommand(CMD_RETURN, 0, response, is_json=True)
 
+    def _can_set_dont_trace_pattern(self, py_db, start_patterns, end_patterns):
+        if py_db.is_cache_file_type_empty():
```
fabioz: Actually, as long as it's empty I think it should be ok to set right? (so, it should return as `success` if empty or if it's the same thing which was already before).

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -1239,6 +1239,62 @@ def test_goto(case_setup):
...
+            assert not dbg_response.success
+
+        stack_trace_request = json_facade.write_request(
+            pydevd_schema.StackTraceRequest(pydevd_schema.StackTraceArguments(threadId=hit.thread_id)))
+        stack_trace_response = json_facade.wait_for_response(stack_trace_request)
+        dont_trace_frames = list(frame for frame in stack_trace_response.body.stackFrames
+                                 if frame['source']['path'].endswith('dont_trace.py'))
+
+        if dbg_property in ('dont_trace', 'change_pattern', 'dont_trace_after_start'):
+            # Since chagnge pattern or don't trace after start is expected to fail,
```
fabioz: typo: `chagnge` -> `change`

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
karthiknadig: For clarification:
* If cache is not empty we should fail the don't trace request.
* If cache is empty we show allow setting the patterns.

Is there any need to check the patterns at all?

fabioz: Humm, not really... I was thinking about something as:

```
if cache is empty:
    return success=True

if pattern was set and new pattern is still the same: 
    return success=True  (I think this can happen when reconnecting)

return success=False
```

src/ptvsd/wrapper.py
```diff
@@ -1739,12 +1726,13 @@ def version_str(v):
...
 
     # VS specific custom message handlers
-    @async_handler
+
     def on_setDebuggerProperty(self, request, args):
         if 'JustMyCodeStepping' in args:
             jmc = int(args.get('JustMyCodeStepping', 0)) > 0
             self.debug_options['DEBUG_STDLIB'] = not jmc
 
+        self._forward_request_to_pydevd(request, args)
```
fabioz: `_forward_request_to_pydevd` already sends a response, so, it should no longer send the response in this method.

karthiknadig:
This is somehow breaking `breakpoint()` function. Investigating.

karthiknadig:
@fabioz see https://github.com/Microsoft/ptvsd/pull/1346/commits/dbce088c5a2161ec15aa0a826c9f1798a64a5f11. For some reason if that is `False` (which it should be), `breakpoint()` stops working on 3.7, when `ptvsd` is launched using cli. Not exactly sure what path it is taking that this fails. Note that i can't repro it locally, it seems to consistently fail on CI. I suspect some race some where that triggers on that path.

Failure results: https://travis-ci.org/Microsoft/ptvsd/jobs/519479016
Success results: https://travis-ci.org/Microsoft/ptvsd/jobs/519483419

karthiknadig:
It looks like in some cases `get_file_type` gets called even before we have had the opportunity to set the patterns. The `breakpoint()` issue on 3.7 is a side effect of that. I think addressing that has to be done in a separate PR where we make sure that no debugger operations occur until all settings are done, essentially support `initialize` and `configurationDone`. as intended by DAP.

fabioz:
> It looks like in some cases `get_file_type` gets called even before we have had the opportunity to set the patterns. The `breakpoint()` issue on 3.7 is a side effect of that. I think addressing that has to be done in a separate PR where we make sure that no debugger operations occur until all settings are done, essentially support `initialize` and `configurationDone`. as intended by DAP.

I think we should hold off this one until that's done then as the current approach is working now and this will break it (created https://github.com/Microsoft/ptvsd/issues/1355 to take a look at that).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -663,5 +663,48 @@ def on_goto_request(self, py_db, request):
...
+
+        if py_db.dont_trace_external_files.__name__ == 'dont_trace_files_property_request':
+            return py_db.dont_trace_external_files.start_patterns == start_patterns and \
+                py_db.dont_trace_external_files.end_patterns == end_patterns
+
+        return True
+
+    def on_setdebuggerproperty_request(self, py_db, request):
+        args = request.arguments.kwargs
+        if 'dontTraceStartPatterns' in args and 'dontTraceEndPatterns' in args:
```
int19h: This feels like it would be more straightforward with a regex instead (which could easily cover both cases, and then some). Or is there a perf issue with them on a hot path here?

src/ptvsd/_vendored/pydevd/tests_python/resources/_debugger_case_dont_trace.py
```diff
@@ -0,0 +1,3 @@
+def call_me_back(callback):
+    if callback is not None and callable(callback):
```
int19h: The first check is redundant, since `None` is not callable.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
karthiknadig: I don't have a preference here. I can change it either way. For now I will be making an update to this where we will be disabling this stuff. temporarily until we address #1355

karthiknadig:
With the last change we should still get the old behavior. The new code is not used anywhere except in pydevd test. we should re-enable those as needed.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -663,5 +663,48 @@ def on_goto_request(self, py_db, request):
...
+                dont_trace_files_property_request.start_patterns = start_patterns
+                dont_trace_files_property_request.end_patterns = end_patterns
+                py_db.dont_trace_external_files = dont_trace_files_property_request
+            else:
+                # don't trace pattern cannot be changed after it is set once. there are caches
+                # throughout the debugger which rely on having always the same file type,
+                message = ("Calls to set or change don't trace patterns (via setDebuggerProperty) are not "
+                           "allowed since debugging has already started or don't trace patterns are already set.")
+                response_args = {'success':False, 'body': {}, 'message': message}
+                response = pydevd_base_schema.build_response(request, kwargs=response_args)
```
fabioz: I think we should fail louder here (i.e.: at least print to stderr so that we get an error report) as other things may function badly because of that.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -563,12 +588,17 @@ def get_file_type(self, abs_real_path_and_basename):
...
+            return _cache_file_type[abs_real_path_and_basename[0]]
+        except:
+            file_type = self._internal_get_file_type(abs_real_path_and_basename)
+            if file_type is None:
+                file_type = PYDEV_FILE if self.dont_trace_external_files(abs_real_path_and_basename[0]) else None
+            _cache_file_type[abs_real_path_and_basename[0]] = file_type
+            return file_type
+
+    def is_cache_file_type_empty(self):
+        return _CACHE_FILE_TYPE == {}
```
fabioz: I think `bool(_CACHE_FILE_TYPE)` may be better...

PR #1360, opened 2019-04-15
Turn off sending continued events by default.

### LINKED ISSUE #1358: 'set next statement' does not work when 'continued' event is sent on VS
Looking at tests there are very few places where we make use of `continued` event. We should use env variable to trigger sending `continued` event for testing purposes.

### LINKED ISSUE #1361: Step over a thread join hangs debugger
@dpoole73 commented on [Mon Apr 15 2019](https://github.com/Microsoft/vscode-python/issues/5293)

## Environment data

- VS Code version: 1.33.1
- Extension version (available under the Extensions sidebar): 1029-3-6558
- OS and version: MacOS version 10.14.4
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.3
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A (reproed with and without venv) 
- Relevant/affected Python packages and their versions: threading, time

## Expected behaviour

F10 (step over) a join operation should move to the next statement or block while threads run

## Actual behaviour

Step-over appears to hang the debugging until you break in and resume the thread showing "PAUSED ON PAUSE" status

## Steps to reproduce:

I have attached a gif with the steps below plus showing that it doesn't repro when continuing rather than stepping over.
Untitled

1. create a python file with the code below
```
import threading
import time

class RunThread(threading.Thread):
    kind = 'Thread'
    def __init__(self, name, fn, args = ()):
        threading.Thread.__init__(self)
        self.name = name
        self.fn   = fn
        self.args = args
        self.start()
    def run(self):
        self.fn(*(self.args))

def th1():
    while True:
        print('th1')
        time.sleep(1)

def th2():
    while True:
        print('th2')
        time.sleep(2)

def th3():
    while True:
        print('th3')
        time.sleep(4)

def main():
    threads = []
    threads.append(RunThread("aaa", th1))
    threads.append(RunThread("bbb", th2))
    threads.append(RunThread("ccc", th3))

    for t in threads:
        t.join()

if __name__ == '__main__':
    main()
```
1. Set breakpoint on line 37
1. Debug using the python current file default settings
1. when the breakpoint is hit, hit F10 (or step over).
1. after a few seconds break in by hitting pause, find the "PAUSED ON PAUSE" thread and continue it, execution will continue

The same thing doesn't happen if you continue rather than stepping over the join() call.

This impacts not only this scenario but seemingly anything that uses threads

## Logs
```
Starting Microsoft Python language server.
[Info  - 3:42:41 PM] GetCurrentSearchPaths /usr/local/opt/python/bin/python3.7 
[Info  - 3:42:42 PM] Python search paths:
[Info  - 3:42:42 PM]     /private/var/folders/mm/2w6btf515j36vd2s3_2p6j740000gn/T/sr2eql2k.px2
[Info  - 3:42:42 PM]     /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7
[Info  - 3:42:42 PM]     /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/lib-dynload
[Info  - 3:42:42 PM]     /Users/dpoole/Library/Python/3.7/lib/python/site-packages
[Info  - 3:42:42 PM]     /usr/local/lib/python3.7/site-packages
[Info  - 3:42:42 PM] Configuration search paths:
[Info  - 3:42:42 PM]     /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python37.zip
[Info  - 3:42:42 PM]     /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7
[Info  - 3:42:42 PM]     /usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/lib-dynload
[Info  - 3:42:42 PM]     /Users/dpoole/Library/Python/3.7/lib/python/site-packages
[Info  - 3:42:42 PM]     /usr/local/lib/python3.7/site-packages
[Info  - 3:42:42 PM]     /usr/local/opt/python/bin
[Info  - 3:42:42 PM] Microsoft Python Language Server version 0.2.47.0
[Info  - 3:42:42 PM] Initializing for /usr/local/opt/python/bin/python3.7
```
# Console

```
dpoole@dpoole-mac:/private/tmp/repro $ cd /private/tmp/repro ; env PYTHONIOENCODING=UTF-8 PYTHONUNBUFFERED=1 /usr/local/opt/python/bin/python3.7 /Users/dpoole/.vscode/extensions/ms-python.python-2019.3.6558/pythonFiles/ptvsd_launcher.py --default --client --host localhost --port 54810 /private/tmp/repro/repro.py 
th3th2
```

karthiknadig:
Fixes #1358, Fixes #1361


tests/func/test_step.py
```diff
@@ -33,8 +33,8 @@ def func():
         session.initialize(
             target=(run_as, code_to_debug),
             start_method=start_method,
-            env={'PTVSD_USE_CONTINUED': '1'},
             ignore_unobserved=[Event('continued')],
+            env={str('PTVSD_USE_CONTINUED'): str('1')},
```
int19h: If this is always necessary on 2.7, `DebugSession` should take care of it on its end.

PR #1380, opened 2019-04-22
Move source request to pydevd. Fixes #1333

### LINKED ISSUE #1333: on_source special cases should be ported to pydevd and it should use _forward_request_to_pydevd
Note: `_forward_request_to_pydevd` should `send_error_response` when needed.

karthiknadig:
Fixes #1333

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
@@ -702,7 +702,9 @@ class _PyDevJsonCommandProcessor(object):
         response_args = {'body': body}
 
         if content is None:
-            if server_filename:
+            if source_reference == 0:
+                message = 'Source unavailable'
+            elif server_filename:
                 message = 'Unable to retrieve source for %s' % (server_filename,)
             else:
                 message = 'Invalid sourceReference %d' % (source_reference,)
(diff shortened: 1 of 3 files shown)


src/ptvsd/wrapper.py
```diff
@@ -1423,23 +1423,7 @@ def on_threads(self, request, args):
...
-            _, _, resp_args = yield self.pydevd_request(
-                pydevd_comm.CMD_LOAD_SOURCE,
-                pydevd_request,
-                is_json=True)
-
-            if resp_args.get('success', True):
-                self.send_response(request, **resp_args['body'])
-            else:
-                self.send_error_response(request, resp_args.get('message', 'Error retrieving source'))
+        self._forward_request_to_pydevd(request, args)
```
fabioz: This should no longer be `async_handler`.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -702,7 +702,9 @@ def on_source_request(self, py_db, request):
         response_args = {'body': body}
 
         if content is None:
-            if server_filename:
+            if source_reference == 0:
```
fabioz: Can you double-check if there are tests for this case? (even if only in the `ptvsd` side)

src/ptvsd/wrapper.py
karthiknadig: oops

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
karthiknadig: Updated `test_with_path_mappings` to test this case.

Issue #1332, opened 2019-04-10
DAP: wrapper still uses on_threads


fabioz:
This could be forwarded when #1331 is done.

PR #1384, opened 2019-04-23
Thread id from pydevd is now int32. Fixes #1331, #1339, #1332

### LINKED ISSUE #1331: DAP: use threadId as int32.
Right now the thread id in `pydevd` is a string and it's exposed in the protocol. We should change what goes in the protocol for an `int32`.

fabioz:
With this change, all the thread notifications come from pydevd through the DAP.

I think this is a big change in how things work internally, so, maybe it'd be nice to do some manual testing especially in disconnect/reconnect scenarios with threads/django/flask/multiprocessing (I haven't actually checked -- do we support disconnect/reconnect in multiprocessing scenarios?)

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -97,6 +97,7 @@ import sys
 import traceback
 from _pydevd_bundle.pydevd_utils import quote_smart as quote, compare_object_attrs_key
 from _pydev_bundle import pydev_log
+from _pydev_bundle.pydev_log import exception as pydev_log_exception
 from _pydev_bundle import _pydev_completer
 
 from pydevd_tracing import get_exception_traceback_str
@@ -116,20 +117,6 @@ from _pydevd_bundle.pydevd_comm_constants import *  # @UnusedWildImport
 
 if IS_JYTHON:
     import org.python.core as JyCore  # @UnresolvedImport
-
-
-def pydevd_log(level, *args):
-    ''' levels are:
-        0 most serious warnings/errors
-        1 warnings/significant events
-        2 informational trace
-    '''
-    if level <= DebugInfoHolder.DEBUG_TRACE_LEVEL:
-        # yes, we can have errors printing if the console of the program has been finished (and we're still trying to print something)
-        try:
-            sys.stderr.write('%s\n' % (args,))
-        except:
-            pass
 
 
 class PyDBDaemonThread(threading.Thread):
@@ -161,8 +148,8 @@ class PyDBDaemonThread(threading.Thread):
                 self._stop_trace()
                 self._on_run()
             except:
-                if sys is not None and traceback is not None:
-                    traceback.print_exc()
+                if sys is not None and pydev_log_exception is not None:
+                    pydev_log_exception()
         finally:
             del created_pydb_daemon[self]
 
@@ -296,7 +283,7 @@ class ReaderThread(PyDBDaemonThread):
                             line = line[:-1]
                 except:
                     if not self.killReceived:
-                        traceback.print_exc()
+                        pydev_log_exception()
                         self.handle_except()
                     return  # Finished communication.
 
@@ -307,8 +294,7 @@ class ReaderThread(PyDBDaemonThread):
                     line = line.decode('utf-8')
 
                 if DebugInfoHolder.DEBUG_RECORD_SOCKET_READS:
-                    sys.stderr.write(u'debugger: received >>%s<<\n' % (line,))
-                    sys.stderr.flush()
+                    pydev_log.critical(u'debugger: received >>%s<<\n' % (line,))
 
                 args = line.split(u'\t', 2)
                 try:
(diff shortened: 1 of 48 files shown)


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -1118,9 +1118,8 @@ def build_exception_info_response(dbg, thread_id, request_seq, set_additional_th
 
                     # Never filter out plugin frames!
                     if not getattr(frame, 'IS_PLUGIN_FRAME', False):
-                        if not dbg.in_project_scope(original_filename):
-                            if not dbg.get_use_libraries_filter():
-                                continue
+                        if dbg.is_files_filter_enabled and dbg.apply_files_filter(frame, original_filename, False):
```
fabioz: Note that the changes in test_exceptions_and_partial_exclude_rules are due to this change (the apply_files_filter handles both include/exclude rules as well as the project scope).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
```diff
@@ -191,9 +204,10 @@ def make_get_thread_stack_message(self, py_db, seq, thread_id, topmost_frame, fm
 
                     presentation_hint = None
                     if not getattr(frame, 'IS_PLUGIN_FRAME', False):  # Never filter out plugin frames!
+                        if py_db.is_files_filter_enabled and py_db.apply_files_filter(frame, original_filename, False):
```
fabioz: Note that the changes in test_exceptions_and_partial_exclude_rules are due to this change (the apply_files_filter handles both include/exclude rules as well as the project scope).

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -3,10 +3,9 @@
 
 from _pydevd_bundle._debug_adapter import pydevd_schema, pydevd_base_schema
 from _pydevd_bundle._debug_adapter.pydevd_base_schema import from_json
-from tests_python.debugger_unittest import IS_JYTHON, REASON_STEP_INTO, REASON_STEP_OVER, \
-    REASON_CAUGHT_EXCEPTION, REASON_STEP_RETURN, IS_APPVEYOR, overrides
-from _pydevd_bundle._debug_adapter.pydevd_schema import ThreadEvent, ModuleEvent, OutputEvent, \
-    ExceptionOptions, Response, StoppedEvent, ContinuedEvent
+from tests_python.debugger_unittest import IS_JYTHON, IS_APPVEYOR, overrides
```
fabioz: Note: I had to rewrite many tests to use only DAP messages (because internally the thread id is still a string and we have a generated thread id when passed through json -- mixing both APIs in the same test is no longer ok because of that).

tests/func/test_exclude_rules.py
```diff
@@ -132,19 +132,19 @@ def call_func():
...
                     'path': Path(os.path.join(call_me_back_dir, 'call_me_back.py'))
                 })
             })
-            assert frames[1] == ANY.dict_with({
-                'line': line_numbers['call_me_back_line'],
-                'source': ANY.dict_with({
-                    'path': Path(code_to_debug)
-                })
-            })
+            # assert frames[1] == ANY.dict_with({ -- filtered out
```
fabioz: Note: left code around to know what's being filtered out.

karthiknadig:
Looks like tests run fine on linux but not on windows/mac.

karthiknadig:
> do we support disconnect/reconnect in multiprocessing scenarios?

We do support reconnect to the parent, but not to the child process. At least the way it works now. In the future, we should cache the connection details for the child processes that are alive, and send those back.

fabioz:
> Looks like tests run fine on linux but not on windows/mac.

I'm taking a look at that.

fabioz:
@karthiknadig this still can't be merged... it seems that Azure devops is not picking my changes and redoing the build (https://ptvsd.visualstudio.com/ptvsd/_build?definitionId=1&_a=summary).

fabioz:
Ok, it's ready now -- after I re-requested the reviews and closed/reopened it seems it started building again ;)

tests/conftest.py
```diff
@@ -15,6 +15,32 @@
...
+    sys.stderr.write('******************************************************************\n')
+
+
+def _on_dump_stack_print_pydevd_log():
+    # On timeout we also want to print the pydev log.
+    _print_pydevd_log('timeout')
+    _original_dump_stacks()
+
+
+pytest_timeout.dump_stacks = _on_dump_stack_print_pydevd_log
```
fabioz: Note that there's an issue open in `pytest-timeout`: https://bitbucket.org/pytest-dev/pytest-timeout/issues/15/new-hook-called-right-before-timeout to support that better (but until that's done, monkeypatching `dump_stacks` seems to be a reasonable workaround).

tests/conftest.py
```diff
@@ -15,6 +15,32 @@
...
 from . import helpers
 from .helpers.printer import wait_for_output
+import pytest_timeout
+import tempfile
+
+_original_dump_stacks = pytest_timeout.dump_stacks
+
+
+def _print_pydevd_log(reason):
+    sys.stderr.write('\n******************************************************************\n')
```
fabioz: Note: with this change, besides the ptvsd log and thread information the pydevd output is also sent (I had to add it to diagnose why some things were not working on the ci).

Issue #1385, opened 2019-04-24
Edit dictionary value hangs VS


huguesv:
## Environment data

- PTVSD version: 4.2.8
- OS and version: Win10
- Python version (& distribution if applicable, e.g. Anaconda): 3.7
- Using VS Code or Visual Studio: VS

## Actual behavior

VS hangs when editing a dict value

## Expected  behavior

Value in dict is changed and watch window shows the new value

Note that it works when reverting to ptvsd 4.2.0

## Steps to reproduce:
1. Enter this code
```
data = {
    'page': 'About',
    'description': 'About this application'
}

print(data)
```

2. Add breakpoint on print
3. Run to breakpoint
4. In locals window, edit the value for 'page' and press enter
5. VS puts up a waiting dialog that never ends and must be canceled

```
 1> [DebugAdapter] --> C (configurationDone-15): {"command":"configurationDone","arguments":{},"seq":15,"type":"request"}
 1> [DebugAdapter] <--   E (process): {"type": "event", "seq": 16, "event": "process", "body": {"name": "C:\\Users\\huvalo\\source\\repos\\editlocaltest\\editlocaltest.py", "systemProcessId": 432, "isLocalProcess": true, "startMethod": "launch"}}
 1> [DebugAdapter] <--   R (configurationDone-15) [63 ms]: {"type": "response", "seq": 17, "request_seq": 15, "success": true, "command": "configurationDone", "message": "", "body": {}}
 1> [DebugAdapter] <--   E (thread): {"type": "event", "seq": 18, "event": "thread", "body": {"reason": "started", "threadId": 1}}
 1> [DebugAdapter] <--   E (stopped): {"type": "event", "seq": 19, "event": "stopped", "body": {"reason": "breakpoint", "threadId": 1, "text": null, "description": null, "preserveFocusHint": false, "allThreadsStopped": true}}
 1> [DebugAdapter] --> C (threads-16): {"command":"threads","arguments":{},"seq":16,"type":"request"}
 1> [DebugAdapter] <--   R (threads-16) [4 ms]: {"type": "response", "seq": 20, "request_seq": 16, "success": true, "command": "threads", "message": "", "body": {"threads": [{"id": 1, "name": "MainThread"}]}}
 1> [DebugAdapter] --> C (stackTrace-17): {"command":"stackTrace","arguments":{"threadId":1,"startFrame":0,"levels":1,"format":{"includeAll":true}},"seq":17,"type":"request"}
[...]

PR #1391, opened 2019-04-30
Fix for #1385 (edit contents of dict/list/set/objects)

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
```diff
@@ -405,6 +405,23 @@ class SetResolver:
         Resolves a set as dict id(object)->object
     '''
 
+    def get_contents_debug_adapter_protocol(self, obj, fmt=None):
+        ret = []
+
+        for i, item in enumerate(obj):
+            ret.append((str(id(item)), item, None))
+
+            if i > MAX_ITEMS_TO_HANDLE:
```
karthiknadig: This should probably use, `start` and `count` from variables request. We should have a separate task to handle that.

fabioz: Created: https://github.com/Microsoft/ptvsd/issues/1392 to handle that.

PR #1395, opened 2019-05-01
Minor improvements to troubleshooting instructions

karthiknadig:


TROUBLESHOOTING.md
```diff
@@ -7,9 +7,30 @@ in [Filing an issue](#filing-an-issue).
 ## Known Issues
 
 There are a few known issues in the current version of the debugger:
-- Multiprocess debugging on a Linux machine requires the `spawn` setting. See [#943](https://github.com/Microsoft/ptvsd/issues/943).
-- If you recieve an error saying  `breakpoint not set`, then look at your path mappings. See Meta-Issue [#2976](https://github.com/Microsoft/vscode-python/issues/2976) for more detail. 
-- If you want to debug library files, you have to enable `debugStdLib`. See [#1354](https://github.com/Microsoft/ptvsd/issues/1354).
+### 1. Multiprocessing on Linux/Mac
+ Multiprocess debugging on a Linux machine requires the `spawn` setting. We are working on improving this experience, see [#943](https://github.com/Microsoft/ptvsd/issues/943). Mean while do this to improve your debugging experience:
```
int19h: "Meanwhile"

TROUBLESHOOTING.md
```diff
@@ -7,9 +7,30 @@ in [Filing an issue](#filing-an-issue).
...
+```py
+import multiprocessing
+multiprocessing.set_start_method('spawn', True)
+```
+Note: `spawn` will work cross platform, on windows `multiprocessing` package uses spawn. If you choose to use `spawn` you may have to structure your `__main__` module like this https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods.
```
int19h: Capital W in "Windows", and it needs a semicolon after "this".

I would rephrase the first sentence thus: "On Windows, the `multiprocessing` package uses "spawn" as the default and only option, so it is recommended for cross-platform code to ensure uniform behavior."

Issue #1355, opened 2019-04-15
Make sure that get_file_type is not called until configurationDone is called.


fabioz:
- [ ] Set ignore patterns in handle launch/attach (see comment in wrapper about this issue)
- [ ] Forward setDebuggerProperty request to pydevd
- [ ] Remove `dont_trace_external_files` usage from wrapper.
- [ ] Ensure settings set setDebuggerProperty are applied

fabioz:
It seems that `ptvsd.attach_server.break_into_debugger` calls `get_file_type` to discover which frame it should stop into.

karthiknadig:
Break into debugger should be blocked until configurationDone is called on ptvsd. You are right this is another integration point that will need custom protocol support.

fabioz:
The frame is used as the last thing when attaching, so, I think that accepting a callable for the frame that's computed on demand may be enough here (I'm checking that now).

fabioz:
@karthiknadig I'm actually having many other failures after I integrated my latest changes into the master (because `_forward_request_to_pydevd` is now sending `success=False` on `pydevd` failures which it wasn't doing properly before).

So, I'm still investigating as it's not only `test_breakpoint_function` which seems to be breaking now (so, the change I proposed is not enough -- still investigating).

fabioz:
@karthiknadig investigating more it seems that besides the problem with `get_file_type` being called on `ptvsd` on `break_into_debugger` (for which I've done a fix at: https://github.com/fabioz/ptvsd/commit/2fb2f574f1734441c52f10d033edb531ad88dedd),  `PyDB.ready_to_run` (which is the flag in `pydevd` which says that everything is setup) is being set to True before `configurationDone`.

`get_file_type` is being called during the `setup_connection` (see the stack below) after the `ready_to_run` flag is set to True but before `on_setdebuggerproperty_request` is called -- it happens reliably on tests for me (when `start_method` == `attach_socket_cmdline`).

```
  File "/home/fabioz/miniconda3/envs/py37/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/fabioz/miniconda3/envs/py37/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/fabioz/Desktop/ptvsd/ptvsd/src/ptvsd/__main__.py", line 405, in <module>
    main(sys.argv)
[...]

PR #1404, opened 2019-05-06
Add support to call pydevd.settrace without blocking. Fixes #1248

### LINKED ISSUE #1248: pydevd support for settrace without blocking

fabioz:


int19h:
Wouldn't this still have the same issue that you've described at https://github.com/microsoft/ptvsd/issues/1355#issuecomment-483319024?

fabioz:
> 
> 
> Wouldn't this still have the same issue that you've described at [#1355 (comment)](https://github.com/microsoft/ptvsd/issues/1355#issuecomment-483319024)?

This should no longer be a problem after https://github.com/fabioz/ptvsd/commit/55e3b23266b39c2b80be4f20123e302924fb8631 (see comment: https://github.com/microsoft/ptvsd/issues/1355#issuecomment-488673488)

There are still some ordering issues (for instance, the path mapping must still be sent before breakpoints), but the caches related to `get_file_type` are now properly reset (so, it should no longer be a problem calling it before the debugger is setup).

PR #1405, opened 2019-05-06
Redirect output setting changes

karthiknadig:
* Deprecates redirect output setting via `enable_attach`
* Removed all redirect output setting via `settrace` from `ptvsd`.
* Moved redirect output debug option handling from `ptvsd` to `pydevd`.
* Uses debug option in `pydevd` to set redirect output.
* Moved show return value setting to pydevd

### DIFF
--- a/src/ptvsd/__main__.py
+++ b/src/ptvsd/__main__.py
@@ -136,7 +136,6 @@ switches = [
     ('--wait',                  None,               set_true('wait'),                       False),
     ('--multiprocess',          None,               set_true('multiprocess'),               False),
     ('--log-dir',               '<path>',           set_arg('log_dir', string),             False),
-    ('--no-warn',               None,               set_false('show_deprecation_warning'), False),
 
     # Switches that are used internally by the IDE or ptvsd itself.
     ('--nodebug',               None,               set_nodebug,                            False),
--- a/src/ptvsd/__main__.py
+++ b/src/ptvsd/__main__.py
@@ -107,11 +107,6 @@ def set_true(varname):
         setattr(ptvsd.options, varname, True)
     return do
 
-def set_false(varname):
-    def do(arg, it):
-        setattr(ptvsd.options, varname, False)
-    return do
-
 def set_target(kind, parser=None):
     def do(arg, it):
         ptvsd.options.target_kind = kind
--- a/src/ptvsd/__main__.py
+++ b/src/ptvsd/__main__.py
@@ -107,6 +107,11 @@ def set_true(varname):
         setattr(ptvsd.options, varname, True)
     return do
 
+def set_false(varname):
+    def do(arg, it):
+        setattr(ptvsd.options, varname, False)
+    return do
+
 def set_target(kind, parser=None):
     def do(arg, it):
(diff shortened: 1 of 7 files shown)


src/ptvsd/attach_server.py
```diff
@@ -114,6 +120,12 @@ def attach(address, redirect_output=True, log_dir=None):
     ptvsd.log.to_file()
     ptvsd.log.info('attach{0!r}', (address, redirect_output))
 
+    if redirect_output is not None:
+        if ptvsd.options.show_deprecation_warning:
+            ptvsd.log.warn(_redirect_output_deprecation_msg)
```
karthiknadig: Probably should add `deprecate` to log. and handle this logic in there.

int19h:
Since it's an explicitly invoked API like any other Python library, I think we should rather use `warn()` with `DeprecationWarning` here, since that's the standard mechanism that users expect:

https://docs.python.org/3/library/warnings.htm

src/ptvsd/attach_server.py
```diff
@@ -23,6 +23,10 @@
 
 _pending_threads = set()
 
+_redirect_output_deprecation_msg = (
+    "'redirect_output' setting via enable_attach will be deprecated in the future versions of the debugger. "
+    "This can be set using redirectOutput in Launch config in VS Code, using Tee output option in Visual Studio, "
```
fabioz: Typo: `using Tee`

fabioz: The message says it will be deprecated in a future version, but I think it's already unused in this commit, no?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -280,6 +280,13 @@ def _set_debug_options(self, py_db, args):
         if bool(path_mappings):
             pydevd_file_utils.setup_client_server_paths(path_mappings)
 
+        if self._debug_options.get('REDIRECT_OUTPUT', False):
```
fabioz: I think it'd be nice to have unit-tests with and without output redirect in `ptvsd`. I searched for `redirect` in the ptvsd tests and only found one reference in the `DebugSession` constructor (so, it seems we only have tests which always redirect on `ptvsd`).

src/ptvsd/__main__.py
```diff
@@ -107,6 +107,11 @@ def do(arg, it):
         setattr(ptvsd.options, varname, True)
     return do
 
+def set_false(varname):
```
fabioz: This seems unused in this pull request.

src/ptvsd/attach_server.py
karthiknadig: The `using tee` part is talking about this setting in Visual Studio:
image

About the deprecation text. Even though user could pass the redirection here, VS and VSC later also set it, always. so no matter what the user choice was here, it would always get overridden by the setting chosen in the IDE. Effectively having the same result as it being unused. Eventually we intend to remove this argument entirely.

Issue #1413, opened 2019-05-09
Stop on entry hangs the debugger


karthiknadig:
## Environment data

- PTVSD version: master
- OS and version: windows
- Python version (& distribution if applicable, e.g. Anaconda): 3.6
- Using VS Code or Visual Studio: VS or VSC

## Actual behavior

Note that you have to change the default for STOP_ON_ENTRY in wrapper.py#1198 to True.
Repros almost always on VS, but infrequently on VSC.

## Expected  behavior

Should run and break on entry

## Steps to reproduce:
For VS:
1. Change wrapper.py:1198 default for STOP_ON_ENTRY  to True.
2. Use F5 to start the debugger.

For VSC:
1. Set `stopOnEntry` in launch json.

karthiknadig:
Note that this occurs with or without the changes below. the change below seems to cause this issues more frequently.
https://github.com/karthiknadig/ptvsd/commit/6a80f7c070c9b652a13eeac325b0056aebb8b34e

qubitron:
@anapo14 let's remove this from the VS Code getting started tutorial since it seems to break from time to time

fabioz:
@karthiknadig are you investigating this? I couldn't reproduce here.

My guess is that it's a racing condition -- it'd be nice to have the `ptvsd` as well as the `pydevd` logs to know what's happening when it fails...

To get the `pydevd` logs, it's possible to set environment variables (not sure about the last release, but it does work in the development version).

```
PYDEVD_DEBUG=True
PYDEVD_DEBUG_FILE=c:\temp\file_to_log.txt
```

karthiknadig:
[ptvsd-7560.log](https://github.com/microsoft/ptvsd/files/3163421/ptvsd-7560.log)
[pydevd_log.txt](https://github.com/microsoft/ptvsd/files/3163422/pydevd_log.txt)
[VS_adapter_host_logs.txt](https://github.com/microsoft/ptvsd/files/3163423/VS_adapter_host_logs.txt)

karthiknadig:
This is a hard repro. Today, i am not able to repro it in VSC but only in VS.

fabioz:
@karthiknadig I still can't reproduce it (tried running the related tests thousands of times in the ci and it didn't fail either).

If you can still reproduce it there, can you check some things?

[...]

PR #1415, opened 2019-05-09
Ensure stop on entry reason is `entry`. Fixes #1387

### LINKED ISSUE #1387: Stop on entry should return stopped reason as 'entry'
This is needed to support F5/F11 scenarios in VS.

### LINKED ISSUE #1418: DAP: Stop on entry should be ported to pydevd
Blocked by:

https://github.com/microsoft/ptvsd/issues/1387
https://github.com/microsoft/ptvsd/issues/1413

karthiknadig:
Fixes #1387
Fixes #1418


fabioz:
To update `PyDBAdditionalThreadInfo` you should change only `pydevd_cython.pxd` and the `__slots__` / constructor initialization in `pydevd_additional_thread_info_regular.py`, then, run the command below to update the related cython files (updating the path to `X:\ptvsd_workspace` to match your installation).

```
cd /D X:\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd
set PYTHONPATH=X:\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd
C:\bin\Miniconda\envs\py36_64\python build_tools\build.py
```

Note that you should also commit the `.c` files.

karthiknadig:
With VS 2019 the existing build.py fails with can't find `vcvarsall` I temporarily changed that to use the 2019 version to build the .c files.

fabioz:
The build is failing because it's missing `pydevd_stop_on_entry` in `pydevd_cython.pxd`.

karthiknadig:
Not sure why it is not picking up the newly generated .c files. Is there any issue in the generated files?

karthiknadig:
Ah... I see the commit was missing a file. Updated it.

karthiknadig:
This is ready for review.  #1413 is due to `debug` entry point used by VS. This can be addressed separately. the workaround is to turn of stop on entry and use breakpoint instead (in VS). This issue exists in the shipped product. so we can check this change in.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -554,3 +555,8 @@ def custom_dont_trace_external_files(abs_path):
         if reset_caches:
             py_db.clear_dont_trace_start_end_patterns_caches()
 
+    def stop_on_entry(self):
+        main_thread = pydevd_utils.get_main_thread()
```
fabioz: I think it'd be nice to check if `main_thread` is not `None` here (and if it is show a `pydev_log.critical()` to the user) just to be on the safe side.

PR #1416, opened 2019-05-10
[DO NOT MERGE]

fabioz:
This is just making a stop on entry test run thousands of times on the ci to see if hang is reproduced.


fabioz:
Closing this pull request (I've set to run 1000 times for each scenario, the stop on entry and it ran on all platforms without any failure and I can't reproduce it locally either).

PR #1420, opened 2019-05-14
Send DAP process event from pydevd. Fixes #1338

### LINKED ISSUE #1338: DAP: process creation event should come from pydevd

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -292,23 +299,39 @@ def _set_debug_options(self, py_db, args, start_reason):
...
+            name=name,
+            systemProcessId=os.getpid(),
+            isLocalProcess=True,
+            startMethod=start_method,
+        )
+        event = ProcessEvent(body)
+        py_db.writer.add_command(NetCommand(CMD_PROCESS_EVENT, 0, event, is_json=True))
+
+    def _handle_launch_or_attach_request(self, py_db, request, start_reason):
+        self._send_process_event(py_db, start_reason)
```
karthiknadig: This location might be a issue. I think we should be sending this from config done. I vaguely remember that in the early days of debugger we ran into a bunch of issues with VSC UI get into a bad state because of this. @int19h do you remember why we did this from specifically configuration done?

fabioz: I put it here because it seemed to make more sense to me (i.e.: `launch` is the place that creates the process), and because at `wrapper.py` it may be sent either from `configurationDone` or when it receives the first `on_pydevd_thread_create`, which seemed a bit odd (so, I thought this was mostly because of historical reasons and should be fine in the current state of the debugger).

Anyways, I've moved it back to `configurationDone` (and now I'm properly forwarding `configurationDone` from `ptvsd` to `pydevd`).

fabioz: @karthiknadig @int19h 

Actually, I reverted that change because it broke tests... it seems that in practice `ptvsd` was not sending that notification from `configurationDone` but rather the first time that a thread was created (i.e.: from `on_pydevd_thread_create` in the wrapper).

The racing condition from tests is that if the `process` is sent from `configurationDone`, on tests without breakpoints, it's possible that the program finishes before the `process response` is sent back (sending back that response is done in a separate thread and the program will start running right away).

Given that, my guess is that the issues were due to the threads handling/mapping which shouldn't be an issue in the current implementation (in which case sending the `process` from `launch` or `attach` should work?) -- it'd be nice to discover the actual reason to know what we're working around (I did some -- non-extensive -- manual testing on VSCode and things seemed to work for me).

karthiknadig: I think we should do a bug bash after this is in to ensure that this is not going to regress some of the old issues.

int19h: I don't think there was any specific reason. I was probably just taking "debugger has begun debugging a new process" in a very literal way, and `configurationDone` is when we start running user code.

Tests failing is not unexpected. We had a bunch of issues around lack of ordering guarantees for various things in the protocol spec. I tried to excise them as much as possible by waiting for A-or-B in situations where the order between A and B is not guaranteed. But we certainly have places where it still depends on what is basically an implementation detail. In most cases, it should be okay to just fix those tests, so long as VSC doesn't mind the new order.

PR #1436, opened 2019-05-20
No longer switch off getpass on import. Fixes# 988

fabioz:
Note: `pydevd` historically always activated the `getpass` fallback -- this is needed because `eclipse` does not have a real terminal, just a debug console. 

Now, this was already solved in another way in `pydevd` by providing a custom `sitecustomize` which does that, so, just removing that customization should be ok (and I tested it locally on the vscode terminal and it seems to work for me, but it'd be nice to have some more manual testing on this since this is very client dependent -- also, I'm not sure about how it works in VS, so, that should be checked too).


src/ptvsd/_vendored/pydevd/pydevconsole.py
```diff
@@ -185,36 +184,43 @@ def init_mpl_in_console(interpreter):
...
+    if not hasattr(os, 'kill'):  # Jython may not have it.
 
-        import errno
-        if pid == 0:
-            # According to "man 2 kill" PID 0 has a special meaning:
-            # it refers to <<every process in the process group of the
-            # calling process>> so we don't want to go any further.
-            # If we get here it means this UNIX platform *does* have
-            # a process with id 0.
+        def pid_exists(pid):
```
fabioz: This is a fix in the interactive console with jython (which doesn't appear to have `os.kill`). This is unused in `ptvsd`, so, just synchronizing the code here.

PR #1454, opened 2019-05-20
Fix #1413: Stop on entry hangs the debugger

### LINKED ISSUE #1413: Stop on entry hangs the debugger

int19h:
Reimplement `ptvsd.debugger.debug()` VS entry point via `ptvsd.__main__`.


src/ptvsd/debugger.py
```diff
@@ -2,50 +2,44 @@
...
 
-    ptvsd.log.to_file()
-    ptvsd.log.info('run{0!r}', (filename, port_num, run_as))
+    run()
 
-    address = (LOCALHOST, port_num)
-    run_main(address, filename, run_as, *args, **kwargs)
+# A legacy entrypoint for Visual Studio, to allow older versions to work with new ptvsd.
+# All new code should use the entrypoints in __main__ directly.
+#def run(filename, port_num, run_as):
```
karthiknadig: Should we just delete this?

int19h: I think it's not needed - can't find anywhere VS uses it. But I wanted to have a commit with the new version of it, so that if we find out later it was needed, it can be easily restored.

PR #1462, opened 2019-05-23
Properly validate breakpoints. Fixes #1191

### LINKED ISSUE #1191: Break-point validation should fail if justMyCode is true and break-point is set in std lib

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command.py
```diff
@@ -257,9 +257,20 @@ def cmd_set_break(self, py_db, cmd_id, seq, text):
...
-        self.api.add_breakpoint(
+        error_code = self.api.add_breakpoint(
             py_db, filename, btype, breakpoint_id, line, condition, func_name, expression, suspend_policy, hit_condition, is_logpoint)
 
+        if error_code:
+            if error_code == self.api.ADD_BREAKPOINT_FILE_NOT_FOUND:
+                pydev_log.critical('pydev debugger: warning: Trying to add breakpoint to file that does not exist: %s (will have no effect).' % (filename,))
+
+            elif error_code == self.api.ADD_BREAKPOINT_FILE_EXCLUDED_BY_FILTERS:
+                pydev_log.critical('pydev debugger: warning: Trying to add breakpoint to file that is excluded by filters: %s (will have no effect).' % (filename,))
```
karthiknadig: I think if we can say what filter is excluding this file, it might be helpful. Especially when `rules` starts to become a thing. This is not blocking, but if we can add that it would really be helpful.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -216,7 +216,7 @@ def wait_pydb_threads_to_finish(self, timeout=0.5):
         while time.time() < started_at + timeout:
             if len(pydb_daemon_threads) == 1 and pydb_daemon_threads.get(self, None):
                 return
-            time.sleep(0.01)
+            time.sleep(1 / 30.)
```
MinKPark: how about adding a comment on what the delay is change from 0.01 to ~0.03? it would be helpful for us to understand the rational behind this waiting time.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -806,10 +807,12 @@ def apply_files_filter(self, frame, filename, force_check_project_scope):
         try:
             return self._apply_filter_cache[cache_key]
         except KeyError:
+            debug_trace_level = DebugInfoHolder.DEBUG_TRACE_LEVEL
             if self.plugin is not None and (self.has_plugin_line_breaks or self.has_plugin_exception_breaks):
                 # If it's explicitly needed by some plugin, we can't skip it.
                 if not self.plugin.can_skip(self, frame):
-                    # print('include (include by plugins): %s' % filename)
+                    if debug_trace_level >= 3:
```
MinKPark: it looks like that this pattern(if level ...) is repeated, if then, couldn't we put this as a method?

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -818,21 +821,36 @@ def apply_files_filter(self, frame, filename, force_check_project_scope):
...
                         # ignore files matching stepping filters
-                        # print('exclude (filtered out): %s' % filename)
+                        if debug_trace_level >= 3:
+                            pydev_log.error_once('File not traced (excluded by filters): %s', filename)
+
                         self._apply_filter_cache[cache_key] = True
                         return True
                     else:
                         # print('include (explicitly included): %s' % filename)
+                        if debug_trace_level >= 3:
```
MinKPark: how about removing it?

src/ptvsd/_vendored/pydevd/pydevd.py
fabioz: That's really minor... my rationale is that 30 times/second would be ok and already 'real time' as far as the user is concerned -- if we don't have the finish condition we have to wait for other threads so, trying 100 times/second seemed a bit too much for me (I didn't add a comment because that seems very minor for me -- trying 100 times/second would not be terribly bad either.

src/ptvsd/_vendored/pydevd/pydevd.py
fabioz: Done

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_filtering.py
```diff
@@ -127,22 +127,23 @@ def __init__(self):
...
-            # list of excludes.
-            filters = pydevd_filters.split(';')
-            pydev_log.debug("PYDEVD_FILTERS %s\n" % filters)
-            new_filters = []
-            for new_filter in filters:
-                if new_filter.strip():
-                    new_filters.append(ExcludeFilter(new_filter.strip(), True, True))
-            self._exclude_filters = new_filters
+        if pydevd_filters:
+            pydev_log.debug("PYDEVD_FILTERS %s\n" % (pydevd_filters,))
```
MinKPark: why not doing debug level based logging here? I'm curious.

src/ptvsd/_vendored/pydevd/pydevd.py
fabioz: You mean remove the logging?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_filtering.py
fabioz: Having `PYDEVD_FILTERS` as an environment variable is very uncommon, so, it doesn't seem useful to debug it in that case (nothing to see).

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command.py
fabioz: I'll take a look at that.

fabioz: I started to look and it'd change more than I'd like for this pull request, so, I created https://github.com/microsoft/ptvsd/issues/1464 to address this.

src/ptvsd/_vendored/pydevd/pydevd.py
MinKPark: no. I meant the comment above('# print...')

fabioz: Done

PR #1468, opened 2019-05-27
Show message when stepping into library code with justMyCode=true. Fixes #1298

### LINKED ISSUE #1298: Show message when stepping into library code with justMyCode: true

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
```diff
@@ -348,3 +348,13 @@ def make_thread_suspend_message(self, *args, **kwargs):
...
 
+    @overrides(NetCommandFactory.make_skipped_step_in_because_of_filters)
+    def make_skipped_step_in_because_of_filters(self, py_db, frame):
+        msg = 'Frame skipped from debugging during step-in.'
+        if py_db.get_use_libraries_filter():
+            msg += '\nNote: may have been skipped because of "justMyCode" option (default == true).'
+
+        body = OutputEventBody(msg, category='console')
+        event = OutputEvent(body)
+        return NetCommand(CMD_WRITE_TO_CONSOLE, 0, event, is_json=True)
```
fabioz: Note: providing an `OutputEvent` here (so, it'll appear in the debug console). I'm not sure if this is the expected way to show this warning, so, waiting on input here.

karthiknadig: @fabioz As far as I know, sending a message is the best we can do for now. One thing to check would be. What happens if you send a message in the optional message field for the `breakpoint`:
https://microsoft.github.io/debug-adapter-protocol/specification#Types_Breakpoint

@DonJayamanne What do you think? This is the message we show when validating breakpoints and if a break point will be skipped due to JMC.  Ideally we should pop up a message in the UI. But as far as I know we don't have a way to that right now.

fabioz: @karthiknadig in this case it's not evaluating a breakpoint (that's the previous request, which is already integrated: https://github.com/microsoft/ptvsd/commit/016b557b88ba89ebedf8176993866e4a518ee5c9#diff-366f52a4aacb1f6078a339b769738928R539 -- in that case I use the message in the breakpoint and it works well... it shows the breakpoint grayed and if the user hovers over it that message is shown).

For this use-case, the user does a step-in and the filters kick in, in which case we (currently) show a message in the console (which I'm not sure is ideal... a pop up which he can silence would probably be better).

PR #1504, opened 2019-06-07
Fix #1502: Concurrent DAP request processing

### LINKED ISSUE #1502: Concurrent DAP request processing
The message handling code in `messaging.py` can only handle incoming DAP requests serially as designed - the problem is that the handler is invoked directly on the event loop, and its return value is the response body. Any requests that come in the meantime are queued for later processing.

This is not a problem for most requests, because their semantics or their implementation requires serial execution anyway. However, the `launch` and `attach` DAP requests are special (https://github.com/microsoft/vscode/issues/4902#issuecomment-368583522), and must allow for `setBreakpoints` and other pre-`configurationDone` requests to be handled before the `launch` request itself can respond.

A more generic solution to this problem would be to utilize futures/async, but our existing implementation (which we need for 2.7) is ad-hoc, handles errors poorly, and makes for code that is difficult to debug. It would also require rewriting the messaging code entirely. 

A more narrow solution is to simply allow the message handler to "pump", i.e. yield to the event loop so that it can handle one more message. Thus, when `launch` reaches the point at which it must handle further requests, it can simply pump in a loop until a flag indicating that `configurationDone` was received and handled is set. 

API: `JsonMessageChannel` shall allow message handlers to be generator functions. If the handler returns a non-generator, it is handled as response body same as today, and is expected to handle the vast majority of cases. But if it returns a generator, then the event loop for the channel will iterate through it until it sees a non-`None` value, and for every iteration, it will read and process another message (i.e. invoke handler and send the response). Yielding a non-`None` value is the same as return from a non-generator handler - the handler terminates, and yielded value is the response to the request. So, for `launch`, it would look something like:
```py
def configurationDone_request(self, request):
  ...
  self.is_configuration_done = False

def launch_request(self, request):
  # Parse arguments, set things up
  ...

  while not self.is_configuration_done:
    yield

  # Launch process
  ...
  yield {}
```

int19h:
Add support for "yield" in request handlers.

### DIFF
--- a/src/ptvsd/common/messaging.py
+++ b/src/ptvsd/common/messaging.py
@@ -5,6 +5,7 @@
 from __future__ import print_function, with_statement, absolute_import
 
 import contextlib
+import inspect
 import itertools
 import json
 import sys
@@ -156,7 +157,6 @@ class JsonIOStream(object):
         ptvsd.common.log.debug('{0} <-- {1!j}', self.name, value)
 
 
-
 class Request(object):
     """Represents an incoming or an outgoing request.
 
@@ -171,6 +171,7 @@ class Request(object):
         self.seq = seq
         self.command = command
         self.arguments = arguments
+        self.response = None
 
 
 class OutgoingRequest(Request):
@@ -180,7 +181,6 @@ class OutgoingRequest(Request):
 
     def __init__(self, *args):
         super(OutgoingRequest, self).__init__(*args)
-        self.response = None
         self._lock = threading.Lock()
         self._got_response = threading.Event()
         self._callback = lambda _: None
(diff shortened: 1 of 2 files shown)


src/ptvsd/common/messaging.py
```diff
@@ -330,7 +331,7 @@ def _send_message(self, type, rest={}):
 
     def send_request(self, command, arguments=None):
         d = {'command': command}
-        if arguments is not None:
+        if arguments is not None and arguments != {}:
```
karthiknadig: we could use `bool(arguments)` here right?

int19h: It wouldn't need to be `bool()` then, just `arguments` is the same.

It breaks because it also treats 0 and "" as false. In practice all real messages have dicts as arguments, and it would be fine. But the protocol doesn't require it - it says `any?` - and a bunch of pre-existing tests for messaging code uses integers for  it, including 0, so changing it here requires changing a lot of tests also.

PR #1513, opened 2019-06-12
Stepping in multi-threaded case should continue other threads. Fixes #1372

### LINKED ISSUE #1372: Stepping in multi-threaded case should continue other threads
Current behavior: Stepping steps current thread but keeps the other threads paused. This should be optional behavior, enabled via a setting (temporarily). We plan on deprecating this behavior.

Expected Default behavior: Stepping steps in the current thread, and it should continue(resume) other threads. At the end of the steeping action in the current thread all threads are paused. This is the default behavior for C#, and expected and only behavior supported by VS.

Note that Pause, breakpoint, exception, etc should pause all threads.

fabioz:


src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -478,6 +478,9 @@ def __init__(self, set_as_global=True):
         # If True, pydevd will send a single notification when all threads are suspended/resumed.
         self._threads_suspended_single_notification = ThreadsSuspendedSingleNotification(self)
 
+        # If True a step command will do a step in one thread and will also resume all other threads.
+        self.stepping_resumes_all_threads = False
```
karthiknadig: The default behavior should be to resume all other threads. So this should be `True`.

fabioz: The default with the DAP is resuming all other threads, this is the default when the DAP is not being used.

See: https://github.com/microsoft/ptvsd/pull/1513/files#diff-366f52a4aacb1f6078a339b769738928R262

And the test which checks that behavior: https://github.com/microsoft/ptvsd/pull/1513/files#diff-13f63937502db168852f653a7a034e28R128

PR #1519, opened 2019-06-13
 Move more code to pydevd

karthiknadig:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -885,7 +885,7 @@ def _evaluate_response(py_db, request, result, error_message=''):
         variables_response = pydevd_base_schema.build_response(request, kwargs={'body':body})
         py_db.writer.add_command(NetCommand(CMD_RETURN, 0, variables_response, is_json=True))
     else:
-        body = pydevd_schema.EvaluateResponseBody(result='', variablesReference=0)
+        body = pydevd_schema.EvaluateResponseBody(result=result, variablesReference=0)
```
fabioz: Actually, reviewing this now it seems that `result` can be any Python object if `py_db.suspended_frames_manager.get_frame_tracker(thread_id)` returns `None`, in `internal_evaluate_expression_json`, which is unlikely but not impossible on a racing condition. 

So, in that condition, when calling `_evaluate_response`, the result should be `result=''` and not the actual result from the evaluation.

i.e.: the line that calls this method after `if frame_tracker is None:` should be:

`_evaluate_response(py_db, request, result='', error_message='Thread id: %s is not current thread id.' % (thread_id,))`

PR #1524, opened 2019-06-18
Refactor messaging

int19h:
Refactor messaging:
- better propagation of messages and errors
- implicit error handling for invalid messages
- fix various issues related to disconnecting

Fix messaging tests.

Separate formatter from logging for reuse.

Add more Python 2/3 compatibility helpers.


int19h:
Bulk of the changes is in src/ptvsd/common/messaging.py (which is collapsed by default in diff, because of how large it is).

int19h:
I'm going to make some further minor changes to `MessageHandlingError` to make `cause` optional.

Its purpose is to prevent `MessageHandlingError` from bubbling up unexpectedly from some request to a server that an adapter handler might make that can fail, making the request to adapter fail as well. For cases where it's intentional, `delegate` takes care of it either way. But for explicit `wait_for_response()` calls, it is supposed to fast-fail instead (just like any other unhandled exception). 

However, this only needs `cause` to be set for exceptions that come via responses. Those that are raised explicitly inside handlers and code they invoke do not really need `cause`, and requiring it makes things more complicated with the need for `get_server` etc. Making it optional allows to revert those changes in messages.py and keep things simpler.

PR #1537, opened 2019-06-24
Ensure port is set on enable_attach

karthiknadig:
Users should be able to call enable attach with port `0` and have a way to get the port address.
```console
> import ptvsd
> ptvsd.enable_attach(('localhost', 0))
('loclahost', 58940)
```


src/ptvsd/attach_server.py
```diff
@@ -94,6 +94,7 @@ def enable_attach(address=(DEFAULT_HOST, DEFAULT_PORT), redirect_output=None, lo
     address = (address[0], port if type(port) is int else int(port))
 
     ptvsd_enable_attach(address)
+    return (address[0], ptvsd.options.port)
```
int19h: Can you also add the explanation of the return value to the docstring?

PR #1538, opened 2019-06-25
[DO NOT  MERGE] Debug options handling

karthiknadig:


src/ptvsd/adapter/messages.py
```diff
@@ -135,8 +137,51 @@ def initialize_request(self, request):
...
+        self.debug_options = dbg_options.extract_debug_options(
+            args.get("options"), args.get("debugOptions")
+        )
+
+        client_os_type = self.debug_options.get("CLIENT_OS_TYPE", "").upper().strip()
+        if bool(client_os_type) and client_os_type not in ("WINDOWS", "UNIX"):
+            ptvsd.log.warn(
+                'Invalid CLIENT_OS_TYPE passed: %s (must be either "WINDOWS" or "UNIX").'
+                % (client_os_type,)
+            )
```
int19h: Just `log.warn`, since it's imported via `from`.

Also, `warn` does string formatting by itself - its first argument is a format string, of the `"{0}"` variety.

src/ptvsd/common/dbg_options.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+def bool_parser(s):
+    return s in ("True", "true", "1")
+
+def str_parser(s):
```
int19h: Dupe of `compat.maybe_utf8`.

src/ptvsd/adapter/messages.py
```diff
@@ -135,8 +137,51 @@ def initialize_request(self, request):
...
+                    # dontTraceStartPatterns=[PTVSD_DIR_PATH],
+                    # dontTraceEndPatterns=['ptvsd_launcher.py'],
+                    "skipSuspendOnBreakpointException": ("BaseException",),
+                    "skipPrintBreakpointException": ("NameError",),
+                    "multiThreadsSingleNotification": True,
+                    "ideOS": client_os_type,
+                },
+            ).wait_for_response()
+        except messaging.MessageHandlingError as exc:
+            request.cant_handle(
```
int19h: These also provide integrated formatting:
```py
request.cant_handle("...: {0}", exc)
```

src/ptvsd/server/_remote.py
```diff
@@ -13,11 +14,17 @@
...
 from ptvsd.server.pydevd_hooks import install
 from ptvsd.server.daemon import session_not_bound, DaemonClosedError
+from pydevd_file_utils import get_abs_path_real_path_and_base_from_file
 
 
 global_next_session = lambda: None
 
 
+def _get_debugger_root():
+    ptvsd_abspath = os.path.abspath(get_abs_path_real_path_and_base_from_file(ptvsd.__file__)[0])
```
int19h: If we consider `set_dont_trace_start_end_patterns` as a public pydevd API, then I think it's best if it does the `NormPath` stuff internally - ptvsd should just be able to do `os.path.abspath(ptvsd.__file__)` here, and call it a day. 

There's one other catch, though. Look at the comment here - it also applies to this case, since user code might do a bunch of things before calling `enable_attach`, including chdir.

https://github.com/microsoft/ptvsd/blob/b29c459ff844709e455d2b3c39068244d58f9a7d/src/ptvsd/server/__main__.py#L46-L54

I'm thinking that perhaps it's best to do the same thing for ptvsd/__init__.py as well? Then everything can just use `ptvsd.__file__`, and it's always guaranteed to be absolute etc.

Or we can make some helper methods there, like `ptvsd.package_dir()`. But I don't really see the downside of just forcing `__file__` to be absolute - and then it's impossible to make a mistake. Coincidentally, it seems to be always absolute in py3, so this is just a py2 compat thing.

PR #1544, opened 2019-06-28
Test refactoring migrated to new api

karthiknadig:


tests/ptvsd/server/test_django.py
```diff
@@ -117,13 +118,13 @@ def test_django_template_exception_no_multiproc(start_method):
                 arguments={'threadId': hit.thread_id, }
             ).wait_for_response()
             exception = resp_exception_info.body
-            assert exception == ANY.dict_with({
-                'exceptionId': ANY.such_that(lambda s: s.endswith('TemplateSyntaxError')),
+            assert exception == some.dict.containing({
+                'exceptionId': some.such_that(lambda s: s.endswith('TemplateSyntaxError')),
```
int19h: `such_that` requires a base pattern, e.g.: `some.thing.such_that(...)`, or better yet `some.str.such_that(...)` in this case - for better error messages.

This was also true before, it's just that `ANY` was the equivalent of `some.thing`.

I forgot to remove the corresponding top-level member in `tests/patterns/__init__.py` - can you kill it there also?

PR #1545, opened 2019-06-29
Update Debug Adapter Protocol links

SamB:
There's a [*New home for the Debug Adapter Protocol*][1], so we might as well link to it.

[1]: https://code.visualstudio.com/blogs/2018/08/07/debug-adapter-protocol-website

P.S. Wow, they were not kidding about the old home being an obscure github repository: https://github.com/Microsoft/vscode-debugadapter-node manages to sound not only vscode-specific, but also like a Node-specific secondary implementation.  And who would suspect that the JSON schema is actually the source of truth?  I guess you guys knew, but I had no clue until I clicked through to Microsoft/vscode#19636 from the blog post.


int19h:
Thanks!

PR #1546, opened 2019-07-01
Custom pydevd event to support InputRequested

### LINKED ISSUE #1496: OOP Debug Adapter: pydevd_module_event, pydevd_input_requested, pydevd_thread_create, pydevd_thread_kill, pydevd_process_event

### LINKED ISSUE #1497: OOP Debug Adapter: pydevd_thread_suspend_single_notification, pydevd_thread_resume_single_notification, pydevd_cmd_write_to_console2

### LINKED ISSUE #1523: Handing events from pydevd that use `pydevd_cmd_id` in json
These are events that are handled via `pydevd_events.handler` in `wrapper.py`. the json for these events has `pydevd_cmd_id` in addition to event name. In some cases where there isn't a DAP equivalent there may not be a event name.

karthiknadig:
This Fixes #1523 in a way that it adds a custom event so that there is always a name for the event.

This also addresses Fixes #1497, Fixes #1496. Those events don't need additional work, and they can be forwarded in their current state.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/_debug_adapter/pydevd_schema.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/_debug_adapter/pydevd_schema.py
@@ -12490,6 +12490,79 @@ class SetDebuggerPropertyResponse(BaseSchema):
         return dct
 
 
+@register_event('pydevdInputRequested')
+@register
+class PydevdInputRequestedEvent(BaseSchema):
+    """
+    The event indicates input was requested by debuggee.
+
+    Note: automatically generated code. Do not edit manually.
+    """
+
+    __props__ = {
+        "seq": {
+            "type": "integer",
+            "description": "Sequence number."
+        },
+        "type": {
+            "type": "string",
+            "enum": [
+                "event"
+            ]
+        },
+        "event": {
+            "type": "string",
+            "enum": [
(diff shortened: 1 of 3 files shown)


src/ptvsd/_vendored/pydevd/_pydevd_bundle/_debug_adapter/pydevd_schema.py
```diff
@@ -12490,6 +12490,79 @@ def to_dict(self, update_ids_to_dap=False):  # noqa (update_ids_to_dap may be un
...
+        :param ['array', 'boolean', 'integer', 'null', 'number', 'object', 'string'] body: Event-specific information.
+        """
+        self.type = 'event'
+        self.event = 'pydevdInputRequested'
+        self.seq = seq
+        self.body = body
+        self.kwargs = kwargs
+
+
+    def to_dict(self, update_ids_to_dap=False):  # noqa (update_ids_to_dap may be unused)
```
int19h: I think you can use `_` as variable name to suppress the unused warning without having to write the comment.

fabioz: This isn't really possible because it wouldn't follow what's expected from the API (callers expect the parameter to be `update_ids_to_dap`).

PR #1550, opened 2019-07-01
Replace wait_for_thread_stopped with wait_for_stop

karthiknadig:


tests/ptvsd/server/test_attach.py
```diff
@@ -44,18 +44,16 @@ def test_attach(run_as, wait_for_attach, is_attached, break_into):
 
         if break_into == "break":
             assert session.read_json() == "break_into_debugger"
-            hit = session.wait_for_thread_stopped()
-            frames = hit.stacktrace.body["stackFrames"]
-            assert 32 == frames[0]["line"]
+            hit = session.wait_for_stop()
+            assert 32 == hit.frames[0]["line"]
```
int19h: Can you switch this one to use `#@bp` also? It should work in named files just the same, but you'll have to call `tests.code.get_marked_line_numbers` on it explicitly.

karthiknadig: Missed this one. I had changed others in a previous PR

tests/ptvsd/server/test_flask.py
```diff
@@ -65,63 +65,56 @@ def test_flask_breakpoint_no_multiproc(bp_target, start_method):
...
+            scopes = resp_scopes.body["scopes"]
+            assert len(scopes) > 0
 
-        web_content = web_request.wait_for_response()
-        assert web_content.find(bp_var_content) != -1
+            resp_variables = session.send_request(
+                "variables",
+                arguments={"variablesReference": scopes[0]["variablesReference"]},
+            ).wait_for_response()
+            variables = list(
```
int19h: `[v for v ...]`

tests/ptvsd/server/test_flask.py
```diff
@@ -65,63 +65,56 @@ def test_flask_breakpoint_no_multiproc(bp_target, start_method):
...
+        with flask_server:
+            home_request = flask_server.get("/")
 
-        thread_stopped = session.wait_for_next(
-            Event("stopped"), some.dict.containing({"reason": "breakpoint"})
-        )
-        assert thread_stopped.body["threadId"] is not None
-
-        tid = thread_stopped.body["threadId"]
+            thread_stopped = session.wait_for_next(
```
int19h: `wait_for_stop`

int19h: (and I think the below can be simplified a lot then? since it's doing what `wait_for_stop` would do, like fetching frames etc)

tests/ptvsd/server/test_flask.py
```diff
@@ -139,65 +132,57 @@ def test_flask_template_exception_no_multiproc(start_method):
         session.start_debugging()
 
         # wait for Flask web server to start
```
int19h: The comment is redundant now (since it applied to `wait_for_connection`)

tests/stacks.py
```diff
@@ -9,8 +9,10 @@
 
 import os
 import sys
+import time
 import threading
 import traceback
+import pytest_timeout.dump_stacks
```
int19h: Just `pytest_timeout`.

PR #1557, opened 2019-07-02
Fix #1551: Backchannel failures in tests

### LINKED ISSUE #1551: Backchannel failures in tests
(https://github.com/microsoft/ptvsd/tree/dbg_adapter_refactor)

On Python 2.7, when the test reads from the backchannel, and debuggee writes to the backchannel and immediately exists, the test sometimes gets `EOFError`. This results in sporadic test failures for many tests.

int19h:
Fix various bugs around handling of disconnect in JsonIOStream and JsonMessageChannel.

Fix handling of DAP "terminated" event in debug.Session.

Add --ptvsd-logs and --pydevd-logs switches to pytest.

Improve message logging to fully capture the raw message data in the logs if deserialization fails.

Log all debuggee environment variables in debug.Session, and improve log readability.

### DIFF
--- a/tests/_logs/README
+++ b/tests/_logs/README
@@ -1,0 +1,13 @@
+Pass --ptvsd-logs and/or --pydevd-logs to pytest to create log files here for a run.
+For example:
+
+    tox -e py37 -- --ptvsd-logs "tests/ptvsd/server/test_run.py::test_run[launch-file]"
+
+A separate ptvsd-{pid}.log file will be created for every ptvsd process spawned by
+the tests. However, because process IDs are reused by OS, logs may end up overwriting
+earlier logs.
+
+All pydevd logs go into the same file named pydevd.log - and subsequent tests will
+overwrite earlier logs.
+
+Thus, it is best to use this when running a single test, or a small number of tests.
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ pip-delete-this-directory.txt
 
 # Unit test / coverage reports
 htmlcov/
+tests/_logs/*.log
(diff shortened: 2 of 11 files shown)


tests/_logs/README
```diff
@@ -0,0 +1,13 @@
+Pass --ptvsd-logs and/or --pydevd-logs to pytest to create log files here for a run.
+For example:
+
+    tox -e py37 -- --ptvsd-logs "tests/ptvsd/server/test_run.py::test_run[launch-file]"
+
+A separate ptvsd-{pid}.log file will be created for every ptvsd process spawned by
```
karthiknadig: Should we add a timestamp to the name?

int19h: It would prevent them being overwritten, but make it hard to correlate ptvsd logs with pydevd logs.

I could add the process ID of the test worker, though, and also add it to the test worker log. That way you could at least find which pydevd log corresponds to which failed test.

int19h: Ah no, that doesn't really work... because the same worker will run multiple tests. 

However, there's already existing logic that generates pydevd logs separately for every `debug.Session` in a temp directory - that's used to dump them if the test fails or times out. I can uncomment the line that logs the name of that temp file - then you can see it in test logs even for successful tests, and inspect it there.

For this switch, let's keep it as is. It's mainly for debugging a specific test, anyway. And once we have end-to-end logs, the problem goes away.

karthiknadig: For clarity, I was thinking `ptvsd-{pid}-{timestamp}.log`. We can do that change if it really becomes a issue.

PR #1562, opened 2019-07-03
Add option to connect to adapter via port from IDE

### LINKED ISSUE #1556: ptvsd.adapter should support debugServer in launch.json
ptvsd.adapter package currently only communicates with the IDE over stdin/out, as is usual for DAP. However, this is inconvenient for testing and debugging, since it requires it to be registered as a debugger extension with VSCode.

To simply this, we can use the `debugServer` option that VSCode provides in launch.json, whereby it connects to an already-running debug adapter over the specified port: https://vscode.readthedocs.io/en/latest/editor/debugging/#launchjson-attributes

This requires the adapter to support an incoming socket connection. It does not need to be a true server - it only has to listen for and accept a single connection, at the very beginning, where it currently does this:

https://github.com/microsoft/ptvsd/blob/1e6d2269dd47b22b5128e9ddc6c1c7be047e9a4f/src/ptvsd/adapter/__main__.py#L13-L21

Since it still needs to work in stdin/out mode by default, this should be configurable via a command line switch that accepts the port number.

karthiknadig:
Fixes #1556

You should be able to connect tot the adapter using this CLI:
```
-m ptvsd.adapter --debug-port 6789
```
This launch json config:
```js
        {
            "name": "Attach port",
            "type": "python",
            "request": "attach",
            "debugServer": 6789,
        },
```

### DIFF
--- a/src/ptvsd/adapter/__main__.py
+++ b/src/ptvsd/adapter/__main__.py
@@ -5,9 +5,16 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import sys
+import argparse
 
 # WARNING: ptvsd and submodules must not be imported on top level in this module,
 # and should be imported locally inside main() instead.
+
+
+def _get_parsed_args():
+    parser = argparse.ArgumentParser(prog="ptvsd.adapter")
+    parser.add_argument("--debug-port", nargs="?", default=None, const=6789)
+    return parser.parse_args(sys.argv[1:])
 
 
 def main():
@@ -15,10 +22,11 @@ def main():
     from ptvsd.common import log
     from ptvsd.adapter import channels
 
+    args = _get_parsed_args()
     log.to_file()
 
     chan = channels.Channels()
-    chan.connect_to_ide()
+    chan.connect_to_ide(address=("localhost", int(args.debug_port)))
 
     chan.ide.send_event(
         "output",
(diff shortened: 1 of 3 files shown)


src/ptvsd/adapter/__main__.py
```diff
@@ -5,20 +5,28 @@
...
+import argparse
 
 # WARNING: ptvsd and submodules must not be imported on top level in this module,
 # and should be imported locally inside main() instead.
 
 
+def _get_parsed_args():
+    parser = argparse.ArgumentParser(prog="ptvsd.adapter")
+    parser.add_argument("--debug-port", nargs="?", default=None, const=6789)
+    return parser.parse_args(sys.argv[1:])
```
int19h: I think you can skip `prog=` and omit the argument here - it should take the default `prog=sys.argv[0]` then. So if you do `python -m ptvsd.adapter`, it'll do the same thing.

src/ptvsd/adapter/channels.py
```diff
@@ -28,13 +29,19 @@ class Channels(singleton.ThreadSafeSingleton):
     """
 
     @singleton.autolocked_method
-    def connect_to_ide(self):
+    def connect_to_ide(self, address=("localhost", None)):
```
int19h: Let's do `address=None` instead for stdin/out. It's a bit confusing that the host gets specified, but then goes unused.

PR #1580, opened 2019-07-11
Test fixes and improvements

int19h:
Fix remaining tests to reflect the debug adapter refactoring changes.

Fix Flask and Django multiprocess tests.

Fix test logs not being captured by pytest.

Fix "import debug_me" check improperly applied in tests where it is unnecessary.

Fix some clarifying patterns not respecting the underlying pattern.

Add pattern helpers for strings: starting_with, ending_with, containing.

Move DAP test helpers to a separate module, and add a helper for frames.

Add support for line markers when setting breakpoints and matching frames.

Assorted test fixes around handling of Unicode and paths.


int19h:
There are some sporadic test failures in output tests (including log points), because the output-not-getting-flushed-on-exit bug is still there, apparently. I'll file a separate issue for that.

Issue #1520, opened 2019-06-14
Re-enable attach by pid tests (test_attaching_by_pid)


fabioz:

int19h:
If the tests are broken after re-enabling in a non-obvious way, we'll probably want to wait until after the refactoring is done, since that will also change this code path substantially.

fabioz:
I've done changes so that the attach to process can run code in the target pid, but there are still issues to fix. Namely:

- If the process was initially launched from vscode it doesn't really attach because it sees the debugger started in no debug mode.

- As it doesn't block before a connection, the tracing may not be set for existing threads (and thus the user is not able to actually debug existing threads) because the debugger is not setup when the method that sets the tracing for existing threads is called.

I'm not sure this is worth fixing now given that as @int19h mentioned, this may code is expected to change quite a bit.

fabioz:
I was thinking a bit more and I think that maybe a better approach would be first fixing https://github.com/microsoft/ptvsd/issues/509 (so, I'd move the code that does the settrace after the attach to pid to a module which would be used during a regular settrace, not only during the attach to pid) -- I think this should be enough to fix the 2nd issue I mentioned without changes that'd conflict with the ptvsd refactoring -- and being able to set the tracing to existing threads (in CPython) will also be pretty nice in general.

The first issue I mentioned above could be a non-issue after the refactoring (depending on how it's done, so, probably something to keep in mind -- in particular, if the adapter does a launch directly without traces of the debugger it becomes a non-issue, if that's still an issue after the refactoring this will need to be better checked, but as it's very likely that it's the same code being refactored I'd rather wait for the refactoring to take a look at that).

int19h:
The main reason why we even need to do anything about this today in ptvsd is for two reasons:

First, because VSCode needs to receive subprocess notifications, so that it can attach to them. That logic can be moved to pydevd entirely, we just need to standardize on the message format. It was impossible in the original model, because pydevd didn't talk DAP directly on its socket. But now that it is a DAP server, I think this is easy to solve.

[...]

PR #1582, opened 2019-07-11
Attach to pid fixes with threading. Fixes #1542,#1520

### LINKED ISSUE #1542: If threading is not imported in main thread, attach to pid does not work properly.
I was doing some manual testing for the attach to pid and in this situation, if the user never imported `threading`, we don't currently set the tracing properly.

The main reason for this is that `threading` seems to expect that it's **imported** from the actual main thread and not from another thread (because presumably one would import `threading` to create a thread in the first place) and it creates the initial `_MainThread` with the current thread (even if that's not the actual main thread).

Now, in the attach to pid, what happens is that we create a thread on the target process and then call the `settrace` in that thread... and if `threading` was never imported it'll consider that dummy thread as the main thread because we import `threading` in the dummy thread and the actual main thread will not be visible to `threading.enumerate()`.

I still don't know the best way to overcome this...

Additional notes:
- This happens in Python 2 and Python 3.
- We can detect that if `threading` not in `sys.modules` during the attach.
- In Linux we don't create a thread during the attach and use the current running thread, so, this shouldn't be a problem (although it could be an issue if the user has multiple threads not started through `threading` and we happen to stop in the wrong thread, then we'd import `threading` in the wrong thread).

fabioz:
As a note, Linux (on Azure) is failing because there's no gdb and I'm not exactly sure how to do that on Azure (see: https://github.com/microsoft/ptvsd/pull/1582/files#diff-354f30a63fb0907d4ad57269548329e3R12 for the changes needed to run on travis -- i.e.: install gdb and set the `ptrace_scope` to `0`).

I'm currently taking a look at the failures on Python 2.7 (windows and mac), so, this still can't be merged.


karthiknadig:
Updated the build to have gdb.

karthiknadig:
With the latest run. Linux passes, windows 2.7 is the flask issue.

karthiknadig:
Just merged the temp fix for the flask issue.

src/ptvsd/_vendored/pydevd/pydevd_attach_to_process/windows/attach.cpp
```diff
@@ -350,50 +100,34 @@ std::wstring GetCurrentModuleFilename() {
...
-HANDLE g_initedEvent;
-int AttachCallback(void *initThreads) {
+
+int AttachCallback(void *voidInitializeThreadingInfo) {
     // initialize us for threading, this will acquire the GIL if not already created, and is a nop if the GIL is created.
     // This leaves us in the proper state when we return back to the runtime whether the GIL was created or not before
     // we were called.
-    ((PyEval_Lock*)initThreads)();
-    SetEvent(g_initedEvent);
+    InitializeThreadingInfo* initializeThreadingInfo = (InitializeThreadingInfo*)voidInitializeThreadingInfo;
```
karthiknadig: NIT: reinterpret_cast<InitializeThreadingInfo*>()

src/ptvsd/_vendored/pydevd/pydevd_attach_to_process/windows/attach.cpp
```diff
@@ -495,7 +229,7 @@ void SuspendThreads(ThreadMap &suspendedThreads, Py_AddPendingCall* addPendingCa
                                 GetThreadContext(hThread, &context);
 
 #if defined(_X86_)
-                                if(context.Eip >= *((DWORD*)addPendingCall) && context.Eip <= (*((DWORD*)addPendingCall)) + 0x100) {
+                                if (context.Eip >= *((DWORD*)addPendingCall) && context.Eip <= (*((DWORD*)addPendingCall)) + 0x100) {
```
karthiknadig: reinterpret_cast

fabioz:
> Other than `tests/func/test_attach.py::test_attaching_by_pid_no_threading[code-attach_pid]` test failure in Windows 2.7. rest looks good.

I just did another change to improve that (so, if it passes I'll merge to master).

fabioz:
All green 🎉

PR #1593, opened 2019-07-15
Fix #1555, #1585 + misc test improvements

### LINKED ISSUE #1555: Child process watchdog for tests
Issues like #1552 can be hard to diagnose, because unrelated tests fail because ptvsd from an older failed test is still around. We need some way to clean those up, so that test reports only contain relevant failures.

We have previously done so by using a process watchdog - in fact, it is still there in watchdog.py. The idea is to have a process external to the test worker, which monitors said worker, and kills any debuggee processes that are left around after the worker exits.

This worked, but we disabled it for perf reasons. It was a naive implementation that used one watchdog per `DebugSession` instance. A full test run thus resulted in accumulation of hundreds of those processes, all watching the same test worker, causing OOM issues.

We need to fix the watchdog such that it's a single process per test worker. The watchdog should communicate with the test worker and receive PIDs for new child processes spawned by `debug.Session` (e.g. over stdin), and clean them all up once the worker exits.

### LINKED ISSUE #1585: Output tests fail sporadically
This is everything in test_output.py, and also test_log_point in test_breakpoints.py.

The problem is the same as before - ptvsd (or rather pydevd) currently doesn't guarantee that all output is flushed by the time the debuggee process exits. 

Until this can be fully fixed, we need to provide some internal hook in the debugger, specifically for the tests, which would then ensure that all pending `output` messages are flushed before exiting. Otherwise, tests will have to resort to hacks such as sleeping (which are not 100% reliable).

int19h:
Fixes #1555, Fixes #1585


tests/ptvsd/server/test_attach.py
```diff
@@ -6,55 +6,47 @@
...
 from tests.timeline import Event
 
 
-@pytest.mark.parametrize("wait_for_attach", ["waitOn", "waitOff"])
-@pytest.mark.parametrize("is_attached", ["attachCheckOn", "attachCheckOff"])
-@pytest.mark.parametrize("break_into", ["break", "pause"])
-def test_attach(run_as, wait_for_attach, is_attached, break_into):
+@pytest.mark.parametrize("wait_for_attach", ["wait_for_attach", ""])
+@pytest.mark.parametrize("is_attached", ["is_attached", ""])
+@pytest.mark.parametrize("break_into_debugger", ["break_into_debugger", ""])
```
karthiknadig: that should say "pause" (or something similar), since this test is doing `pause` if it is not `break_into_debugger`

tests/ptvsd/server/test_attach.py
```diff
@@ -6,55 +6,47 @@
...
-        session.initialize(
-            target=(run_as, attach1_py),
-            start_method="launch",
-            env=env,
-        )
+        session.initialize(target=(run_as, attach1_py))
         session.start_debugging()
 
-        if wait_for_attach == "waitOn":
+        if wait_for_attach:
```
karthiknadig: do we need bool() for clarity?

int19h: I think it's a common enough pattern ("" vs "foo") in tests that it's not worth the bother. Since there's no chance of `None` here, the usual problem with distinguishing it vs empty string is not applicable.

PR #1595, opened 2019-07-15
Create client socket and connect to debug server

karthiknadig:


src/ptvsd/adapter/channels.py
```diff
@@ -65,11 +63,23 @@ def connect_to_ide(self, address=None):
...
-        raise NotImplementedError
+
+        # Import message handlers lazily to avoid circular imports.
+        from ptvsd.adapter import messages
+
+        host, port = address
+        sock = socket.create_client(host, port)
+        sock.connect(address)
+
+        server_stream = messaging.JsonIOStream.from_socket(sock, "SVR")
```
int19h: I've moved off terse acronyms for these, now that we have a lot of channels all around :) Let's just do "server" etc.

The main reason why it was terse and all-caps before is because the messages were all logged as a single line, and so I wanted the channel name to stand out on that line, but without making it much longer. Now that timestamps are in their own column, and JSON is wrapped and indented, the first line of any message just has the channel name by itself - so it doesn't need caps to stand out, and it can be as verbose as we want.

src/ptvsd/adapter/messages.py
```diff
@@ -155,8 +155,12 @@ def attach_request(self, request):
         self.terminate_on_disconnect = False
         self._debug_config(request)
 
-        # TODO: get address and port
-        channels.connect_to_server()
+        host = request.arguments.get("host", constants.DEFAULT_DEBUG_SERVER_HOST)
+        port = request.arguments.get("port", constants.DEFAULT_DEBUG_SERVER_PORT)
```
int19h: To mirror the pattern in `ptvsd.common` and `ptvsd.server`, let's add `ptvsd.adapter.options` to this package along the same lines. Then all the defaults can go there - and the code here will overwrite them if they were supplied in the message, but then just use `options.port` etc to connect.

PR #1609, opened 2019-07-18
Add launcher script that handles wait-on-* feature

### LINKED ISSUE #1608: Script to handle out-of-proc Wait on normal and abnormal exits

karthiknadig:
Fixes #1608

### DIFF
--- a/src/ptvsd/common/launcher.py
+++ b/src/ptvsd/common/launcher.py
@@ -1,0 +1,69 @@
+import sys
+import subprocess
+
+
+_wait_on_normal_exit = False
+_wait_on_abnormal_exit = False
+
+
+HELP = ('''Usage: launcher [--wait-on-normal] [--wait-on-abnormal] <args>
+python launcher.py --wait-on-normal --wait-on-abnormal myscript.py -a 123      # script with args
+python launcher.py --wait-on-normal --wait-on-abnormal -m mymodule -a 123      # module with args
+python launcher.py --wait-on-normal --wait-on-abnormal -c "<code>"             # run code
+''')
+
+
+def main(argv=sys.argv):
+    try:
+        process_args = list(parse(argv))
+    except Exception as ex:
+        print(HELP + '\nError: ' + str(ex), file=sys.stderr)
+        sys.exit(2)
+
+    p = subprocess.Popen(args=process_args, executable=sys.executable)
+    exit_code = p.wait()
+
+    if _wait_on_normal_exit and exit_code == 0:
+        _wait_for_user()
+    elif _wait_on_abnormal_exit and exit_code != 0:
+        _wait_for_user()
+
+    sys.exit(exit_code)
+
+
+def _wait_for_user():
+    if bool(sys.__stdout__) and bool(sys.__stdin__):
+        try:
+            import msvcrt
+        except ImportError:
+            sys.__stdout__.write('Press Enter to continue . . . ')
+            sys.__stdout__.flush()
+            sys.__stdin__.read(1)
+        else:
+            sys.__stdout__.write('Press any key to continue . . . ')
+            sys.__stdout__.flush()
+            msvcrt.getch()
+
+
+def parse_arg(arg):
+    if arg == "--wait-on-normal":
+        global _wait_on_normal_exit
+        _wait_on_normal_exit = True
+    elif arg == "--wait-on-abnormal":
+        global _wait_on_abnormal_exit
+        _wait_on_abnormal_exit = True
+    else:
+        return False
+    return True
+
+
+def parse(argv):
+    ret_argv = []
+    for arg in argv:
+        if not parse_arg(arg):
+            ret_argv += [arg]
+    return ret_argv
+
+
+if __name__ == "__main__":
+    main()
--- a/src/ptvsd/common/launcher.py
+++ b/src/ptvsd/common/launcher.py
@@ -9,25 +9,27 @@ import subprocess
 import sys
 
 
+WAIT_ON_NORMAL_SWITCH = "--wait-on-normal"
+WAIT_ON_ABNORMAL_SWITCH = "--wait-on-abnormal"
+
+
 _wait_on_normal_exit = False
 _wait_on_abnormal_exit = False
 
 
-HELP = """Usage: launcher [--wait-on-normal] [--wait-on-abnormal] <args>
-python launcher.py --wait-on-normal --wait-on-abnormal myscript.py -a 123      # script with args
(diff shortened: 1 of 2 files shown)


src/ptvsd/common/launcher.py
```diff
@@ -0,0 +1,69 @@
+import sys
```
int19h: Needs the copyright header and `from __future___`

(The second part isn't really necessary, but I'd rather keep it uniform so that we don't forget to add it when it does become necessary for new additions to the file.)

int19h: Oh, and can you also run Black on it for the initial commit?

src/ptvsd/common/launcher.py
```diff
@@ -0,0 +1,76 @@
...
+        _wait_on_normal_exit = True
+    elif arg == "--wait-on-abnormal":
+        global _wait_on_abnormal_exit
+        _wait_on_abnormal_exit = True
+    else:
+        return False
+    return True
+
+
+def parse(argv):
```
int19h: This will break if the user does:
```
python launcher.py myscript.py --wait-on-normal
```
(i.e. if it's supposed to be a switch for their own script, that just happens to use the same name)

To work in this case, it needs to do something similar to what arg parsing code in `ptvsd/server/__main__.py` does today. So basically parse them, separate out our args like `--wait` from Python args like `-X`, and stop once it sees the filename, or `-m`, or `-c`.

But this is complicated. Since this script isn't meant to be run by the users, I think we can do it much easier by requiring an explicit separator, e.g. `--`. So the adapter would always invoke it like so:
```
python launcher.py --wait-on-normal -- myscript.py --wait-on-normal
```
and then it just needs to parse up to `--` (and complain about any invalid switches in that part), and hand everything after `--` to Python as is.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,109 @@
...
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+import os.path
+import pytest
+import subprocess
+import sys
+import time
+import ptvsd.common.launcher
+from ptvsd.common.launcher import parse, WAIT_ON_NORMAL_SWITCH, WAIT_ON_ABNORMAL_SWITCH
```
int19h: For the imports, in new code, I've been sticking to two patterns to keep it readable, and avoid some issues we had in the past.

1. For modules from `ptvsd` and `tests`, consistently use `from ... import` - so `from ptvsd.common import launcher` here. I also insert a blank line between the modules that come from stdlib or site-packages (including non-standard pip-installed ones like `pytest`), and modules from src/ or tests/ (including `pydevd`) - just so that it's clear which is our code, and which is external dependencies.

2. Always import modules, never specific things from them. The reason is mainly because globals can be reassigned, and then you're stuck with the old value for one in any code that did `from ... import`. Most globals aren't reassigned; but the problem is that if that ever changes, the old code will quietly do the wrong thing. In addition, this also seems to work better in many cases where circular dependencies occur.

There is one exception to the last rule - `ptvsd.common.compat`. For that one, it's okay to do `from ptvsd.common.compat import bytes, unicode, xrange` etc; but specifically only for things that correspond to `builtins`. This is so that the rest of the code looks as if they really were built-in on all Python versions - i.e. `unicode(s)` rather than `compat.unicode(s)`.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,109 @@
...
+
+launcher_py = os.path.abspath(ptvsd.common.launcher.__file__)
+
+
+@pytest.mark.parametrize("run_as", ["file", "module", "code"])
+@pytest.mark.parametrize("mode", ["normal", "abnormal", "both", ""])
```
int19h: Nifty trick: you can use `"normal+abnormal"` here instead. Then `split()` it inside the test to get a proper list, and do checks like `if "normal" in mode` etc. That way it only needs two branches; and pytest will still print the names out properly.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.parametrize("seperator", ["separator", ""])
```
int19h: Spelling

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,109 @@
...
+
+    wait_for_user = (exit_code, mode) in [
+        (0, "normal"),
+        (10, "abnormal"),
+        (0, "both"),
+        (10, "both"),
+    ]
+
+    if wait_for_user:
+        while not p.stdout.read(5).startswith(b"Press"):
```
int19h: This one might not do the right thing. In unbuffered mode, there's no guarantee that a single `read()` from a pipe will actually read that many bytes - it can read less, so you have to check the returned value.

Also, regardless of the mode, this code needs to read the entirety of the message, because otherwise the child process might get blocked trying to write into a pipe, because nobody is reading on the other end.

So, to be foolproof, this needs to spin in a loop, reading one byte at a time, until the entire string collected so far `endswith(". . . ")`.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,109 @@
...
+            time.sleep(0.1)
+        # TODO: Fix sending input to the launched process to terminate
+        # p.stdin.write(b" ")
```
int19h: This should work properly on non-Windows platforms if you do `write("\n")` there.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,128 @@
...
+launcher_py = os.path.abspath(launcher.__file__)
+
+
+@pytest.mark.parametrize("run_as", ["file", "module", "code"])
+@pytest.mark.parametrize("mode", ["normal", "abnormal", "normal+abnormal", ""])
+@pytest.mark.parametrize("seperator", ["seperator", ""])
+def test_launcher_parser(mode, seperator, run_as):
+    args = []
+
+    switch = mode.split("+") if "+" in mode else [mode]
```
int19h: No need to check - you can just split, and you'll get back a single-item list if the separator was not there.

tests/ptvsd/common/test_launcher.py
```diff
@@ -0,0 +1,128 @@
...
+def test_launcher(pyfile, mode, exit_code, run_as):
+    @pyfile
+    def code_to_run():
+        import sys
+
+        sys.exit(int(sys.argv[1]))
+
+    args = [sys.executable, launcher_py]
+
+    switch = mode.split("+") if "+" in mode else [mode]
```
int19h: Same here.

PR #1615, opened 2019-07-23
Fix #1614: Test run times out on Python 2.7

### LINKED ISSUE #1614: Test run times out on Python 2.7
(dbg_adapter_refactor branch only)

It's not just a CI issue - it repros locally on Linux, as well. The tests themselves complete fine, but something prevents the main pytest process from exiting - it gets stuck waiting for child nodes. Most likely, our test infrastructure does something on exit that blocks.

int19h:
Close the watchdog message channel from the test process side after "stop" command is sent and acknowledged.

### DIFF
--- a/src/ptvsd/common/messaging.py
+++ b/src/ptvsd/common/messaging.py
@@ -99,8 +99,16 @@ class JsonIOStream(object):
         """Closes the stream, the reader, and the writer.
         """
         self._is_closing = True
-        self._reader.close()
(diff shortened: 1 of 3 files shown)


int19h:
Not quite working yet. I'll ping you when it does...

int19h:
@karthiknadig It's working now - no more timeouts, and our CI is passing as well.

PR #1634, opened 2019-07-26
Gracefully fall back to ASCII when printing through debugger (!1633)

Olivier-OH:


int19h:
Thank you for the PR!

The bug is in part which creates the dict that is being serialized here - generic `NetCommand` is not the best place to tackle that. Basically, by the time it gets to this code, that string should have already been Unicode, using the appropriate encoding depending on the context (not the least because the context often determines the correct encoding without having to guess which one is the most suitable).

int19h:
In fact, the problem is even further up the stack.

This is the code that makes the message:
https://github.com/microsoft/ptvsd/blob/7ceda11fc43ce95ac63e77da1d0108aaee5d128a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py#L248-L253

This one could take care of the encoding. But it, in turn, is invoked from here:
https://github.com/microsoft/ptvsd/blob/7ceda11fc43ce95ac63e77da1d0108aaee5d128a/src/ptvsd/_vendored/pydevd/pydevd.py#L2039-L2060

Which *does* take care of encoding - but it does so by forcing it to be UTF-8 `bytes` on py2. I believe this is because this code was originally written for the XML pydevd protocol. But for DAP (JSON-based), it should really just be Unicode regardless of which Python version it is.

To minimize the scope of the fix, it is easier to just decode it in the JSON implementation of `make_io_message` if needed. It doesn't even need a Python version check - you can check for `isinstance(bytes)` there, and if it is, you can safely assume that it is UTF-8 and decode it as such before passing it to `OutputEventBody`.

int19h:
Also, can you please add a test for this to `tests/func/test_output.py`? I'm kinda horrified that we don't already have one - and if we did, this bug wouldn't get through. 

You can just take `test_with_tab_in_output` and adjust it to include some non-ASCII characters. Ideally, not just Latin-1, to make sure that it handles the more exotic stuff properly as well - `test_path_with_unicode` in `test_breakpoints.py` has some string literals you could reuse. The test can force Python to use a specific encoding for stdout by setting `PYTHONIOENCODING` like so:
```py
        session.initialize(
            ...
            env={'PYTHONIOENCODING': 'utf-8'},
        )
```
(the encoding must be 'utf-8', because on Python 3.6+ on Win32, it's always that regardless of the environment variable)

Olivier-OH:
Thanks for the feedback. I'll have a look into all of this and try to clean all this up.

Olivier-OH:
Actually, the encoding of this string MIGHT be unicode, not just latin-1 (which is a subset of unicode).

I'll try to decode the binary string as unicode and replace unknown character codes, then.

int19h:
It looks like in the `IS_PY2` branch, it'll try to encode it as UTF-8 if what was passed to `write()` was `unicode`, but otherwise it'll leave it as is. It's somewhat strange, because I would have expected it to be encoded using `self.encoding` instead. 

It's not a good arrangement, because in `make_io_message` it means that you basically don't know what the encoding is - it could be either of these two. Trying to guess which one it is is a bad idea - you can't assume that it's Latin-1 if it isn't UTF-8, and another encoding might accidentally match...

I think it would be best to move the entire encoding logic out of `write()` into the corresponding `make_io_message` implementations. This way, each can do what it needs: the XML one (in `pydevd_net_command_factory.py`) should just use the existing logic, if only for backwards compatibility; and the JSON one (in `pydevd_net_command_factory_json.py`) can use any `unicode` instance as is, and decode`bytes` using the supplied encoding without any guessing. But this also means that `make_io_message` should now take string + encoding, not just string. 

@fabioz, does this make sense?

fabioz:
@int19h 

The current assumption is that `make_io_message` receives a `utf-8` encoded message on Python 2 and an unicode on Python 3 (and it's really a bug that it doesn't verify that the bytes sent were actually `utf-8`) -- that's for both backends (xml and json).

So, the simplest fix would actually be adding an `else` to the handling on `write` when `IS_PY2` (in `pydevd.py` -- `pydevd._CustomWriter.write`) which makes sure that the text sent is actually `utf-8` (and if it's not, it should replace characters).

i.e.:

```
if isinstance(s, unicode):  # noqa
    ...
else:
    s = s.decode(self.encoding, 'replace').encode('utf-8', 'replace')
```

And this definitely should have tests added...

@Olivier-OH do you think you'll be able to provide a patch with that fix + tests?

Olivier-OH:
There will be a delay, I'm on vacation. I'll look into this next week and see whether I have time to deliver something. 

Cheers

Olivier-OH:
Hi there.

I've made the fix in the _CustomWriter class, but i've had hell of a time running the test (i could not run it). I would like some help.

It seems that there's something to do with utf8 chars being in my test function. Maybe I should have bytes and convert them into a py2 `str` at runtime just to be sure?

fabioz:
> Hi there.
> 
> I've made the fix in the _CustomWriter class, but i've had hell of a time running the test (i could not run it). I would like some help.
> 
> It seems that there's something to do with utf8 chars being in my test function. Maybe I should have bytes and convert them into a py2 `str` at runtime just to be sure?

I think that you can just use the hex notation to create the string (maybe that'll be enough). 

i.e.: you can just print something as the string below in the test (which is invalid utf-8 and the new code should handle):

```
b'\xe8\xF0\x80\x80\x80'
```

As a note, on py2 `byte` and `str` is the same thing, so, no conversion is needed.

Olivier-OH:
> As a note, on py2 `byte` and `str` is the same thing, so, no conversion is needed.

Oh yes, I forgot ^_^

karthiknadig:
@Olivier-OH Ignore the pypy test failures.

Olivier-OH:
I've just updated the tests again. there are errors on the continuous integration platform that I don't understand, they seem to be about the test environment.

Anyway, the test fails when run on my machine here.

``` python
def test_non_ascii_output(pyfile, run_as, start_method):
    if IS_PY2:
        @pyfile
        def code_to_debug():
            from dbgimporter import import_and_enable_debugger
            import_and_enable_debugger()
            a = b'\xc3\xa9 \xc3\xa0 \xc3\xb6 \xc3\xb9'
            print(a)
            # Break here so we are sure to get the output event.
            a = 1  # @bp1
    else:
        @pyfile
        def code_to_debug():
            from dbgimporter import import_and_enable_debugger
            import_and_enable_debugger()
            a = b'\xc3\xa9 \xc3\xa0 \xc3\xb6 \xc3\xb9'.decode('utf-8')
            print(a)
            # Break here so we are sure to get the output event.
            a = 1  # @bp1

    line_numbers = get_marked_line_numbers(code_to_debug)
    with DebugSession() as session:
        session.initialize(
            target=(run_as, code_to_debug),
            start_method=start_method,
            env={'PYTHONIOENCODING': 'utf-8'}
        )

        session.set_breakpoints(code_to_debug, [line_numbers['bp1']])
        session.start_debugging()

        # Breakpoint at the end just to make sure we get all output events.
        session.wait_for_thread_stopped()
        session.send_request('continue').wait_for_response(freeze=False)
        session.wait_for_exit()

        output = session.all_occurrences_of(Event('output', ANY.dict_with({'category': 'stdout'})))
        output_str = ''.join(o.body['output'] for o in output).strip()
        assertion_value = b'\xc3\xa9 \xc3\xa0 \xc3\xb6 \xc3\xb9'
        if IS_PY3K:
            assertion_value = assertion_value.decode('utf-8')
        assert isinstance(output_str, str)
        assert output_str == assertion_value
```

The assertion `assert isinstance(output_str, str)` fails. With python2, the type is actually `unicode`. It seems the `_CustomWriter` class is not used here, so I'm not sure of how I can test my changes.

**However**, with his version of the `_CustomWriter` in my VSCode install, the error message is decoded with unknown charpoints replaced, so the crash is averted as expected.

fabioz:
> The assertion `assert isinstance(output_str, str)` fails. With python2, the type is actually `unicode`. It seems the `_CustomWriter` class is not used here, so I'm not sure of how I can test my changes.

Have you actually checked that the `_CustomWriter` is not used? (that seems unlikely to me)

i.e.: it's expected that if you put utf-8 bytes it may come out as unicode.
```
Python 2.7.16 |Anaconda, Inc.| (default, Mar 14 2019, 15:42:17) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.dumps(b'\xc3\xa9 \xc3\xa0 \xc3\xb6 \xc3\xb9')
'"\\u00e9 \\u00e0 \\u00f6 \\u00f9"'
>>> s = json.loads(json.dumps(b'\xc3\xa9 \xc3\xa0 \xc3\xb6 \xc3\xb9'))
>>> s
u'\xe9 \xe0 \xf6 \xf9'
>>> type(s)
<type 'unicode'>
>>>
```

So, it may be that you just don't need that assert in the test...

fabioz:
I'll take a look into fixing this issue/testing based on what's been already done...

fabioz:
Closing in favor of: https://github.com/microsoft/ptvsd/pull/1862

Issue #1448, opened 2019-05-20
Support and test remote attach


karthiknadig:
- [x] ptvsd.attach()
- [x] Fix and replace pydevd hooks
- [x] Update argument parsing code
- [x] Remove un-used code

karthiknadig:
Manually tested attach via CLI and import. Created a separate item to update automated tests.

PR #1651, opened 2019-08-01
#1449, #1498, partially #1448

### LINKED ISSUE #1449: Support and test enable_attach, wait_for_attach, is_attached
- [x] enable_attach
- [x] wait_for_attach
- [x] is_attached
- [x] break_into_debugger
- [x] debug_this_thread

### LINKED ISSUE #1498: pydevd must send custom pydevdSubprocess event when a process is forked or spawned
The `pydevdSubprocess` event should be forwarded to IDE as ` ptvsd_subprocess` event.

karthiknadig:
Fixes #1449
Fixes #1498


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -404,18 +413,34 @@ def do_kill_pydev_thread(self):
...
 
 def create_server_socket(host, port):
-    s = socket(AF_INET, SOCK_STREAM)
-    s.settimeout(None)
+    server = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP)
+    if IS_WINDOWS:
+        try:
+            from socket import SIO_LOOPBACK_FAST_PATH
+            server.ioctl(SIO_LOOPBACK_FAST_PATH, True)
+        except ImportError:
```
int19h: You can use a tuple here to catch both: `except ImportError, AttributeError`.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -404,18 +413,34 @@ def do_kill_pydev_thread(self):
...
+            pass
+        except AttributeError:
+            pass  # Not supported in python 2.* or <3.6
+        except OSError as ose:
+            if ose.winerror == 10045:  # Not supported by OS
```
int19h: If you invert the condition, it only needs one branch here.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -404,18 +413,34 @@ def do_kill_pydev_thread(self):
 
 
 def create_server_socket(host, port):
-    s = socket(AF_INET, SOCK_STREAM)
-    s.settimeout(None)
+    server = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP)
+    if IS_WINDOWS:
+        try:
+            from socket import SIO_LOOPBACK_FAST_PATH
+            server.ioctl(SIO_LOOPBACK_FAST_PATH, True)
```
int19h: `ioctl` call shouldn't be inside `try`, since we're only interested in `ImportError` for that one constant being imported. It can go into `else:`.

src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -1957,6 +1981,14 @@ def _wait_for_attach():
     py_db.block_until_configuration_done()
 
 
+def _is_attached():
+    '''
+    Can be called any time to check if the connection was established and the DAP (Debug Adapter Protocol) has sent
+    the ConfigurationDone request.
+    '''
+    py_db = get_global_debugger()
+    return bool(py_db) and py_db.is_attached()
```
int19h: It doesn't make sense to use `bool()` in this context, since it's the same with or without it. If you want to make it a more narrow check, it should be `py_db is not None`.

src/ptvsd/adapter/messages.py
```diff
@@ -153,6 +154,7 @@ def initialize_request(self, request):
     def _debug_config(self, request):
         assert request.command in ("launch", "attach")
         self._shared.start_method = request.command
+        self._shared.arguments = request.arguments
```
int19h: I don't think these need to be shared - `_configure` is invoked by `launch_request` and `attach_request`, and they can just pass the request object to it.

src/ptvsd/adapter/messages.py
```diff
@@ -213,6 +233,7 @@ def _configure(self):
 
         # Let the IDE know that it can begin configuring the adapter.
         state.change("configuring")
+        self._set_debugger_properties(self._shared.arguments)
```
int19h: This should go before the state change, right after the messages are replayed. Just to make sure that we get that message in first before any other comes (the fact that handlers are processed one by one already kinda ensures that, but the state transition makes it clearer).

src/ptvsd/server/attach.py
```diff
@@ -83,26 +69,23 @@ def enable_attach(address=(DEFAULT_HOST, DEFAULT_PORT), redirect_output=None, lo
...
-    dbg = get_global_debugger()
-    if bool(dbg):
-        dbg.dap_debugger_attached.clear()
+        return None, None
 
     # Ensure port is int
-    port = address[1]
-    address = (address[0], port if type(port) is int else int(port))
+    host, port = address
+    address = (host, port if type(port) is int else int(port))
```
int19h: No need to check for type - `int()` is a no-op if it's already an int. Not sure why it was doing it like that before...

src/ptvsd/server/multiproc.py
```diff
@@ -209,11 +207,8 @@ def notify_root(port):
 
     if not response['incomingConnection']:
         ptvsd.server.log.debug('No IDE connection is expected for this subprocess; unpausing.')
-        debugger = get_global_debugger()
-        while debugger is None:
+        while not ptvsd.is_attached():
```
int19h: This is not doing the right thing - note the message it logs right above.

The point of that code originally was to basically unblock the `wait_for_attach()` call that is happening on the main thread because we started the subprocess with `--wait`. 

So, it looks like we'll need an API for that also. Basically, `_pydevd.wait_for_attach()` needs to be interruptible somehow. So it'll probably need to use a condition variable there so that it can wait on two events simultaneously - one for `configurationDone`, and one for the interrupt request.

src/ptvsd/adapter/messages.py
```diff
@@ -183,12 +184,30 @@ def attach_request(self, request):
...
+    def _set_debugger_properties(self, args):
+        client_os_type = None
+        if 'WindowsClient' in args.get('debugOptions', []):
+            client_os_type = 'WINDOWS'
+        elif 'UnixClient' in args.get('debugOptions', []):
+            client_os_type = 'UNIX'
+        else:
+            client_os_type = 'WINDOWS' if platform.system() == 'Windows' else 'UNIX'
+
+        self._server.send_request("setDebuggerProperty", arguments={
```
int19h: This should propagate the error from the server back to the IDE. You need to catch `MessageHandlingError`, and then call `propagate(request)` on it, so this will need to take request as an argument, not just the body.

Also, `send_request(...).wait_for_response()` can be shortened to just `request(...)`.

src/ptvsd/adapter/messages.py
```diff
@@ -183,12 +184,30 @@ def attach_request(self, request):
         options.port = int(request.arguments.get("port", options.port))
         _channels.connect_to_server(address=(options.host, options.port))
 
-        return self._configure()
+        return self._configure(request)
+
+    def _set_debugger_properties(self, args):
+        client_os_type = None
+        if 'WindowsClient' in args.get('debugOptions', []):
```
int19h: Take a look at how "debugOptions" are handled in debuggee.py to add some validation here. If `debugOptions` is not a JSON array, this will crash the adapter immediately, rather than reporting request as invalid.

karthiknadig: Is this for the VS style debugOptions?

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -404,18 +413,31 @@ def do_kill_pydev_thread(self):
...
     try:
-        from socket import SO_REUSEPORT
-        s.setsockopt(SOL_SOCKET, SO_REUSEPORT, 1)
-    except ImportError:
-        s.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
+        server.bind((host, port))
+        server.settimeout(None)
+    except Exception:
+        server.close()
+        raise
```
int19h: Didn't notice this one earlier, but I think this should also do cleanup for all the calls on the socket above - e.g. `ioctl` can definitely fail, and I'm not sure, but `setsockopt` might also fail? There's definitely that line above that will re-raise exception from `ioctl` which will result in the socket not getting cleaned up.

It's probably best to wrap the entire function body after `server` is created into one giant `try..except` that closes and re-raises.

PR #1653, opened 2019-08-03
Add --log-dir switch to adapter

karthiknadig:


src/ptvsd/adapter/__main__.py
```diff
@@ -75,6 +76,15 @@ def _parse_argv():
...
 
+    parser.add_argument(
+        "--log-dir",
+        type=str,
+        nargs="?",
+        default=None,
+        metavar="DIR",
+        help="enable logging and use DIR to save adapter logs"
+    )
+
```
int19h: If I'm reading this right, it would allow `--log-dir` without any command line argument, in which case it sets it to `None`, which disables it. I don't think that's what somebody who just said `--log-dir` would intend it to do.

So I think the argument should be mandatory - so no `nargs` and no `default`.

PR #1660, opened 2019-08-06
Update tests to use python debug adapter

karthiknadig:
This is a partial PR, to iron out some issues with DA and see how the test infrastructure should look like with the new DA


karthiknadig:
Note: This is still partial. this PR unblocks tests work so we can work in parallel restoring the tests.

Issue #1680, opened 2019-08-11
Fatal errors in message loop do not fail fast


int19h:
When a fatal error happens on one of the background threads that parse and handle messages, that error is logged by Python, but the adapter continues running. If the error happens on just one of the threads, the other thread remains alive, and the adapter never exits, even though it is broken. 

Fatal errors (i.e. anything that is not `MessageHandlingError`) should cause the adapter to fail fast instead, so that the failure is apparent, and the adapter doesn't try to continue running in an invalid state.

Issue #1679, opened 2019-08-11
"exited" event sometimes reports "exitCode": null


int19h:
This happens if the debuggee exits, but its exit code couldn't be retrieved for whatever reason.

This is incorrect per the DAP spec - "exitCode" must always be present, and must be an int.

Issue #1678, opened 2019-08-11
Do not rely on "processId" being returned by "runInTerminal" request


int19h:
VSCode does not return "processId" from "runInTerminal" requests: https://github.com/microsoft/vscode/issues/61640

Thus, support for "console":"integratedTerminal" and "console":"externalTerminal" cannot rely on that - it has to retrieve the PID from the "process" event sent by the debug server, and gracefully handle situations where the debuggee exits or crashes before it gets a chance to send it.

PR #1681, opened 2019-08-12
Fix #1648, #1678, #1679, #1680

### LINKED ISSUE #1648: Messaging does not allow reverse requests
While handling a DAP message that came from the IDE, it is currently impossible to issue a request to the IDE and wait for a response - the handler executes on the same thread that processes all incoming messages, and thus the response won't be processed until it returns.

This is now an issue because we need to issue "runInTerminal" request from the adapter to the IDE while handling IDE's "launch" request, in order to properly implement "integratedTerminal" and "externalTerminal".

int19h:


src/ptvsd/adapter/debuggee.py
```diff
@@ -258,19 +269,60 @@ def _spawn_terminal(request, spawn_info):
...
+            _got_pid.set()
+            ProcessTracker().track(pid, after_exit=after_exit)
+        except Exception as exc:
+            # If we can't track it, we won't be able to detect if it exited or retrieve
+            # the exit code, so fail immediately.
+            raise process_event.cant_handle(
+                "Couldn't get debuggee process handle: {0}", str(exc)
+            )
+
+    elif pid != sys_pid:
```
karthiknadig: This check will always fail if launch handler uses launcher.py to run user code.

src/ptvsd/adapter/messages.py
```diff
@@ -407,13 +425,24 @@ def initialized_event(self, event):
         # also remove the 'initialized' event sent from IDE messages.
         pass
 
+    @_only_allowed_while("initializing")
+    def process_event(self, event):
+        try:
+            debuggee.parse_pid(event)
+        except Exception:
+            # If we couldn't retrieve or validate PID, we can't safely continue
```
karthiknadig: Shouldn't this be the case only for launch? for attach it should not really matter.

int19h: This particular bit will work as intended, because `parse_pid()` won't do anything, so it won't raise an exception.

However, the bit in `_configure()` that does `wait_for_pid()` is wrong for this reason, and breaks "attach".

int19h: I'll put this in a conditional for clarity tho.

PR #1694, opened 2019-08-14
 noDebug mode support

### LINKED ISSUE #1664: Support nodebug flag for launch

karthiknadig:
Fixes #1664

This PR enables nodebug mode, but has a limitation see. #1693


karthiknadig:
We discussed an alternate solution, we will try that.

PR #1696, opened 2019-08-16
Support for noDebug

### LINKED ISSUE #1664: Support nodebug flag for launch

### LINKED ISSUE #1692: Value error during state transitions
```
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\common\messaging.py", line 733, in _handle
                 result = handler(self)
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\adapter\messages.py", line 309, in _disconnect_or_terminate_request
                 state.change("shutting_down")
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\adapter\state.py", line 58, in change
                 State().state = new_state
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\common\singleton.py", line 163, in __setattr__
                 return object.__setattr__(self, name, value)
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\common\singleton.py", line 185, in lock_and_call
                 return func(self, *args, **kwargs)
               File "C:\GIT\ptvsd\src\ptvsd\adapter/../..\ptvsd\adapter\state.py", line 45, in state
                 if STATES.index(state) >= STATES.index(new_state):
             ValueError: tuple.index(x): x not in tuple
```

### LINKED ISSUE #1693: In NoDebug mode there is no way to notify pid to adapter

karthiknadig:
Fixes #1664, Fixes #1692, Fixes #1693


int19h:
Looks like it needs some squashing to clean up the dot commits.

src/ptvsd/adapter/debuggee.py
```diff
@@ -461,3 +457,45 @@ def _pid_to_handle(pid):
...
+                self.host,
+                self.port
+            )
+            sock, (l_host, l_port) = self.listener.accept()
+        finally:
+            self.listener.close()
+        log.info("Launcher connection accepted from {0}:{1}.", l_host, l_port)
+
+        data = b''
+        with sock:
```
int19h: I think you can skip the loop if you do `makefile(sock).read()` - this should give you everything that was sent right up until the socket gets closed. This way, you can also specify text mode for `makefile()`, and don't need to decode anything.

src/ptvsd/adapter/debuggee.py
```diff
@@ -461,3 +457,45 @@ def _pid_to_handle(pid):
...
+    def __init__(self):
+        self.listener = socket.create_server("127.0.0.1", 0)
+        self.host, self.port = self.listener.getsockname()
+        self.wait_thread = threading.Thread(
+            target=self._wait_for_pid, name="Process Pid Server"
+        )
+        self.wait_thread.daemon = True
+        self.wait_thread.start()
+
+    def _wait_for_pid(self):
```
int19h: The name is a bit confusing between this and the one without underscore. Since this does all the work to obtain the pid, and it's also inside a class for which this is the sole purpose, I think something like `_worker` would be clearer.

src/ptvsd/adapter/debuggee.py
```diff
@@ -461,3 +457,45 @@ def _pid_to_handle(pid):
...
+                    data += d
+            except OSError:
+                pass
+
+        self.pid = 0 if data == b'' else int(data.decode("ascii"))
+        log.info("Debuggee process Id received: ", self.pid)
+
+    def wait_for_pid(self):
+        self.wait_thread.join()
+        return self.pid
```
int19h: If the worker directly sets the global `pid` and notifies everybody via `_got_pid` event, this won't be needed, since it's exactly what the caller of this method does.

src/ptvsd/adapter/messages.py
```diff
@@ -147,7 +150,8 @@ def event(self, event):
 
     # Generic request handler, used if there's no specific handler below.
     def request(self, request):
-        return self._server.delegate(request)
+        if self._no_debug is False:
```
int19h: This needs to fail the requests otherwise. As it is, it will crash the message loop, because the handler didn't respond to a request.

Actually, come to think of it, this is also the case if the IDE sends us some request we don't know about, before we actually connect to the server - even if it's not a nodebug scenario. So this should check `_channels.server`, and delegate if it's there, and fail the request otherwise.

src/ptvsd/adapter/messages.py
```diff
@@ -256,15 +265,19 @@ def _configure(self, request):
             # before it had a chance to send the event - so wake up periodically, and
             # check whether server channel is still alive.
             while not debuggee.wait_for_pid(1):
-                if _channels.server() is None:
+                if self._no_debug is False and _channels.server() is None:
```
int19h: `not self._no_debug`

src/ptvsd/adapter/messages.py
```diff
@@ -256,15 +265,19 @@ def _configure(self, request):
...
-        return messaging.NO_RESPONSE  # will respond on "configurationDone"
+        if self._no_debug is False:
+            self._set_debugger_properties(request)
+            # Let the IDE know that it can begin configuring the adapter.
+            state.change("configuring")
+            self._ide.send_event("initialized")
+            return messaging.NO_RESPONSE  # will respond on "configurationDone"
+        else:
+            state.change("running_nodebug")
+            self._start_request.respond({})
```
int19h: It can just `return {}` here.

src/ptvsd/adapter/messages.py
```diff
@@ -342,6 +357,9 @@ def continue_request(self, request):
     @_only_allowed_while("configuring", "running")
     def ptvsd_systemInfo_request(self, request):
         result = {"ptvsd": {"version": ptvsd.__version__}}
+        if self._no_debug:
+            return result
+
```
int19h: This shouldn't be possible, since nodebug will never enter any of the states in which this request is allowed - so no need to check.

src/ptvsd/common/launcher.py
```diff
@@ -78,5 +80,21 @@ def parse(argv):
...
 
+def _send_pid(pid):
+    if "PTVSD_PID_SERVER_PORT" not in os.environ:
+        return
+
+    from ptvsd.common import socket
+    port = int(os.environ["PTVSD_PID_SERVER_PORT"])
+    with socket.create_client() as sock:
+        sock.connect(("127.0.0.1", port))
+        sock.sendall(str(pid).encode("ascii"))
```
int19h: I'm not entirely sure what this does on Python 2, since `str` is `bytes` there... does it even have `encode()`? 

But there's a handy trick with `%` operator for getting a number represented as `bytes` on any Python version consistently:
```py
b"%d" % pid
```
(In Python 3, `bytes` doesn't have `format()`, but it still supports `%`)

src/ptvsd/common/launcher.py
```diff
@@ -78,5 +80,21 @@ def parse(argv):
...
+
+
 if __name__ == "__main__":
+    # This is so we can correctly import socket from ptvsd.common
+    if "ptvsd" not in sys.modules:
+        sys.path[0] = os.path.dirname(__file__) + "/../../"
+        __import__("ptvsd")
+        del sys.path[0]
```
int19h: Since it's a client socket, and it doesn't really need any special `setsockopt` sauce, it might be easier to just use the standard `socket` module directly?

src/ptvsd/adapter/debuggee.py
```diff
@@ -58,11 +58,13 @@ def spawn_and_connect(request):
     Caller is responsible for calling start() on the returned channel.
     """
 
-    channel = channels.Channels().accept_connection_from_server(
-        ("127.0.0.1", 0),
-        before_accept=lambda address: _parse_request_and_spawn(request, address),
-    )
-    return channel
+    if request("noDebug", json.default(False)):
+        _parse_request_and_spawn(request, (None, None))
```
int19h: It might be better to pass `None` here to indicate that it's not an address at all. And then in `_parse_request`, it should deconstruct `address` into `host, port` only in the branch where it needs it for the command line args.

src/ptvsd/common/launcher.py
karthiknadig: I tried this, seems that importing stdlib socket here conflicts with ptvsd.common.socket. Since that ptvsd's socket is in the current working directory.

src/ptvsd/adapter/messages.py
```diff
@@ -256,15 +272,18 @@ def _configure(self, request):
             # before it had a chance to send the event - so wake up periodically, and
             # check whether server channel is still alive.
             while not debuggee.wait_for_pid(1):
-                if _channels.server() is None:
+                if not self._no_debug and _channels.server() is None:
```
int19h: There's one interesting new possibility here: if the launcher crashes or gets killed before it can communicate the PID back to us, this loop will never exit.

This isn't entirely new. Previously, if the server crashed while we were waiting for "process", that would also close the socket, and we'd detect it with the channel going away. But if it crashed before it connected, then we'd sit forever in `accept_connection_from_server()`.

So it looks like we need to add some timeouts. Since you're already in this code, can you limit the number of times this loop will spin, before it bails out and fails the request? Or actually maybe just convert it to a simple wait with a timeout... it's not really expected for the server to crash before it sends "process", so if we take a while longer to detect that, it's okay, so long as we don't hang.

And then `accept_connection_from_server()` just needs to call `listener.settimeout(...)` before `accept()`.

karthiknadig: What is a good timeout? say about 10 seconds.

int19h: It's all arbitrary, but I think if it can't start and connect within 10 seconds, the debug session is going to be unusably slow anyway :)

PR #1701, opened 2019-08-18
 Attach to pid support

### LINKED ISSUE #1686: Support for attach to pid in the new adapter

karthiknadig:
Fixes #1686 and some tests

### DIFF
--- a/src/ptvsd/adapter/debuggee.py
+++ b/src/ptvsd/adapter/debuggee.py
@@ -80,6 +80,47 @@ def spawn_and_connect(request):
             before_accept=lambda address: _parse_request_and_spawn(request, address),
         )
 
+def connect_to_process(request):
+    """Start server to receive connection from the debug server injected into the
+    debuggee process.
+    """
+
+    channels.Channels().accept_connection_from_server(
+        ("127.0.0.1", 0),
+        before_accept=lambda address: _parse_request_and_inject(request, address),
+    )
+
+
+def _parse_request_and_inject(request, address):
+    cmdline = [sys.executable]
+
+    host, port = address
+    ptvsd_args = request("ptvsdArgs", json.array(unicode))
+    cmdline += [
+        compat.filename(ptvsd.__main__.__file__),
+        "--client",
+        "--host",
+        host,
+        "--port",
+        str(port),
+        "--pid",
+        str(request("processId", int))
+    ] + ptvsd_args
+
+    try:
+        # This process will immediately exit after injecting debug server
+        subprocess.Popen(
+            cmdline,
+            bufsize=0,
+        )
+    except Exception as exc:
+        raise request.cant_handle(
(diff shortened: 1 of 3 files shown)


karthiknadig:
There seems to be some issue with attaching to pid:
```
Connecting to 64 bits target
Injecting dll
Dll injected
Allocating code in target process
Writing code in target process
Allocating return value memory in target process
Injecting code to target process
Waiting for code to complete
Error when injecting code in target process. Error code: 3 (on windows)
```
This is independent of this change.

src/ptvsd/adapter/debuggee.py
```diff
@@ -80,6 +80,47 @@ def spawn_and_connect(request):
             before_accept=lambda address: _parse_request_and_spawn(request, address),
         )
 
+def connect_to_process(request):
```
int19h: Let's call it `attach_by_pid` or something similar - in line with whatever is being used by the tests currently. As is, it's a bit unclear that either this function, or the worker below, uses "processId" to inject.

Also, looks like `_parse_request_and_inject` could be nested?

src/ptvsd/adapter/debuggee.py
```diff
@@ -80,6 +80,47 @@ def spawn_and_connect(request):
...
+    cmdline += [
+        compat.filename(ptvsd.__main__.__file__),
+        "--client",
+        "--host",
+        host,
+        "--port",
+        str(port),
+        "--pid",
+        str(request("processId", int))
+    ] + ptvsd_args
```
int19h: Let's always log the command line here before doing `Popen`, and then skip it in the error message.

src/ptvsd/adapter/debuggee.py
```diff
@@ -80,6 +80,47 @@ def spawn_and_connect(request):
...
+        subprocess.Popen(
+            cmdline,
+            bufsize=0,
+        )
+    except Exception as exc:
+        raise request.cant_handle(
+            "Error launching debug process: {0}\n\nCommand line:{1!r}",
+            exc,
+            cmdline,
+        )
```
int19h: This should also check `returncode` in case `Popen` succeeded, but ptvsd itself failed. It will still detect that via timeout on the server listener eventually, but a `returncode` check will be faster.

fabioz:
>  seems to be some issue with attaching to pid:
> 
> ```
> Connecting to 64 bits target
> Injecting dll
> Dll injected
> Allocating code in target process
> Writing code in target process
> Allocating return value memory in target process
> Injecting code to target process
> Waiting for code to complete
> Error when injecting code in target process. Error code: 3 (on windows)
> ```

I think that this should be fixed when we merge `master` > `dbg_adapter_refactor` (the attach to pid had many improvements/fixes which are still not incorporated in the refactor branch).

PR #1702, opened 2019-08-19
Autoreload in CherryPy no longer blocked by pydevd threads. Fixes #1691

### LINKED ISSUE #1691: Waiting for thread pydevd.CheckAliveThread.
## Environment data

- PTVSD version: 4.3.2
- OS and version: Ubuntu
- Python version (& distribution if applicable, e.g. Anaconda): 3.*
- Using VS Code or Visual Studio: VSCode
- Library: cherrypy

## Actual behavior
```
[13/Aug/2019:18:30:27] ENGINE Listening for SIGHUP.
[13/Aug/2019:18:30:27] ENGINE Listening for SIGTERM.
[13/Aug/2019:18:30:27] ENGINE Listening for SIGUSR1.
[13/Aug/2019:18:30:27] ENGINE Bus STARTING
CherryPy Checker:
The Application mounted at '' has an empty config.

[13/Aug/2019:18:30:27] ENGINE Started monitor thread 'Autoreloader'.
[13/Aug/2019:18:30:27] ENGINE Serving on http://127.0.0.1:8080
[13/Aug/2019:18:30:27] ENGINE Bus STARTED
127.0.0.1 - - [13/Aug/2019:18:30:43] "GET / HTTP/1.1" 200 12 "" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"
127.0.0.1 - - [13/Aug/2019:18:30:43] "GET /favicon.ico HTTP/1.1" 200 1406 "http://127.0.0.1:8080/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"
[13/Aug/2019:18:30:52] ENGINE Restarting because /mnt/c/GIT/pyscratch2/cherrypysample.py changed.
[13/Aug/2019:18:30:52] ENGINE Stopped thread 'Autoreloader'.
[13/Aug/2019:18:30:52] ENGINE Bus STOPPING
[13/Aug/2019:18:30:53] ENGINE HTTP Server cherrypy._cpwsgi_server.CPWSGIServer(('127.0.0.1', 8080)) shut down
[13/Aug/2019:18:30:53] ENGINE Bus STOPPED
[13/Aug/2019:18:30:53] ENGINE Bus EXITING
[13/Aug/2019:18:30:53] ENGINE Bus EXITED
[13/Aug/2019:18:30:53] ENGINE Waiting for child threads to terminate...
[13/Aug/2019:18:30:53] ENGINE Waiting for thread pydevd.CheckAliveThread.
```

## Expected  behavior

Should not block on debugger.

## Steps to reproduce:
1. Use this file:
```py
import cherrypy

class HelloWorld(object):
    @cherrypy.expose
    def index(self):
        return "Hello World!"

cherrypy.quickstart(HelloWorld())
```
2. Start debugging.
3. Make changes to line with "Hello World" and save.

fabioz:


src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -222,6 +222,12 @@ def wait_pydb_threads_to_finish(self, timeout=0.5):
         pydev_log.debug("The following pydb threads may not have finished correctly: %s",
                         ', '.join([t.getName() for t in pydb_daemon_threads if t is not self]))
 
+    def join(self, timeout=None):
+        # If someone tries to join this thread, mark it to be killed.
+        # This is the case for CherryPy when auto-reload it turned on.
```
webknjaz:
```suggestion
        # This is the case for CherryPy when auto-reload is turned on.
```

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -2159,6 +2161,135 @@ def test_source_mapping(case_setup):
...
+    a new process connection (everything is the same except that the
+    existing process is stopped).
+    '''
+    port = get_free_port()
+    # We write a temp file because we'll change it to autoreload later on.
+    f = tmpdir.join('_debugger_case_cherrypy.py')
+
+    tmplt = '''
+import cherrypy
+cherrypy.config.update({'server.socket_port': %(port)s})
```
webknjaz: Explicit is better than implicit + disable checker so that you won't have to filter out logs.
```suggestion
cherrypy.config.update({
    'checker.on': False,
    'engine.autoreload.on': False,
    'server.socket_port': %(port)s,
})
```

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -2159,6 +2161,135 @@ def test_source_mapping(case_setup):
...
+        contents = t.wait_for_contents()
+        assert 'Hello World INITIAL!' in contents
+
+        # Sleep a bit more to make sure that the initial timestamp was gotten in the
+        # CherryPy background thread.
+        time.sleep(2)
+        f.write(tmplt % dict(port=port, str='NEW'))
+
+        secondary_process_thread_communication.join(10)
+        if secondary_process_thread_communication.isAlive():
```
webknjaz: camelcased version of this method is deprecated
```suggestion
        if secondary_process_thread_communication.is_alive():
```

src/ptvsd/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -2159,6 +2161,135 @@ def test_source_mapping(case_setup):
...
 
+@pytest.mark.skipif(not TEST_CHERRYPY, reason='No CherryPy available')
+def test_process_autoreload_cherrypy(case_setup_multiprocessing, tmpdir):
+    '''
+    CherryPy does an os.execv(...) which will kill the running process and replace
+    it with a new process when a reload takes place, so, it mostly works as
+    a new process connection (everything is the same except that the
+    existing process is stopped).
+    '''
+    port = get_free_port()
```
webknjaz: Could just use ephemeral port 0 and parse the real one from console output...

fabioz: That's mostly what `get_free_port()` does... I know there's a (really) small chance that some other process uses the same port from that call before the server actually uses it, but I think is negligible enough to keep as is (and it'd only affect tests).

src/ptvsd/wrapper.py
```diff
@@ -1291,6 +1291,10 @@ def on_pydevd_thread_resume_single_notification(self, seq, args):
             # https://github.com/microsoft/ptvsd/issues/1530.
             self.send_event('continued', **body)
 
+    @pydevd_events.handler(pydevd_comm.CMD_EXIT)
+    def on_pydevd_cmd_exit(self, seq, args):
```
karthiknadig: @int19h do we need to specifically handle this case in the new adapter?

int19h: I'm not entirely sure what this even is. The handler does nothing, so it's here because this notification gets received in that scenario, and it wasn't received previously. 

In any case, we do handle DAP "exited", and we do allow unknown events (we log them as errors, but continue running), so nothing should break - and if anything is missing, we'll find out quickly enough.

fabioz: This may be sent from the pydevd backend when it's about to exit... 

I'm reviewing the code and it's added as an `atexit` handler to add that message, but it kills all the threads right after adding that message, so, in practice it's only sent very seldomly (I added it  because a test failed due to that).

Thinking about it a bit more, I think what I did is not correct, this message should not even be sent when using the DAP backend. I've just provided a fix for this.

PR #1709, opened 2019-08-21
Merge master > dbg_adapter_refactor

fabioz:
Note: did a merge commit which only merged the pydevd bits from the master to the refactor branch (changes on tests from ptvsd need to be manually checked later on for changes).

I did a manual test with VSCode just to check if the debugger starts and reaches a breakpoint (which seems to be working).


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -373,9 +373,9 @@ def _on_run(self):
 
                             return  # break if queue is empty and killReceived
                         else:
-                            if time is None:
-                                break  # interpreter shutdown
-                            time.sleep(self.timeout)
+                            # if time is None: -- TODO: Check why this was done (commented out for now).
```
fabioz: As a note, I commented this code out which was added in the branch (the cmd queue already has a 0.1 waiting time, so, this `time.sleep()` seemed redundant and would slow the reader thread).

PR #1747, opened 2019-09-04
Fix for attach to pid tests

### LINKED ISSUE #1737: Refactor branch: Restore attach to pid tests

karthiknadig:
Fixes #1737

### DIFF
--- a/src/ptvsd/common/attach_script_ptvsd_pid.py
+++ b/src/ptvsd/common/attach_script_ptvsd_pid.py
@@ -1,0 +1,78 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+import os.path
+
+
+__file__ = os.path.abspath(__file__)
+_ptvsd_dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+
+
+def attach(port, host, client, log_dir):
+    try:
+        import sys
+        if 'threading' not in sys.modules:
+            try:
+
+                def on_warn(msg):
+                    from ptvsd.commom import log
+                    log.warn(msg)
+
+                def on_exception(msg):
+                    from ptvsd.common import log
+                    log.exception(msg)
+
+                def on_critical(msg):
+                    from ptvsd.common import log
+                    log.error(msg)
+
+                pydevd_attach_to_process_path = os.path.join(
+                    _ptvsd_dir,
+                    'ptvsd',
+                    '_vendored',
+                    'pydevd',
+                    'pydevd_attach_to_process')
+                assert os.path.exists(pydevd_attach_to_process_path)
+                sys.path.append(pydevd_attach_to_process_path)
+
+                # Note that it's not a part of the pydevd PYTHONPATH
+                import attach_script
+                attach_script.fix_main_thread_id(
+                    on_warn=on_warn, on_exception=on_exception, on_critical=on_critical)
+            except:
+                from ptvsd.common import log
+                log.exception()
+
+        if not log_dir:
+            log_dir = None
+
+
+        
+        sys.path.insert(0, _ptvsd_dir)
+        import ptvsd
+        assert sys.path[0] == _ptvsd_dir
+        del sys.path[0]
+
+        from ptvsd.common import options as common_opts
+        from ptvsd.server import options
+        common_opts.log_dir = log_dir
+        options.client = client
+        options.host = host
+        options.port = port
+
+        from ptvsd.common import log
+        log.to_file()
+        log.info("Debugger injection begin")
+
+        if client:
+            ptvsd.attach((host, port))
+        else:
+            ptvsd.enable_attach((host, port))
+
+        log.info("Debugger successfully injected")
+    except:
+        import traceback
+        traceback.print_exc()
--- a/src/ptvsd/server/main.py
+++ b/src/ptvsd/server/main.py
@@ -276,57 +276,56 @@ def attach_to_pid():
     if log_dir is None:
         log_dir = ""
 
-    attach_script_ptvsd_pid_dirname = os.path.join(os.path.dirname(ptvsd.__file__), 'common')
-    assert os.path.exists(attach_script_ptvsd_pid_dirname)
-
-
-    log_dir = log_dir.replace('\\', '/')
-    setup = {'host': host, 'port': port, 'client': client, 'log_dir': log_dir, 'pid': pid}
-
-    # We need to be able to import attach_script_ptvsd_pid without importing ptvsd first.
-    if sys.platform == 'win32':
-        setup['pythonpath'] = attach_script_ptvsd_pid_dirname.replace('\\', '/')
-
-        python_code = '''import sys;
-sys.path.insert(0, "%(pythonpath)s");
-import attach_script_ptvsd_pid;
-del sys.path[0];
-attach_script_ptvsd_pid.attach(port=%(port)s, host="%(host)s", client=%(client)s, log_dir="%(log_dir)s");
-'''.replace('\r\n', '').replace('\r', '').replace('\n', '')
-    else:
-        setup['pythonpath'] = attach_script_ptvsd_pid_dirname
-
-        # We have to pass it a bit differently for gdb
-        python_code = '''import sys;
-sys.path.insert(0, \\\"%(pythonpath)s\\\");
-import attach_script_ptvsd_pid;
-del sys.path[0];
-attach_script_ptvsd_pid.attach(port=%(port)s, host=\\\"%(host)s\\\", client=%(client)s, log_dir=\\\"%(log_dir)s\\\");
-'''.replace('\r\n', '').replace('\r', '').replace('\n', '')
-
-    python_code = python_code % setup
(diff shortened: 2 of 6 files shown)


src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
int19h: Since it's already inside the `ptvsd` package, the name seems redundant. Maybe something like `attach_pid_injected.py`?

Also, I don't think this should be under `ptvsd.common` - it's the implementation detail of attach-by-PID code that is inside the server specifically, and nothing else can use it for any other purpose, so it should be in `ptvsd.server` as well.

src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
...
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+import os.path
```
int19h: `import os` is enough - `os.path` is not really a subpackage, and it's documented as such.

src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
...
+                    from ptvsd.commom import log
+                    log.warn(msg)
+
+                def on_exception(msg):
+                    from ptvsd.common import log
+                    log.exception(msg)
+
+                def on_critical(msg):
+                    from ptvsd.common import log
+                    log.error(msg)
```
int19h: `ptvsd` is not loaded here because we need to do the `sys.path` dance first. If these get called by `fix_main_thread_id()`, it would also be the case, so I don't think trying to load it deferred is going to help. At this point, the best you can do is print to stderr.

src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
...
+                assert os.path.exists(pydevd_attach_to_process_path)
+                sys.path.append(pydevd_attach_to_process_path)
+
+                # Note that it's not a part of the pydevd PYTHONPATH
+                import attach_script
+                attach_script.fix_main_thread_id(
+                    on_warn=on_warn, on_exception=on_exception, on_critical=on_critical)
+            except:
+                from ptvsd.common import log
+                log.exception()
```
int19h: Same thing here about logging. But also, shouldn't it be re-raising?

src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if not log_dir:
+            log_dir = None
```
int19h: This could be the default value for the argument.

src/ptvsd/common/attach_script_ptvsd_pid.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if client:
+            ptvsd.attach((host, port))
+        else:
+            ptvsd.enable_attach((host, port))
+
+        log.info("Debugger successfully injected")
+    except:
+        import traceback
+        traceback.print_exc()
```
int19h: Since we do have logging at this point, it might as well do `raise log.exception()` here. Python will print the traceback itself if the exception escapes on a background thread; and it doesn't kill the process.

src/ptvsd/server/main.py
```diff
@@ -266,62 +266,68 @@ def run_code():
...
+        python_code = '''import sys;
+sys.path.insert(0, "%(pythonpath)s");
+import attach_script_ptvsd_pid;
+del sys.path[0];
+attach_script_ptvsd_pid.attach(port=%(port)s, host="%(host)s", client=%(client)s, log_dir="%(log_dir)s");
+'''.replace('\r\n', '').replace('\r', '').replace('\n', '')
+    else:
+        setup['pythonpath'] = attach_script_ptvsd_pid_dirname
+
+        # We have to pass it a bit differently for gdb
```
int19h: It looks like the only difference here is escaping of quotes? If so, it would be better to have a single snippet, and just add backslashes if it's gdb. Or see if we can find out a way to avoid it - e.g. will single quotes work for both?

src/ptvsd/server/main.py
```diff
@@ -266,62 +266,68 @@ def run_code():
...
-debug = ptvsd.attach if {client} else ptvsd.enable_attach
-debug(({host}, {port}), multiprocess={multiprocess})
-""".format(
-        **locals()
-    )
+        python_code = '''import sys;
+sys.path.insert(0, "%(pythonpath)s");
+import attach_script_ptvsd_pid;
+del sys.path[0];
+attach_script_ptvsd_pid.attach(port=%(port)s, host="%(host)s", client=%(client)s, log_dir="%(log_dir)s");
```
int19h: It feels like this could be simplified if it were to use `execfile()` instead of importing the script. Then you don't need to change `sys.path` at all - just use the full path directly to the script. Values can be passed via `globals=`. Then this whole thing becomes a single-liner:
```py
execfile(..., globals={...})
```
Note that it's called `exec()` in Python 3. This distinction can be handled in `compat`, so that here it's just `compat.execfile()` (can't be `exec()`, because in Python 2 that's a keyword).

int19h: Another thing - consider how this code handles nested backslashes and quotes in `log_dir`. 

Then there's the issue of string literals, and what happens if e.g. the process is running Python 2, but the code performing the injection is Python 3, and vice versa. Suppose `log_dir` has Unicode characters in it, for example - I don't think it'll survive the trip.

That was why I had that `quote_str()` thing... it takes care of it by making sure that literals in injected code are always explicitly Unicode, `u"..."`, and using `\u` escapes for anything other than ASCII inside, and also for all nested quotes and backslashes.

src/ptvsd/server/main.py
int19h: Given that you'll probably need to resurrect `quote_str()`, it sounds like it would be easiest to just add that check there for the initial and the final `"` (and it will escape any inner ones).

src/ptvsd/server/main.py
```diff
@@ -266,62 +266,68 @@ def run_code():
...
+
+        # We have to pass it a bit differently for gdb
+        python_code = '''import sys;
+sys.path.insert(0, \\\"%(pythonpath)s\\\");
+import attach_script_ptvsd_pid;
+del sys.path[0];
+attach_script_ptvsd_pid.attach(port=%(port)s, host=\\\"%(host)s\\\", client=%(client)s, log_dir=\\\"%(log_dir)s\\\");
+'''.replace('\r\n', '').replace('\r', '').replace('\n', '')
+
+    python_code = python_code % setup
```
int19h: You can use `format()` here as well for a bit more readable input - `python_code.format(**setup)` will do the trick.

tests/start_methods.py
```diff
@@ -403,15 +403,18 @@ def configure(self, run_as, target, **kwargs):
...
+
+        if self.method == "attach_pid":
+            self._attach_args["processId"] = pid
+        else: 
+            connected = False
+            while connected is False:
+                time.sleep(0.1)
+                connections = psutil.net_connections()
+                connected = (
+                    len(list(p for (_, _, _, _, _, _, p) in connections if p == pid)) > 0
```
int19h: `any(p == pid for ... in connections)`

src/ptvsd/server/main.py
int19h: Yet another possibility - encode strings as lists of ints, each representing an UTF-8 byte:
```py
s = [ord(ch) for ch in s.encode("utf-8")]
```
That way, the injected snippet could have no quotes in it at all. On the receiving side (inside the attach script file), you can then do this:
```
s = bytearray(s).decode("utf-8")
```

src/ptvsd/server/main.py
karthiknadig: For `compat.execfile` you would still need to load `compat` from `ptvsd.common`. So the `sys.path` dance cannot be avoided. how about this:
```py
import sys
_execfile = execfile if sys.version_info < (3,) else exec
_execfile(<injector-path>, globals={...})
```

src/ptvsd/common/attach_script_ptvsd_pid.py
fabioz: I was thinking about this and this whole code is pretty much a copy of `pydevd_attach_to_process/attach_pydevd.py` and `pydevd_attach_to_process/attach_script.py`... if multiprocessing will work with the plain `pydevd` attach later on, maybe this code shouldn't be needed (we may need to have multiprocessing done first so that this works though).

i.e.: it could use `pydevd_attach_to_process/attach_pydevd.main()` just passing the proper `port` and `pid` to connect to an adapter that'd be waiting for the connection and then do the same handshake that `multiprocessing` will do (right now it doesn't have the mode to wait for an incoming connection, but that should be easy to do).

We don't necessarily need to do this now, but it may be worth doing it in the future so that we don't have to change both places if some change is needed.

int19h: Ah, so we'd use `evaluate` request to load ptvsd into the process here after pydevd does its own injection and connects?

fabioz: Yes ;)

Issue #1713, opened 2019-08-22
adapter: multiple concurrent sessions


int19h:
The adapter process currently has a 1:1 mapping to DAP process, and the associated debug session. For #1706, we need to change this to 1:N.

DAP requires a separate connection for every process / debug session, so we will still have separate communication channels for each one. And for every connection from the IDE to the adapter, there's a corresponding connection from the adapter to the debug server inside the debugged process. Specific requirements for the IDE and the server will be described below.

int19h:
## Connections between the adapter and the IDE

Since stdin/stdout can only be used for one channel at a time, all the extra channels will have to be socket channels, utilizing client access tokens (#1711) for security. The stdio channel is not really necessary at that point, and we could do everything over sockets - but we already have it implemented, and DAP clients other than VS and VSCode might not support connection to the server via sockets. Also the DAP spec still defines stdio as _the_ DAP transport. So it's best to keep it supported.

Thus, for "launch" scenarios, we will keep the current arrangement for single-proc : the IDE launches the adapter directly, and uses stdio for communication. This will also be the case for "launch" that involves multiple processes, but only for the first process; with the child processes, the adapter sends notifications (`ptvsd_subprocess`) about them to the IDE, and the IDE then starts an "attach" debug session for that process. This logic is already implemented in vscode-python.

For "attach", both single-proc and multiproc, there will be no local adapter at all. Instead, the adapter is spawned on the remote side (#1712), and the IDE will connect directly to it over a socket. This will require a small change to `DebugAdapterFactory` shim in vscode-python: it currently always spawns the adapter, but now it will need to route "attach" debug configs via [DebugAdapterServer](https://code.visualstudio.com/api/references/vscode-api#DebugAdapterServer) instead.

We'll build on the existing `debugServer` support in the adapter for socket connections. The existing implementation always listens on 127.0.0.1, and only accepts a single connection before closing the listener. This needs to be changed to listen on the specified host (but still default to 127.0.0.1), and to accept connections repeatedly. It also needs to add support for client access tokens for security reasons.

int19h:
## Connections between the adapter and the debug server (ptvsd/pydevd)

All of these connections are over sockets - the only difference is that instead of having zero or one connection, we need to support zero or more connections. 

[...]

PR #1757, opened 2019-09-10
#1695, #1721, #1722, some of #1713

### LINKED ISSUE #1695: Handle ide, debug server, and no-debug disconnect

### LINKED ISSUE #1721: "runInTerminal" is broken on non-Windows platforms
(dbg_adapter_refactor branch)

On platforms other than Win32, `os.waitpid()` cannot be used to wait on a process that isn't a child process, to obtain its exit code. When the process is launched via "runInTerminal" request, it is a child process of VSCode, not a child process of the adapter. Thus, the adapter cannot track those processes, and cannot detect when they exit in nodebug scenario, since there's no connection.

The launcher needs to keep track of the process instead, and notify the adapter.

### LINKED ISSUE #1722: Output is not captured in "noDebug" with "runInTerminal"
Output capture is done by the adapter when it spawns the process. Since "runInTerminal" has the IDE spawn the process instead, there's no way for the adapter to do output redirection.

Instead, launcher should capture the output, and send it over to the adapter.

int19h:
Fix #1695: Handle IDE, launcher, debug server, and no-debug disconnect

Fix #1721 "runInTerminal" is broken on non-Windows platforms.

Fix #1722: Output is not captured in "noDebug" with "runInTerminal"

Groundwork for #1713: adapter: multiple concurrent sessions

Move "launch" request parsing and debuggee process spawning, PID reporting and tracking, stdio "output" capture, and exit code reporting into launcher. Launcher now communicates to the adapter via a full-fledged message channel.

Refactor adapter. Add an abstraction for a debug session, and treat IDE, launcher, and debug server as separate components managed by that session.

Improve adapter logging to capture information about current debug session, and current message handler if any.

Fix reporting exceptions from message handlers.

Various test fixes.


src/ptvsd/adapter/components.py
```diff
@@ -0,0 +1,166 @@
...
+
+        @functools.wraps(f)
+        def lock_and_handle(self, message):
+            try:
+                with self.session:
+                    return f(self, message)
+            except ComponentNotAvailable as exc:
+                raise message.cant_handle("{0}", exc, silent=True)
+            except messaging.MessageHandlingError as exc:
+                exc.propagate(message)
```
karthiknadig: Should this be `cant_handle` also?

int19h: Not necessarily, because it can also be `InvalidMessageError` (it's a subclass) - `exc.propagate()` will take care of all that.

src/ptvsd/adapter/ide.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+
+    def propagate_after_start(self, event):
+        # pydevd starts sending events as soon as we connect, but the IDE doesn't
+        # expect to see any until it receives the response to "launch" or "attach"
+        # request. If IDE is not ready yet, save the event instead of propagating
+        # it immediately.
+        if self._deferred_events is not None:
+            self._deferred_events.append(event)
+            log.debug("Propagation deferred.")
```
karthiknadig: >log.debug("Propagation deferred.") [](start = 12, length = 34)

Should this have `event.describe()` ?

int19h: It had initially, but it's redundant, since `@message_handler` will put that on top of _every_ log entry coming from within the message handler. Looks like this:
```
D+00000.890: [handling #4 event "process" from Server-1]
             Propagation deferred.

D+00000.890: [handling #3 request "configurationDone" from IDE-1]
             Server-1 <-- {
                 "seq": 4,
                 "type": "request",
                 "command": "configurationDone"
             }
```

src/ptvsd/adapter/server.py
```diff
@@ -0,0 +1,144 @@
...
+            "supportTerminateDebuggee": False,
+            "supportsDelayedStackTraceLoading": False,
+            "supportsLoadedSourcesRequest": False,
+            "supportsLogPoints": False,
+            "supportsTerminateThreadsRequest": False,
+            "supportsSetExpression": False,
+            "supportsTerminateRequest": False,
+            "supportsDataBreakpoints": False,
+            "supportsReadMemoryRequest": False,
+            "supportsDisassembleRequest": False,
```
karthiknadig: nit: sorting
```
"supportsCompletionsRequest": False,
"supportsConditionalBreakpoints": False,
"supportsConfigurationDoneRequest": False,
"supportsDataBreakpoints": False,
"supportsDelayedStackTraceLoading": False,
"supportsDisassembleRequest": False,
"supportsEvaluateForHovers": False,
"supportsExceptionInfoRequest": False,
"supportsExceptionOptions": False,
"supportsFunctionBreakpoints": False,
"supportsGotoTargetsRequest": False,
"supportsHitConditionalBreakpoints": False,
"supportsLoadedSourcesRequest": False,
"supportsLogPoints": False,
"supportsModulesRequest": False,
"supportsReadMemoryRequest": False,
"supportsRestartFrame": False,
"supportsRestartRequest": False,
"supportsSetExpression": False,
"supportsSetVariable": False,
"supportsStepBack": False,
"supportsStepInTargetsRequest": False,
"supportsTerminateDebuggee": False,
"supportsTerminateRequest": False,
"supportsTerminateThreadsRequest": False,
"supportsValueFormattingOptions": False,


```

int19h: The current order follows the order in the spec. But yeah, I don't think there's any particular reason to do that.

src/ptvsd/adapter/server.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @message_handler
+    def continued_event(self, event):
+        # https://github.com/microsoft/ptvsd/issues/1530
+        #
+        # DAP specification says that a step request implies that only the thread on
+        # which that step occurred is resumed for the duration of the step. However,
+        # for VS compatibility, pydevd can operate in a mode that resumes all threads
+        # instead, and it is the default mode for all clients.
+        #
```
karthiknadig:
```suggestion
        #
        # This can be changed using "steppingResumesAllThreads" in launch.json.
        # Default is true. To resume the stepped thread only set it to false. 
        #
```

Issue #1715, opened 2019-08-22
Update debugProtocol.json


karthiknadig:
In the latest version the ID types have been changed from `number`  to `integer` and restricted to (2^31 - 1). Along with some new features.

PR #1762, opened 2019-09-11
Support for terminate and terminateDebuggee. Fixes #1755

### LINKED ISSUE #1755: pydevd does not support "terminate" nor "terminateDebuggee"
Curiously, the "initialize" response claims that it does, although that particular item in it is also misspelled:
```py
            'supportTerminateDebuggee': True,
```
Either way, in practice it doesn't do anything. Thus, there's currently no way to terminate the process when the launcher is not in the picture - e.g. for all attach scenarios.

This is going to be less important once we do #1712, since then the adapter can kill the process by PID in all "attach" scenarios. However, that has a potential race condition, in that the debuggee may exit, and a new process with the same PID can be spawned, by the time the adapter tries to kill it. It would be better if pydevd could self-terminate when asked, and the adapter would then only try kill-by-PID as a fallback if it notices that debuggee is still running after "terminate".

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -803,3 +805,128 @@ def set_source_mapping(self, py_db, source_filename, mapping):
...
+        Terminates the current process (and child processes if the option to also terminate
+        child processes is enabled).
+        '''
+        if not py_db.terminate_child_processes:
+            os._exit(1)
+
+        if IS_WINDOWS:
+            self._terminate_process_and_children_windows()
+        else:
+            self._terminate_process_and_children_linux_and_mac()
```
int19h: My suggestion would be to make these handle the children only (conditional on the flag), and then wrap the call in a `try..finally` that does `os._exit` at the end. This way, if anything goes wrong while processing the subprocess tree, at least the process itself definitely exits. 

The main reason is that the code that deals with children itself spawns another process and then communicates with it, and that can fail in so many ways - e.g. the helper process can spawn successfully but then crash.

karthiknadig:
This fixes #1715 as well.

PR #1774, opened 2019-09-16
Provide a way to peek at DAP messages received by the debug server. Fixes #1765

### LINKED ISSUE #1765: Need a way to peek at DAP messages received by the debug server.
See comment here for one case where we need it:
https://github.com/microsoft/ptvsd/issues/1712#issuecomment-530983842

fabioz:


src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -2070,6 +2070,33 @@ def wait_for_commands(self, globals):
             time.sleep(0.01)
 
 
+class INetCommandsListener(object):
+
+    def before_send(self, net_command):
+        '''
+        Called just before a net command is sent to the IDE.
+
+        :type net_command: _pydevd_bundle.pydevd_net_command._BaseNetCommand
```
int19h: This creates a dependency on `NetCommand`, and any changes to that would break the code.

At least for JSON messages, it feels like it would be better to pass `as_dict` directly instead. It has everything already, including message metadata, and it's just a regular `dict` instance, so the only thing one needs to interpret it is the DAP spec.

int19h:
Wouldn't this only expose messages that are *sent* by the server to the client, rather than received by it?

fabioz:
Hummm... it seems I got confused on who's the debug server from the description, sorry. I'll fix that and change to receive only the dict.

fabioz:
I made it possible to listen to messages sent and received now.

PR #1775, opened 2019-09-16
Fix #1769: Logpoint messages aren't printed for "launch"

### LINKED ISSUE #1769: Logpoint messages aren't printed for "launch"
For breakpoints with "logMessage", when they're hit, pydevd sends an "output" event with `"category": "stdout"`, but doesn't actually print that message to stdout. Since we're now capturing the output directly from the process for launch scenarios, this means that the event from the server is ignored by the adapter on the assumption that launcher will supply more accurate output. In this case, the assumption doesn't hold, but the IDE ends up never receiving that message.

I can see three options to handle this.

1. Send output with `"category": "console"` instead. This shouldn't make any practical difference in the UI - it will show up in the Debug Console, but not in the terminal, same as it is today - but such events will always be propagated by the adapter, since it only relies on the launcher for stdout/stderr.

2. Add a custom property to "output" events that pydevd can use to mark synthetic ones that do not correspond to any actual output. The adapter would then check that property, and always propagate such events, even if the launcher is capturing actual stdout/err.

3. Don't synthesize the events, but rather just print the message directly to stdout. This will make it show in the terminal. 

It seems that the second option isn't really adding anything useful, and the choice between 1 and 3 depends on whether we want logpoint messages to show up in the terminal, or only in Debug Console.

The DAP spec doesn't say anything about this, but so far we've been doing the equivalent of 1. This feels more reasonable, since it doesn't affect the observable behavior of the app; e.g. imagine if it is debugged with its stdout redirected to be parsed by some other app - logpoints being actually printed would mess that up.

int19h:
When using launcher to redirect output, tell pydevd to not perform redirection.

Always propagate "output" events from the server to the IDE.

Fix test_log_point to not assume that logpoint output is strictly ordered wrt regular output.

### DIFF
--- a/src/ptvsd/adapter/ide.py
+++ b/src/ptvsd/adapter/ide.py
@@ -223,6 +223,10 @@ class IDE(components.Component):
 
         self.session.spawn_debuggee(request, sudo, args, console, console_title)
 
+        if "RedirectOutput" in self.session.debug_options:
+            # The launcher is doing output redirection, so we don't need the server.
+            request.arguments["debugOptions"].remove("RedirectOutput")
+
     @_start_message_handler
     def attach_request(self, request):
         if self.session.no_debug:
--- a/src/ptvsd/adapter/server.py
+++ b/src/ptvsd/adapter/server.py
@@ -129,13 +129,6 @@ class Server(components.Component):
             self.ide.propagate_after_start(event)
 
     @message_handler
-    def output_event(self, event):
-        category = event("category", "console")
-        # If there is a launcher, it's handling stdout and stderr.
-        if not self.launcher or category not in ("stdout", "stderr"):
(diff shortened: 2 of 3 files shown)


src/ptvsd/adapter/ide.py
```diff
@@ -223,6 +223,10 @@ def launch_request(self, request):
 
         self.session.spawn_debuggee(request, sudo, args, console, console_title)
 
+        if "RedirectOutput" in self.session.debug_options:
```
karthiknadig: This should also check the plain request for `"redirectOutput": true,`, in the launch/attach json.

int19h: We don't actually support "redirectOutput" as an option anywhere else right now (launcher, pydevd) - it's strictly a debug property.

PR #1784, opened 2019-09-18
enable_attach should spawn adapter

### LINKED ISSUE #1712: ptvsd: enable_attach() should spawn adapter
In attach scenarios, `ptvsd.enable_attach()` is executed directly inside user code inside the process to be debugged, but there's no adapter at that point. Thus, the debug server needs to spawn the adapter in this case, and establish a communication channel with it.

### LINKED ISSUE #1766: Support for enable_attach with port 0
This is needed for DS extension.

### LINKED ISSUE #1783: add_dap_messages_listener fails with PyDB.reader is still not setup
```
  File "test_import.py", line 4, in <module>
    ptvsd.enable_attach(("127.0.0.1", 0))
  File "C:\GIT\ptvsd\src\ptvsd\__init__.py", line 78, in enable_attach
    return api.enable_attach(address, log_dir)
  File "C:\GIT\ptvsd\src\ptvsd\server\api.py", line 97, in debug
    return func(start_patterns, end_patterns)
  File "C:\GIT\ptvsd\src\ptvsd\server\api.py", line 112, in enable_attach
    _DAPMessagesListener.initialize()
  File "C:\GIT\ptvsd\src\ptvsd\server\api.py", line 50, in initialize
    pydevd.add_dap_messages_listener(_dap_listener)
  File "C:\GIT\ptvsd\src\ptvsd\_vendored\pydevd\pydevd.py", line 2111, in add_dap_messages_listener
    raise AssertionError('PyDB.reader is still not setup.')
AssertionError: PyDB.reader is still not setup.
```

karthiknadig:
Fixes #1783
Fixes #1766
Fixes #1712

I made a temp solution to #1783.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -199,18 +199,12 @@ class ReaderThread(PyDBDaemonThread):
         self.setName("pydevd.Reader")
         self.process_net_command = process_net_command
         self.process_net_command_json = PyDevJsonCommandProcessor(self._from_json).process_net_command_json
-        self.global_debugger_holder = GlobalDebuggerHolder
-
-        self._dap_messages_listeners = []
 
     def _from_json(self, json_msg, update_ids_from_dap=False):
         return pydevd_base_schema.from_json(json_msg, update_ids_from_dap, on_dict_loaded=self._on_dict_loaded)
 
-    def add_dap_messages_listener(self, dap_commands_listener):
-        self._dap_messages_listeners.append(dap_commands_listener)
-
     def _on_dict_loaded(self, dct):
-        for listener in self._dap_messages_listeners:
+        for listener in GlobalDebuggerHolder.global_dbg.dap_messages_listeners:
             listener.after_receive(dct)
 
     @overrides(PyDBDaemonThread.do_kill_pydev_thread)
@@ -292,7 +286,7 @@ class ReaderThread(PyDBDaemonThread):
                                 return  # Finished communication.
 
                             # We just received a json message, let's process it.
-                            self.process_net_command_json(self.global_debugger_holder.global_dbg, json_contents)
+                            self.process_net_command_json(GlobalDebuggerHolder.global_dbg, json_contents)
 
                         continue
                     else:
@@ -358,11 +352,6 @@ class WriterThread(PyDBDaemonThread):
         else:
             self.timeout = 0.1
 
-        self._dap_messages_listeners = []
-
-    def add_dap_messages_listener(self, dap_commands_listener):
-        self._dap_messages_listeners.append(dap_commands_listener)
-
     def add_command(self, cmd):
         ''' cmd is NetCommand '''
         if not self.killReceived:  # we don't take new data after everybody die
@@ -401,7 +390,7 @@ class WriterThread(PyDBDaemonThread):
                     return
 
                 if cmd.as_dict is not None:
-                    for listener in self._dap_messages_listeners:
+                    for listener in GlobalDebuggerHolder.global_dbg.dap_messages_listeners:
                         listener.before_send(cmd.as_dict)
 
                 cmd.send(self.sock)
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -73,7 +73,7 @@ from _pydev_imps._pydev_saved_modules import thread
 from _pydev_imps._pydev_saved_modules import threading
 from socket import AF_INET, SOCK_STREAM, SHUT_RD, SHUT_WR, SOL_SOCKET, SO_REUSEADDR, SHUT_RDWR, IPPROTO_TCP
 from _pydevd_bundle.pydevd_constants import (DebugInfoHolder, get_thread_id, IS_WINDOWS, IS_JYTHON,
-    IS_PY2, IS_PY36_OR_GREATER, STATE_RUN, dict_keys, ASYNC_EVAL_TIMEOUT_SEC, GlobalDebuggerHolder,
+    IS_PY2, IS_PY36_OR_GREATER, STATE_RUN, dict_keys, ASYNC_EVAL_TIMEOUT_SEC,
     get_global_debugger, GetGlobalDebugger, set_global_debugger)  # Keep for backward compatibility @UnusedImport
 from _pydev_bundle.pydev_override import overrides
 import weakref
@@ -187,7 +187,7 @@ def run_as_pydevd_daemon_thread(func, *args, **kwargs):
 class ReaderThread(PyDBDaemonThread):
     ''' reader thread reads and dispatches commands in an infinite loop '''
 
-    def __init__(self, sock, terminate_on_socket_close=True):
+    def __init__(self, sock, py_db, terminate_on_socket_close=True):
         assert sock is not None
         from _pydevd_bundle.pydevd_process_net_command_json import PyDevJsonCommandProcessor
         from _pydevd_bundle.pydevd_process_net_command import process_net_command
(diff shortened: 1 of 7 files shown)


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -199,18 +199,12 @@ def __init__(self, sock, terminate_on_socket_close=True):
...
 
     def _from_json(self, json_msg, update_ids_from_dap=False):
         return pydevd_base_schema.from_json(json_msg, update_ids_from_dap, on_dict_loaded=self._on_dict_loaded)
 
-    def add_dap_messages_listener(self, dap_commands_listener):
-        self._dap_messages_listeners.append(dap_commands_listener)
-
     def _on_dict_loaded(self, dct):
-        for listener in self._dap_messages_listeners:
+        for listener in GlobalDebuggerHolder.global_dbg.dap_messages_listeners:
```
fabioz: It should use `self.global_debugger_holder` as was previously in the code... the reason for this is that module globals become `None` during interpreter shutdown, so, we need to keep a reference.

Reading the code it seems that the `PyDB` could actually be passed in the constructor (that would make the dependency more explicit too), so, I think the best here would be to pass `py_db` in both constructors (`ReaderThread` and `WriterThread`), store the reference as `self._py_db` and use that instance instead of `GlobalDebuggerHolder.global_dbg`.

src/ptvsd/server/api.py
```diff
@@ -68,12 +106,44 @@ def debug(address, log_dir=None, multiprocess=True):
...
+        if common_opts.log_dir is not None:
+            adapter_args += ["--log-dir", common_opts.log_dir]
+
+        log.info(
+            "enable_attach() spawning attach-to-PID debugger injector: {0!r}", adapter_args
+        )
+
+        # Adapter life time is expected to be longer than this process,
+        # so never wait on the adapter process
+        subprocess.Popen(adapter_args, bufsize=0)
```
fabioz: If the adapter life is expected to be longer, the `PyDevJsonCommandProcessor.on_terminate_request` shouldn't kill that process when terminate is called -- as right now it'll kill all child processes by default.

i.e.: we may need to add an api to add pids which shouldn't be killed by `pydevd` when terminating and use it here (although this is probably a different task).

src/ptvsd/adapter/__main__.py
```diff
@@ -57,6 +62,26 @@ def _parse_argv(argv):
...
+    )
+
+    parser.add_argument(
+        "--connect-to-port",
+        type=int,
+        nargs="?",
+        default=None,
+        const=5678,
+        metavar="PORT",
+        help="FOR INTERNAL USE ONLY. Connect to local debug server",
```
int19h: `help=argparse.SUPPRESS` will hide it from the output entirely.

src/ptvsd/adapter/__main__.py
```diff
@@ -33,22 +33,27 @@ def main(args):
...
     session.wait_for_completion()
 
 
 def _parse_argv(argv):
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
-        "-d",
-        "--debug-server",
+        "--port",
```
int19h: I think at this point you need to remove `nargs` and `const` below. What they did is allow to specify `--debugServer` without any value, and then `const` would be the implied value. But this doesn't make any sense for `--port` and `--host`.

src/ptvsd/adapter/__main__.py
```diff
@@ -57,6 +62,26 @@ def _parse_argv(argv):
...
+        metavar="HOST",
+        help="start the adapter in debugServer mode on the specified host",
+    )
+
+    parser.add_argument(
+        "--connect-to-port",
+        type=int,
+        nargs="?",
+        default=None,
+        const=5678,
```
int19h: Same thing here wrt `nargs` and `const`.

The switch also needs a better name, since the adapter can connect to a bunch of different things, and this one is for a very specific scenario. How about `--for-server-on-port`?

src/ptvsd/adapter/ide.py
```diff
@@ -232,6 +232,10 @@ def attach_request(self, request):
         if self.session.no_debug:
             raise request.isnt_valid('"noDebug" is not supported for "attach"')
 
+        if self.server is not None:
+            # we are already connected to the debug server
+            return
+
```
int19h: This should go into the first branch below, and the PID branch should check for `self.server is not None` and fail the request if so.

src/ptvsd/adapter/server.py
```diff
@@ -60,6 +61,10 @@ def initialize(self, request):
         request.wait_for_response()
         self.capabilities = self.Capabilities(self, request.response)
 
+    def set_debugger_property(self, **kwargs):
```
int19h: I'd rather not use kwargs for this - you get unfortunate intersection with Python keywords not being valid keys etc. Let's stick to dicts for everything that has to go through JSON as is.

src/ptvsd/adapter/server.py
```diff
@@ -60,6 +61,10 @@ def initialize(self, request):
...
+        request = self.channel.send_request("setDebuggerProperty", arguments=kwargs)
+        request.wait_for_response()
```
int19h: `self.channel.request(...)`

src/ptvsd/server/api.py
```diff
@@ -17,6 +17,44 @@
...
+    @staticmethod
+    def initialize():
+        global _dap_listener
+        if _dap_listener is not None:
+            return
+
+        _dap_listener = _DAPMessagesListener()
+        pydevd.add_dap_messages_listener(_dap_listener)
+
+
```
int19h: This all can go inside `enable_attach()`, since it's its private implementation detail that's not used elsewhere.

Also, we really only support calling `enable_attach()` once inside the process already. It doesn't check for it at the moment, but it probably should do so, and raise an error for subsequent calls. That will then also ensure that only one instance of this object is created and registered by that first `enable_attach()` call, and so it doesn't need to do any checks itself.

To simplify it even further, consider using `Queue` (`from ptvsd.compat import queue`). You can create it as a local variable inside `enable_attach()`, and then have the listener `put()` the port number into it, and the caller `get()` it. Then this class would only need the pydevd callback methods, and nothing else.

src/ptvsd/server/api.py
karthiknadig: Created a task for it: https://github.com/microsoft/ptvsd/issues/1786

src/ptvsd/server/api.py
```diff
@@ -13,10 +13,19 @@
 import ptvsd
 from ptvsd.common import log, options as common_opts
 from ptvsd.server import options as server_opts
+from ptvsd.common.compat import queue
 from _pydevd_bundle.pydevd_constants import get_global_debugger
 from pydevd_file_utils import get_abs_path_real_path_and_base_from_file
 
 
+_ADAPTER_START_TIMEOUT = 10
+_dap_listener = None
```
int19h: These two are unused now.

src/ptvsd/server/api.py
```diff
@@ -13,10 +13,19 @@
...
+
+
+class EnableAttachError(Exception):
+    pass
```
int19h: No need for a custom exception type for this, since it's not really meant to be caught. `RuntimeError` with the appropriate message type will do the trick.

src/ptvsd/server/api.py
```diff
@@ -68,12 +77,62 @@ def debug(address, log_dir=None, multiprocess=True):
...
+    port_queue = queue.Queue()
+    class _DAPMessagesListener(pydevd.IDAPMessagesListener):
+        def before_send(self, msg):
+            pass
+
+        def after_receive(self, msg):
+            try:
+                if msg["command"] == "setDebuggerProperty":
+                    port_queue.put(msg["arguments"]["adapterPort"])
+            except:
```
int19h: This should catch specific exception types, if it expects any. `KeyError`?

src/ptvsd/server/api.py
karthiknadig: Renamed one of then to use with `get()`. The other one was unnecessary.

PR #1791, opened 2019-09-20
Add ppid to PydevdProcessInfo

### LINKED ISSUE #1759: pydevd should report parent process ID in "initialize" response
This is mostly to allow us constructing a proper process tree inside the adapter. With the current proposed design, we know that an incoming server->adapter connection is a subprocess, if it's not the first such connection - the root process is always the first, and everything else is its subprocess. The access tokens ensure that there are no spurious connections.

However, in the case where you have child processes spawning other child processes, we don't have any way to determine that relationship.

karthiknadig:
Fixes #1759


karthiknadig:
Had to restore the webhooks. So did a close and reopen to trigger pipelines.

Issue #1787, opened 2019-09-18
Stabilize tests


karthiknadig:
Tracking item for test stabilization.

PR #1803, opened 2019-09-25
#1787

int19h:
Refactor tests.debug to accommodate ptvsd.server spawning the adapter, and remove the need for "custom_client" and "custom_server" start methods.

Fix launcher not propagating debuggee exit code.

Fix attach-by-PID without explicit --log-dir overriding PTVSD_LOG_DIR (and disabling logging).

Improve test logging, with a separate directory for every test.

Various test fixes.


src/ptvsd/adapter/__main__.py
```diff
@@ -20,32 +20,28 @@ def main(args):
...
             session.connect_to_server(("127.0.0.1", args.for_server_on_port))
         with session.accept_connection_from_ide((args.host, args.port)) as (_, port):
-            try:
-                session.server.set_debugger_property({"adapterPort": port})
-            except AttributeError:
-                pass
+            if session.server:
+                try:
+                    session.server.set_debugger_property({"adapterPort": port})
+                except AttributeError:
```
karthiknadig: The `AttributeError` here was to handle case when `server` was not set. If it guaranteed to be set we don't need that.

src/ptvsd/adapter/__main__.py
```diff
@@ -76,10 +72,6 @@ def _parse_argv(argv):
         help=argparse.SUPPRESS
     )
 
-    parser.add_argument(
-        "--cls", action="store_true", help="clear screen before starting the debuggee"
```
karthiknadig: See this: https://www.reddit.com/r/vscode/comments/d6oggo/how_do_i_clear_the_terminal_screen_upon_executing/ 

We may have a case for `--cls`. But handle it via launcher? or this could be a debug config.

src/ptvsd/adapter/ide.py
```diff
@@ -208,6 +208,8 @@ def launch_request(self, request):
             args = ["-c"] + request(
                 "code", json.array(unicode, vectorize=True, size=(1,))
             )
+        else:
+            args = []
```
karthiknadig: Should we fail here? if it is not a file/dir, module, or string.

tests/debug/runners.py
```diff
@@ -0,0 +1,234 @@
...
+The function is exposed directly as a method on Session, with the session argument
+becoming self.
+
+The function is also exposed as a Runner object from this module. Runner objects are
+callable, and invoke the wrapped function when called, but in addition, they can also
+be bound to specific arguments, by using either [] or with_options(), which can be
+chained arbitrarily::
+
+    # Direct invocation:
+    session.attach_by_socket(cli, log_dir="...")
```
karthiknadig:
```suggestion
    session.attach_by_socket("cli", log_dir="...")
```

src/ptvsd/adapter/ide.py
int19h: Proper handling of args, including validation, is done in launcher - this bit here is solely to generate the (dummy) args to launcher to show up in the terminal. So the part here needs to only do enough to get to the launcher - I'd rather not duplicate the validation logic.

src/ptvsd/adapter/__main__.py
int19h: Yeah, for user scenarios like that, this is definitely something that the launcher should do, triggered by a boolean flag in debug config. In the adapter, it was solely for testing the adapter itself, and even then I ended up not using it much.

src/ptvsd/adapter/__main__.py
int19h: Yep, it's always guaranteed to be set, and not `None`- it's either the actual `Server` instance, or else `Missing` (which is "false", so this is really what it's testing for).

PR #1806, opened 2019-09-27
Fix tests

karthiknadig:


tests/ptvsd/server/test_evaluate.py
```diff
@@ -20,10 +21,10 @@ def code_to_debug():
         c = 3
         print([a, b, c])  # @bp
 
-    with debug.Session(start_method) as session:
-        session.configure(run_as, code_to_debug)
-        session.set_breakpoints(code_to_debug, [code_to_debug.lines["bp"]])
-        session.start_debugging()
+    with debug.Session() as session:
+        with run(session, target(code_to_debug)):
+            session.set_breakpoints(code_to_debug, [code_to_debug.lines["bp"]])
```
int19h: This can use `set_breakpoints(..., all)` for brevity.

tests/ptvsd/server/test_flask.py
```diff
@@ -29,56 +29,54 @@ class lines:
...
-    "start_method", [runners.launch, runners.attach_by_socket["cli"]]
-)
+@pytest.fixture
+@pytest.mark.parametrize("run", [runners.launch, runners.attach_by_socket["cli"]])
+def start_flask(run):
+    def start(session, multiprocess=False):
+        if multiprocess:
+            pytest.skip("https://github.com/microsoft/ptvsd/issues/1706")
+
+        env = {
```
int19h: It doesn't make a difference in this case, but it's generally better to access `session.config.env` directly rather than setting it via `config["env"]`, because you get some extra features (like PYTHONPATH handling) there. So it might be best to do it in all tests for consistency, as they get copy/pasted around a lot when writing new ones.

tests/ptvsd/server/test_flask.py
```diff
@@ -29,56 +29,54 @@ class lines:
...
+            "FLASK_APP": paths.app_py,
+            "FLASK_ENV": "development",
+            "FLASK_DEBUG": "1" if multiprocess else "0",
+        }
+        if platform.system() != "Windows":
+            locale = "en_US.utf8" if platform.system() == "Linux" else "en_US.UTF-8"
+            env.update({"LC_ALL": locale, "LANG": locale})
+
+        # No clean way to kill Flask server, expect non-zero exit code
+        session.expected_exit_code = some.int
```
int19h: Let's move this bit down after `args` and before `run`, so that it's not in the middle of editing `session.config`.

tests/ptvsd/server/test_flask.py
int19h: (and can you also do this to `test_django`?)

karthiknadig: We don't really use env in django. Only flask requires FLASK_APP to be set via env.

karthiknadig: I think we should entirely block people from using `config["env"]`. Essentially using that should fail with `Use config.env instead.` witha comment explaining why.

int19h:
:shipit:

Issue #1813, opened 2019-09-30
Occasional missing response to "configurationDone"


int19h:
Sometimes, when receiving "configurationDone", and starting to execute user code, if the script exits almost immediately, the adapter closes the connection before the response to "configurationDone" is actually delivered to the client. 

I've only seen it on CI Linux and macOS runs, and it's very rare, so it's hard to diagnose. But I suspect that this is another manifestation of pydevd messaging being asynchronous (#1699) - basically, if the process exits too fast, pydevd simply doesn't deliver the response, or delivers a partial response. The adapter then treats it as a failed response from the server, and in turn responds to IDE's "configurationDone" with failure "Server-1 disconnected unexpectedly".

Part of the problem may be that `on_configurationdone_request()` in `pydevd_process_net_command_json.py` invokes `api.run()` _before_ sending the response. Perhaps that needs to be re-ordered?

int19h:
Sample failure from a test run:
```
_ test_vsc_exception_options_raise_with_except[module-attach_by_socket(api)-uncaught] _
[gw3] linux -- Python 3.7.1 /home/travis/build/microsoft/ptvsd/.tox/py37-cov/bin/python
pyfile = <function pyfile.<locals>.factory at 0x7fc11c0ff378>
target = <class 'tests.debug.targets.Module'>, run = attach_by_socket(api)
raised = '', uncaught = 'uncaught'
    @pytest.mark.parametrize("raised", ["raised", ""])
    @pytest.mark.parametrize("uncaught", ["uncaught", ""])
    def test_vsc_exception_options_raise_with_except(
        pyfile, target, run, raised, uncaught
    ):
        @pyfile
        def code_to_debug():
            import debug_me  # noqa
    
            def raise_with_except():
                try:
                    raise ArithmeticError("bad code")  # @exc
                except Exception:
                    pass
    
            raise_with_except()
    
        with debug.Session() as session:
            session.expected_exit_code = some.int
            with run(session, target(code_to_debug)):
                session.request(
>                   "setExceptionBreakpoints", {"filters": list({raised, uncaught} - {""})}
                )
tests/ptvsd/server/test_exception.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/debug/session.pym:548: in __exit__
    self.session.request("configurationDone")
[...]

PR #1814, opened 2019-09-30
Make it possible to reconnect on attach to pid. Fixes #1760

### LINKED ISSUE #1760: Disconnect and reconnect via pid should work when using --client mode
Currently you can attach to a process by pid. But detach and re-attach fails. when using --client.

fabioz:
Note that in this request I've also done changes to improve the shutdown procedure to avoid abruptly closing the socket -- although note that this should only improve the situation, not fix it completely (i.e.: things like https://github.com/microsoft/ptvsd/issues/1813 can still happen -- this wasn't really the main point of this pull request).

I'll take a look at https://github.com/microsoft/ptvsd/issues/1813 afterwards.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydev_bundle/pydev_console_utils.py
+++ b/src/ptvsd/_vendored/pydevd/_pydev_bundle/pydev_console_utils.py
@@ -126,6 +126,10 @@ class DebugConsoleStdIn(BaseStdIn):
             py_db = self._py_db
             if py_db is None:
                 py_db = get_global_debugger()
+
+            if py_db is None:
+                return
+
             cmd = py_db.cmd_factory.make_input_requested_message(is_started)
             py_db.writer.add_command(cmd)
         except Exception:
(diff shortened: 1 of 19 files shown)


fabioz:
As a note, there is a copy of the attach to pid script in `ptvsd` and I wasn't able to get the attach to pid working in the refactor branch locally with `--client` and it's also not properly tested in `ptvsd`, so, it may need some more work to fix the portion related to `ptvsd` to actually consider #1760 fixed (this pull request makes sure that it works in plain `pydevd` and makes it possible for it to work in `ptvsd`, but some more work is needed in `ptvsd`).

The choices to make it work in `ptvsd` would be either fixing the way the attach is done in its private attach script copy or porting to use the same scripts used in `pydevd` for the attach (i.e.: https://github.com/microsoft/ptvsd/issues/1815).

karthiknadig:
I created a separate item to address the issue on the debugger side. We can for now consider #1760 to be done.

fabioz:
I'm taking a look at the failure in `test_break_api` to be able to merge this after changing the target to `master`.

Issue #1819, opened 2019-10-01
Capturing debuggee output in tests is broken on py2.7


int19h:
It appears that reading from `Popen.stdout` and `.stderr` on Python 2.7 does not handle partial reads as expected - it waits for the exact number of bytes specified, instead of returning as soon as there's any data to read. Since `tests.debug.output` reads in chunks of 0x1000, this means that output doesn't actually get captured until the debuggee prints that much data, or exits and closes its end of the pipe.

This problem also occurred in the launcher, and is solved there by working directly with file descriptors using `os.read()`, rather than relying on `subprocess.PIPE`. The tests should use the same approach.

This issue is what breaks `test_wait_on_normal_exit_enabled` on 2.7 at the moment on all platforms, because the test doesn't "see" the "Press Enter to continue" message. Other output tests generally look at output only after debuggee exits, and are not affected, but this can cause issues in the future.

PR #1820, opened 2019-10-01
Final batch of test fixes / disabling before merging with master

int19h:
With these, all remaining failures are sporadic.

### DIFF
--- a/tests/ptvsd/server/test_start_stop.py
+++ b/tests/ptvsd/server/test_start_stop.py
@@ -4,7 +4,6 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-import platform
 import pytest
 import sys
 import time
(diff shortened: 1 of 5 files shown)


tests/ptvsd/server/test_start_stop.py
```diff
@@ -58,8 +62,7 @@ def code_to_debug():
 
 
 @pytest.mark.skipif(
-    sys.version_info < (3, 0) and platform.system() == "Windows",
-    reason="On Windows + Python 2, unable to send key strokes to test.",
+    sys.version_info < (3, 0), reason="https://github.com/microsoft/ptvsd/issues/1819"
```
karthiknadig: Fixing #1819 will not fix the Windows + Py 2.7 issue. It should skip for this separately?

int19h: I changed the press any key code to detect when the terminal is redirected, and use "Press Enter" rather than "Press any key" code path for that. So it should work even on Windows now - the reason why it didn't despite the fix is because of #1819.

https://github.com/microsoft/ptvsd/blob/a2d58ee7c925fe90313b9d28cd771090357a2f9d/src/ptvsd/launcher/debuggee.py#L147-L152

PR #1825, opened 2019-10-03
Updating telemetry and some comments

### LINKED ISSUE #1818: Change wording in warning message for justMyCode
When a user attempts to debug library code (with justMyCode set to 'false'), they see the current message in the Debug Console:

image

We should change the wording to tell users _where_ to set this configuration as well as the exact value they should insert into launch.json i.e. _"justMyCode": False._ Lastly, we should include a learn more link that they can CTRL+click to see our debugger configuration documentation [here](https://code.visualstudio.com/docs/python/debugging#_justmycode).

karthiknadig:
Fixes #1818

### DIFF
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
@@ -357,7 +357,7 @@ class NetCommandFactoryJson(NetCommandFactory):
         msg = 'Frame skipped from debugging during step-in.'
         if py_db.get_use_libraries_filter():
             msg += ('\nNote: may have been skipped because of "justMyCode" option (default == true). '
-                    'Try setting \"justMyCode\": false in the launch.json file.')
+                    'Try setting \"justMyCode\": false in the debug configuration (e.g., launch.json).')
 
         body = OutputEventBody(msg, category='console')
         event = OutputEvent(body)
--- a/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
+++ b/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
@@ -356,7 +356,8 @@ class NetCommandFactoryJson(NetCommandFactory):
     def make_skipped_step_in_because_of_filters(self, py_db, frame):
         msg = 'Frame skipped from debugging during step-in.'
         if py_db.get_use_libraries_filter():
-            msg += '\nNote: may have been skipped because of "justMyCode" option (default == true).'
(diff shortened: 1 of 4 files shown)


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
```diff
@@ -356,7 +356,8 @@ def make_input_requested_message(self, started):
     def make_skipped_step_in_because_of_filters(self, py_db, frame):
         msg = 'Frame skipped from debugging during step-in.'
         if py_db.get_use_libraries_filter():
-            msg += '\nNote: may have been skipped because of "justMyCode" option (default == true).'
+            msg += ('\nNote: may have been skipped because of "justMyCode" option (default == true). '
+                    'Try setting \"justMyCode\": false in the launch.json file.')
```
int19h: launch.json is a VSCode artifact, so if we mention it here, let's phrase it somewhat more generic. Something like:
> Try setting ... in debug configuration (launch.json)

fabioz: Maybe we could store the `clientID` on `PyDB` and then check the `clientID` to create an explicit message for VSCode/other IDEs? (creating this part of the message should be refactored to a method in this case)

karthiknadig: For now i will make this a simple string as suggested by @int19h. We have a separate work item to be more accurate on identifying why the file was skipped. That will require getting different messages based on rules or just-my-code setting.

Anapo14: We may need to circle back around to this. I'm thinking about the case in which a user starts a debugging session from the Debug menu option: Debug > Start Debugging. In this case, a launch.json file isn't created on their behalf. I think having this is a good first 'catch-all' for users who do manage to go through the 'Add Configuration' process.

karthiknadig: Note that from the debugger side we really cannot distinguish between configuration from 'launch.json' or if it is created on the fly. Ideally as I have prescribed before there has to be a mechanism in DAP and VSCode to show these type of potential setting changes in a way that doesn't cause notification overload.

int19h: Going forward, it's likely that we'll need a custom DAP message to report a file being skipped. The client can then look at its debug configuration, and formulate the message appropriately (or even offer to remedy it directly). This would also provide better control over notifications, in accordance with the UX guidelines of the client.

PR #1826, opened 2019-10-03
Finalize test refactoring

int19h:
Move all remaining tests to the new APIs.

Reformat all test and product code with Black (ptvsd.common and ptvsd.server are no longer excluded).


tests/tests/test_patterns.py
```diff
@@ -22,16 +22,26 @@ def log_repr(x):
...
-    {}, {'a': 1, 'b': 2},
+    True,
+    False,
+    0,
+    -1,
+    -1.0,
+    1.23,
+    b"abc",
+    b"abcd",
+    "abc",
```
karthiknadig:
```suggestion
    u"abc",
```

tests/tests/test_patterns.py
```diff
@@ -22,16 +22,26 @@ def log_repr(x):
...
+    "abcd",
```
karthiknadig:
```suggestion
    u"abcd",
```

tests/tests/test_patterns.py
```diff
@@ -113,7 +123,7 @@ def test_in_range():
 
 def test_str():
     log_repr(some.str)
-    assert some.str == u"abc"
+    assert some.str == "abc"
```
karthiknadig:
```suggestion
    assert some.str == u"abc"
```

tests/tests/test_patterns.py
int19h: These are all from Black - since we use `unicode_literals`, the prefix is redundant, and apparently this means that it gets removed.

PR #1828, opened 2019-10-07
Add sequence diagrams for subprocess debugging.

int19h:


doc/Subprocess debugging.md
```diff
@@ -0,0 +1,318 @@
...
+Debuggee_2 ->>+ Debuggee_2: start debug session
+
+Adapter -->>- IDE: respond to "attach"
+
+loop every message between IDE and Debuggee_2
+Note over IDE,Debuggee_2: propagate message
+end
+
+
+Note left of IDE: user stops debugging
```
fabioz: When the user `stops debugging`, if it does a disconnect to the `Debugee_2`, it'll resume the program if it was paused in any breakpoint... is this intended? (if that's the case, the note should probably be `user stops debugging Debugee_2`)

doc/Subprocess debugging.md
```diff
@@ -0,0 +1,318 @@
...
+
+participant IDE
+participant Adapter
+participant Debuggee_1
+participant Debuggee_2
+
+
+Note left of Debuggee_1: user spawns<br/>debuggee
+
+Debuggee_1 ->>+ Adapter: spawn and pass adapter listener port (cmdline)
```
fabioz: Should this diagram already handle https://github.com/microsoft/ptvsd/issues/1830?

doc/Subprocess debugging.md
```diff
@@ -0,0 +1,318 @@
...
+
+Note over Adapter,Debuggee_1: TCP connection is maintained
+
+Adapter -->>- IDE: confirm "disconnect" from Debuggee_1
+
+Note over Adapter: continues running
+
+
+
+Note left of IDE: User re-attaches IDE<br/>(same host/port)
```
fabioz: Would be nice to say what happens if a new subprocess is created in the meanwhile (say a `Debuggee_3`) 

i.e.: we could do an attach to pid for new processes found (or we could just initially ignore those).

doc/Subprocess debugging.md
int19h: The scenario here is that user hits the Stop button in the IDE. VSCode has the notion of a tree of processes in the debug session, and so long as we report parent and children properly, doing that will always disconnect and terminate the entire tree.

There's a separate scenario where they open the process list, and disconnect specifically from that one process. It's not fully captured in this diagram, but the expected behavior is for it to continue running - same as it would if it were a single-process attach scenario.

doc/Subprocess debugging.md
int19h: It originally captured the current state of affairs, so that we can see what exactly is missing - #1830 came out of that. I need to update the diagram for all the issues that we open.

doc/Subprocess debugging.md
int19h: The adapter still remains connected to all the processes, so any new ones will also connect to it as they spawn. Since there's no IDE, the adapter won't report "ptvsd_subprocess" at that point, but it will maintain the connection and the session. When the IDE reconnects later, "ptvsd_subprocess" for every subprocess that the adapter is currently aware of will be sent to the IDE, and it will then open additional connections to the adapter for each.

Similarly, if an existing subprocess exits after detach but before re-attach, the IDE doesn't receive "ptvsd_subprocess" for it when it re-attaches. 

This probably needs a separate diagram to show both clearly.

doc/Subprocess debugging.md
fabioz: Ok, so, it seems that sometimes disconnect should run user code and sometimes it shouldn't.

I've created https://github.com/microsoft/ptvsd/issues/1837 to address this in `pydevd` (we should probably update the diagram to be explicit about that message after we decide how this should be handled so that user code is not run on disconnect).

fabioz: Resolving since it should be discussed in #1837.

Issue #1831, opened 2019-10-08
Use pid for files created in pydevd logging.


fabioz:
Note: consider forking.

PR #1832, opened 2019-10-08
Make sure that messages are sent before shutdown. Fixes #1813

### LINKED ISSUE #1813: Occasional missing response to "configurationDone"

fabioz:
Also fixes deadlock with fork.


src/ptvsd/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -61,6 +62,19 @@ def _is_managed_arg(arg):
 
 
 def _on_forked_process(setup_tracing=True):
+    pydevd_constants.after_fork()
+    pydevd_constants.DebugInfoHolder.DEBUG_STREAM = NULL
+
+    # TODO: We should use the pid for logging in pydevd to deal with multiple processes.
```
fabioz: Note: I created https://github.com/microsoft/ptvsd/issues/1831 to address this properly.

PR #1833, opened 2019-10-09
Reverse the connection direction used by enable attach internals.

### LINKED ISSUE #1830: Reverse the debug server internal connection when using enable_attach
This is needed to ensure that debug server always connects to the adapter. Currently `enable_attach` goes against the grain that it creates the server that the adapter connects to. We want to have a unified interaction between debug server and the adapter.

karthiknadig:
Fixes #1830

### DIFF
--- a/src/ptvsd/_vendored/pydevd/pydevd.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd.py
@@ -2253,6 +2253,7 @@ def _enable_attach(
     patch_multiprocessing=False,
     access_token=None,
     ide_access_token=None,
+    as_client=False
     ):
     '''
     Starts accepting connections at the given host/port. The debugger will not be initialized nor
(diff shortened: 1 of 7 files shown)


src/ptvsd/_vendored/pydevd/pydevd.py
```diff
@@ -2253,6 +2253,7 @@ def _enable_attach(
     patch_multiprocessing=False,
     access_token=None,
     ide_access_token=None,
+    as_client=False
```
fabioz: I know this is already merged, but I don't think we should do this change on `pydevd`... the API to connect as a client is calling `settrace` directly with the proper host/port, whereas `_enable_attach()` was really only meant to be called to act as a server and wait until a connection was done, so, I think the changes on `pydevd` should be reverted and `ptvsd` should just use `settrace` directly here (the code related to the checking the `port` is also not interesting in this case... if it's starting a server it should be checked, but if it'll connect to a port in the client it could connect one time to one port and later on connect to another port).

PR #1838, opened 2019-10-09
pydevd authorize request. Fixes #1829

### LINKED ISSUE #1829: "pydevd_authorize" request
We currently exchange the authorization tokens in the "initialize" request and response. 

However, our most recent design for subprocess debugging implies that, while the TCP connection between the subprocess server and the adapter is established immediately as soon as subprocess starts, "initialize" is not sent by the adapter immediately. Instead, the adapter sends "pydevd_systemInfo" request to the debuggee as soon as connection is established, to determine PID and parent PID, and waits for IDE to request attach to the process. When IDE connects to adapter and requests attach, it sends an "initialize" request to the adapter, and that request is then relayed to the server.

Thus, we need to move the token exchange out of "initialize". We can do this via debugger properties, as mentioned in #1782, but this would require adding "getDebuggerProperty". However, this seems to be better suited for a dedicated request, since when authorization is in place, no other messages can be handled until the tokens are validated. Rather than special-casing one property like that, let's refactor this into a distinct request, in which the server token (if present) is sent as part of the request, and the IDE token (if present) is sent as part of the response. 

On pydevd side, this should be very similar to the existing implementation, but instead of special-casing "initialize" in `process_net_command_json()`, it would special-case "pydevd_authorize". Property names are the same - "debugServerAccessToken" in request, "ideAccessToken" in response.

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/_debug_adapter/debugProtocolCustom.json
```diff
@@ -256,6 +256,52 @@
 					"description": "Integer value indicating the bitness of the current process."
 				}
 			}
-		}
+		},
+        "PydevdAuthorizeRequest": {
```
karthiknadig: >         [](start = 0, length = 8)

Tabs

PR #1842, opened 2019-10-10
Use pid for files created in pydevd logging. Fixes #1831

### LINKED ISSUE #1831: Use pid for files created in pydevd logging.

fabioz:


tests/pytest_fixtures.py
```diff
@@ -113,8 +112,8 @@ def with_pydevd_log(request, tmpdir):
     prefix = "pydevd_debug_file-{0}".format(os.getpid())
     filename = tempfile.mktemp(suffix=".log", prefix=prefix, dir=str(tmpdir))
 
-    with pydevd_log.enabled(filename):
-        yield
+    pydevd_log.enable(filename)
```
fabioz: Note: I had to do this change because `pydevd_log.dump` was being called out of the context manager (I could just have put the code below under the context manager, but `pydevd_log.dump` could also be called in a timeout which is also not guaranteed to be called under that context manager).

int19h: We need to think about how to combine this with `--ptvsd-logs` - take a look at `test_wrapper()` above, and `pytest_addoption()` in pytest_hooks.py. As it is, they conflict - if you use `--ptvsd-logs`, then `Session` will set `PYDEVD_DEBUG_FILE` for all processes that it spawns, and that overrides the value set by `pydevd_log`. Then the log dumping code can't find the file where it expects to find it...

It would probably be the easiest to merge this bit with `test_wrapper()`, and then in the latter, make it always log to file - but if `options.log_dir` isn't set (i.e. the switch wasn't used), then log to tmpdir. And then on timeout, we can dump all logs, not just pydevd logs.

fabioz: I changed the pydevd logging logic to be handled in the `test_wrapper` fixture. I didn't change the other logging options related to `ptvsd` itself... I was thinking that maybe in the future we should only have `PTVSD_LOG_DIR` and then `ptvsd` would set the `PYDEVD_DEBUG` and `PYDEVD_DEBUG_FILE` settings accordingly (as the focus on this task is mostly getting the `pydevd` logs to handle multiple processes, I think other changes should be done under a different task).

PR #1855, opened 2019-10-17
Support Python 3.8. Fixes #1845

### LINKED ISSUE #1845: Python 3.8 support (pydevd)
This is for adding ci support for Python 3.8 in `pydevd` (adding it to `ptvsd` should be done after things pass in `pydevd`).

fabioz:


tests/ptvsd/server/test_breakpoints.py
```diff
@@ -328,7 +332,11 @@ def code_to_debug():
             bps = session.set_breakpoints(code_to_debug, bp_markers)
             actual_lines = [bp["line"] for bp in bps]
 
-            expected_markers = ["bp1-expected", "bp2-expected", "bp3-expected"]
+            if sys.version_info >= (3, 8):
+                # See: https://bugs.python.org/issue38508
```
karthiknadig: We will need a issue in our tracker for these to address when the upstream fix is in.

fabioz: Well, the response that I got from https://bugs.python.org/issue38508 is that this is a feature, not a bug, so, it seems this change was intended and will remain this way...

karthiknadig:
I will close and reopen to trigger Azure pipelines.

Issue #1877, opened 2019-10-25
"pydevdSystemInfo" does not report "ppid" on Python 2.7 on Windows


int19h:
`os.getppid()` was not available on Windows until Python 3.2. Pydevd catches any exceptions from that call, and omits the property from JSON, so this doesn't crash the debugger. But because the adapter doesn't receive "ppid" for the incoming server connection, it doesn't know which of the existing debug sessions is the parent, and doesn't send the subprocess notification to any of them. This effectively breaks multiproc debugging on 2.7.

int19h:
Pydevd might have to propagate the parent PID via command line when spawning the child process to ensure that it's always available. Or perhaps there's a way to re-implement `os.getppid()` in 2.7 in a way similar to how it's done in 3.2.

int19h:
The fix doesn't seem to be working reliably. This is from a test run for `test_flask_breakpoint_multiproc[launch(integratedTerminal)]`:
```
D+00001.268: Server[?] <-- {
                 "seq": 1, 
                 "type": "request", 
                 "command": "pydevdSystemInfo"
             }

D+00001.277: Server[?] --> {
                 "pydevd_cmd_id": 502, 
                 "seq": 2, 
                 "type": "response", 
                 "request_seq": 1, 
                 "success": true, 
                 "command": "pydevdSystemInfo", 
                 "body": {
                     "process": {
                         "pid": 128364, 
                         "bitness": 64, 
                         "executable": "c:\\git\\ptvsd\\.tox\\py27\\scripts\\python.exe"
                     }, 
                     "python": {
                         "version": "2.7.16final0", 
                         "implementation": {
                             "description": "CPython", 
                             "version": "", 
                             "name": ""
                         }
                     }, 
                     "platform": {
                         "name": "win32"
                     }
                 }
             }
```
Note the missing "ppid". Looking at pydevd log for the process reporting, it has this entry in it, but no further details:
```
Process32First failed (getting process from CreateToolhelp32Snapshot).
```

fabioz:
I'm investigating.

Issue #1876, opened 2019-10-25
Subprocesses need "pathMappings" in "launch" scenario


int19h:
The "ptvsd_attach" event for subprocesses is generated by the adapter from the request that started the parent debug session. If that request was "attach", then it contains "pathMappings" (or not), and that is propagated accordingly. But if it was "launch", it never contains "pathMappings" - and thus subprocesses don't get any, either. And not having "pathMappings" for "attach" generally breaks things that rely on file paths, such as breakpoints.

Thus, when the parent debug session is "launch", any "ptvsd_attach" event in it should include auto-generated "pathMappings" that properly map local files to themselves via "cwd".

int19h:
After some investigation, it is not clear to me why this is necessary.

If path mappings are explicitly defined for the launch config as follows:
```json5
            "pathMappings": [
                {
                    "localRoot": "",
                    "remoteRoot": "",
                }
            ]
```
then everything works fine. However, from pydevd perspective, this arrangement effectively *disables* path mappings, because it specifically checks for empty strings:
https://github.com/microsoft/ptvsd/blob/f51c96450c99f97bd5f53b93ad049ccf2084e085/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L337-L346

(which is exactly what we want, since there's no need for path mapping at all for local scenarios - all paths are already valid, and they're all absolute due to normalization!)

The reason why it makes a difference is because the *extension* doesn't try to inject its path mapping defaults into the "attach" request, if it sees any existing entries there - even if those entries are effectively no-op, as in this case:

https://github.com/DonJayamanne/pythonVSCode/blob/967f7b1a63dc3211569a6dafbd9be4d0cac95095/src/client/debugger/extension/configuration/resolvers/attach.ts#L111-L116

Note that this specifically injects those entries for localhost, which is exactly where they should *not* be used!

I believe that this is a legacy of the old implementation in ptvsd 4, which handled "pathMappings" itself, and translated it to pre-DAP pydevd path mappings. In that one, we special-cased "launch" to disable "sourceReference" altogether, but for "attach", we returned "sourceReference" unless the path was mapped:
https://github.com/microsoft/ptvsd/blob/14982d90bae7a5f83e8ecbde15c77187e235e83c/ptvsd/wrapper.py#L1554-L1563
[...]

Issue #1875, opened 2019-10-25
Inject ptvsd into subprocesses


int19h:
The adapter should use the "evaluate" request to run `import ptvsd` (after setting up `sys.path`) in subprocesses when the debug server connects to it, so that user code can import the module later and do `ptvsd.break_into_debugger()` etc.

This also needs a test.

Issue #1874, opened 2019-10-25
Subprocesses should inherit PydevdCustomization


int19h:
The crucial one is `DEFAULT_PROTOCOL`. At the moment, pydevd in subprocesses tries to connect to the adapter using the hardcoded default, which is not JSON.

int19h:
This one should be considered a priority, because it's blocking adapter-based multiproc (I worked around it in my branch by changing the hardcoded value directly in pydevd, but I don't want to commit that, obviously).

fabioz:
Sure... it'll be the first thing I'll tackle on Monday ;)

PR #1878, opened 2019-10-28
Fix #1706: Multiprocessing via debug adapter

### LINKED ISSUE #1706: Multiprocessing via debug adapter
The conclusion of the investigation in #1703 is that to deliver the best, most reliable implementation of multiprocessing, we need to switch from the existing design in which the debug server in the root process manages the child processes, to the design in which the debug adapter manages all processes.

In the new design, there is always a single adapter in a multiproc debug session. Debug servers in all processes - both the main process and subprocesses - all connect to that adapter, treating it as a DAP client. On the other side, the IDE establishes multiple connections to the debug adapter, each connection representing a separate process being debugged (at most one of these connections is over stdio, and that only in the "launch" scenario; the rest will be over sockets, using the [DebugAdapterServer](https://code.visualstudio.com/api/references/vscode-api#DebugAdapterServer) API in VSCode).

int19h:
There are a few test failures, but we can track these separately.

The change in pydevd_defaults.py is a workaround for #1874. 

#1877 is preventing this from functioning on Windows; however, test_multiprocessing passes on Python 2.7 on Linux (which also means that fork is working!).

#1875, #1876 are still TBD.

### DIFF
--- a/src/ptvsd/adapter/components.py
+++ b/src/ptvsd/adapter/components.py
@@ -2,11 +2,12 @@
 # Licensed under the MIT License. See LICENSE in the project root
 # for license information.
 
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 import functools
 
 from ptvsd.common import fmt, json, log, messaging, util
+
 
 ACCEPT_CONNECTIONS_TIMEOUT = 10
 
--- a/src/ptvsd/adapter/components.py
+++ b/src/ptvsd/adapter/components.py
@@ -34,19 +34,34 @@ class Component(util.Observable):
     to wait_for() a change caused by another component.
     """
 
-    def __init__(self, session, stream):
+    def __init__(self, session, stream=None, channel=None):
+        assert (stream is None) ^ (channel is None)
+
+        try:
+            lock_held = session.lock.acquire(blocking=False)
+            assert lock_held, "__init__ of a Component subclass must lock its Session"
+        finally:
+            session.lock.release()
+
         super(Component, self).__init__()
 
         self.session = session
-        stream.name = str(self)
-        self.channel = messaging.JsonMessageChannel(stream, self)
+
+        if channel is None:
+            stream.name = str(self)
+            channel = messaging.JsonMessageChannel(stream, self)
+            channel.start()
+        else:
+            channel.name = channel.stream.name = str(self)
+            channel.handlers = self
+        self.channel = channel
         self.is_connected = True
 
-        self.observers += [lambda *_: session.notify_changed()]
-        self.channel.start()
+        # Do this last to avoid triggering useless notifications for assignments above.
+        self.observers += [lambda *_: self.session.notify_changed()]
 
     def __str__(self):
-        return fmt("{0}-{1}", type(self).__name__, self.session.id)
+        return fmt("{0}[{1}]", type(self).__name__, self.session.id)
 
     @property
     def ide(self):
--- a/src/ptvsd/adapter/ide.py
+++ b/src/ptvsd/adapter/ide.py
@@ -4,15 +4,17 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 
+import os
 import platform
+import sys
 
 import ptvsd
-from ptvsd.common import json, log, messaging
+from ptvsd.common import json, log, messaging, sockets
 from ptvsd.common.compat import unicode
-from ptvsd.adapter import components
+from ptvsd.adapter import components, server, session
 
 
-class IDE(components.Component):
+class IDE(components.Component, sockets.ClientConnection):
     """Handles the IDE side of a debug session."""
 
     message_handler = components.Component.message_handler
@@ -33,23 +35,43 @@ class IDE(components.Component):
             "pathFormat": json.enum("path"),  # we don't support "uri"
         }
 
-    def __init__(self, session, stream):
-        super(IDE, self).__init__(session, stream)
+    def __init__(self, sock):
+        if sock == "stdio":
+            log.info("Connecting to IDE over stdio...", self)
+            stream = messaging.JsonIOStream.from_stdio()
+            # Make sure that nothing else tries to interfere with the stdio streams
+            # that are going to be used for DAP communication from now on.
+            sys.stdout = sys.stderr
+            sys.stdin = open(os.devnull, "r")
+        else:
+            stream = messaging.JsonIOStream.from_socket(sock)
 
-        self.client_id = None
-        """ID of the connecting client. This can be 'test' while running tests."""
+        with session.Session() as new_session:
+            super(IDE, self).__init__(new_session, stream)
 
-        self._initialize_request = None
-        """The "initialize" request as received from the IDE, to propagate to the
-        server later."""
+            self.client_id = None
+            """ID of the connecting client. This can be 'test' while running tests."""
 
-        self._deferred_events = []
-        """Deferred events from the launcher and the server that must be propagated
-        only if and when the "launch" or "attach" response is sent.
-        """
+            self.has_started = False
+            """Whether the "launch" or "attach" request was received from the IDE, and
+            fully handled.
+            """
 
-        assert not session.ide
-        session.ide = self
+            self.start_request = None
+            """The "launch" or "attach" request as received from the IDE.
+            """
+
+            self._initialize_request = None
+            """The "initialize" request as received from the IDE, to propagate to the
+            server later."""
+
+            self._deferred_events = []
+            """Deferred events from the launcher and the server that must be propagated
+            only if and when the "launch" or "attach" response is sent.
+            """
+
+            new_session.ide = self
+            new_session.register()
 
         self.channel.send_event(
             "output",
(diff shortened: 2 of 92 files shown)


src/ptvsd/adapter/components.py
```diff
@@ -34,19 +34,29 @@ class Component(util.Observable):
     to wait_for() a change caused by another component.
     """
 
-    def __init__(self, session, stream):
+    def __init__(self, session, stream=None, channel=None):
+        assert (stream is None) ^ (channel is None)
```
karthiknadig: >(stream is None) ^ (channel is None) [](start = 15, length = 36)

The rare XOR operator usage :)

int19h: While we're at it, I feel very stymied by the lack of logical implication operator in Python. 😂

src/ptvsd/adapter/ide.py
```diff
@@ -149,19 +173,31 @@ def handle(self, request):
                 self.server.initialize(self._initialize_request)
                 self._initialize_request = None
 
+                arguments = request.arguments
+                if self.launcher and "RedirectOutput" in debug_options:
+                    # The launcher is doing output redirection, so we don't need the
+                    # server to do it, as well.
+                    arguments = dict(arguments)
+                    arguments["debugOptions"] = list(debug_options - {"RedirectOutput"})
```
karthiknadig: >arguments["debugOptions"] = list(debug_options - {"RedirectOutput"}) [](start = 20, length = 68)

This is such a pretty syntax to remove items from a list.

src/ptvsd/adapter/ide.py
```diff
@@ -222,58 +260,101 @@ def launch_request(self, request):
...
+        # back to us. Since the injected server can crash, there must be a timeout.
+        #
+        # If "subProcessId" is specified, this is attach to a known subprocess, likely
+        # in response to a "ptvsd_attach" event. If so, the debug server should be
+        # connected already, and thus the wait timeout is zero.
+        #
+        # If neither are specified, but "listen" is true, this is attach-by-socket
+        # with the server expected to connect to the adapter via ptvsd.attach(). There
+        # is no PID known in advance, so just wait until the first server connection
+        # indefinitely, with no timeout.
```
karthiknadig: >indefinitely, with no timeout. [](start = 10, length = 30)

Do we need a way to time out?

int19h: No, this is intentional - the whole point of "listen" is to have an adapter waiting for the server to attach to it, however long that takes.

src/ptvsd/adapter/ide.py
```diff
@@ -222,58 +260,101 @@ def launch_request(self, request):
...
-            self._start_request.respond({})
-            self._start_request = None
+            self.start_request.respond({})
             self._propagate_deferred_events()
 
+        # Notify the IDE of any child processes of the debuggee that aren't already
+        # being debugged.
+        for conn in server.connections():
+            if conn.server is None and conn.ppid == self.session.pid:
+                # FIXME: race condition with server.Connection()
```
karthiknadig: Do we need a bug for this?

src/ptvsd/adapter/ide.py
```diff
@@ -222,58 +260,101 @@ def launch_request(self, request):
...
+        sub_pid = request("subProcessId", int, optional=True)
+        if pid != ():
+            if sub_pid != ():
                 raise request.isnt_valid(
-                    '"attach" with "processId" cannot be serviced by adapter '
-                    "that is already associated with a debug server"
+                    '"processId" and "subProcessId" are mutually exclusive'
                 )
-
             ptvsd_args = request("ptvsdArgs", json.array(unicode))
```
karthiknadig: Unrelated to this PR. I think we should get rid of `ptvsdArgs`. Other than logging we don't really use it for anything. In case of logging we have a `--log-dir` anyway, that the IDE can pass to the adapter and server.

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_defaults.py
```diff
@@ -1,8 +1,8 @@
 '''
 This module holds the customization settings for the debugger.
 '''
-from _pydevd_bundle.pydevd_constants import QUOTED_LINE_PROTOCOL
+from _pydevd_bundle.pydevd_constants import HTTP_JSON_PROTOCOL
 
 
 class PydevdCustomization(object):
-    DEFAULT_PROTOCOL = QUOTED_LINE_PROTOCOL
+    DEFAULT_PROTOCOL = HTTP_JSON_PROTOCOL
```
fabioz: This should be fixed by https://github.com/microsoft/ptvsd/pull/1881 (in review).

src/ptvsd/adapter/server.py
```diff
@@ -141,4 +261,116 @@ def exited_event(self, event):
...
+    and returns the corresponding Connection.
+
+    If there is more than one server connection already available, returns the oldest
+    one.
+    """
+
+    def wait_for_timeout():
+        time.sleep(timeout)
+        wait_for_timeout.timed_out = True
+        with _lock:
```
fabioz: Minor: I don't think you need the lock to set the event.

src/ptvsd/adapter/server.py
```diff
@@ -45,12 +161,32 @@ class Capabilities(components.Capabilities):
             "supportedChecksumAlgorithms": [],
         }
 
-    def __init__(self, session, stream):
-        super(Server, self).__init__(session, stream)
+    def __init__(self, session, connection):
+        with _lock:
```
fabioz: The only place I could see creating the server is in `attach_to_session`:

```
with _lock:
    ...
    self.server = Server(session, self)
```

Maybe the `Server` could actually be pseudo-private (i.e.: `_Server`) to notify that it's constructor should not be used anywhere else and then the server constructor wouldn't need to acquire the lock and do `connection.server = self` since this is already taken care of (unless this really needs to be set before the remainder of the constructor -- if that's the case, it'd be nice to add some comment regarding it).

src/ptvsd/adapter/server.py
```diff
@@ -45,12 +161,32 @@ class Capabilities(components.Capabilities):
...
+            connection.server = self
+        self.connection = connection
 
-        self.pid = None
+        super(Server, self).__init__(session, channel=connection.channel)
+
+        self.pid = connection.pid
```
fabioz: Maybe the `pid` and `ppid` could be properties that always got the value from the connection? (it doesn't seem like the connection is ever set to None afterwards)

src/ptvsd/adapter/server.py
fabioz: On a 2nd thought, maybe we do (and we should also add the `with _lock` to `dont_expect_connections`).

src/ptvsd/adapter/server.py
```diff
@@ -141,4 +261,116 @@ def exited_event(self, event):
...
+            _connections_changed.clear()
+            if not len(_connections):
+                return
+
+
+def dont_expect_connections():
+    """Unblocks any pending wait_until_disconnected() call that is waiting on the
+    first server to connect.
+    """
+    _connections_changed.set()
```
fabioz: (add `with _lock` here)

tests/ptvsd/server/test_multiproc.py
```diff
@@ -15,130 +15,123 @@
...
-    platform.system() != "Windows",
-    reason="Debugging multiprocessing module only works on Windows",
-)
 @pytest.mark.parametrize(
-    "start_method", [runners.launch, runners.attach_by_socket["cli"]]
+    "start_method",
+    [""]
+    if sys.version_info < (3,)
+    else ["spawn"]
+    if platform.platform == "Windows"
```
fabioz: `platform.platform` is actually a callable (i.e.: platform.platform() returns `'Windows-10-10.0.17763-SP0'` to me).

I think you could use `sys.platform == 'win32'` instead to check if it's windows...

src/ptvsd/adapter/ide.py
int19h: Yep. I'll file one after merging the PR, so that I can reference the commit/line directly.

src/ptvsd/adapter/server.py
int19h: The idea was to sync with `ServerConnection.disconnect()`. But looking at it again, it's not doing that correctly, because `Server.disconnect()` can end up getting invoked before there's enough object state present for it to execute correctly.  I was trying to avoid holding `_lock` for the entirety of the constructor here, but it seems unavoidable. And then, yeah, it makes more sense for `attach_to_session()` to take care of it all.

tests/ptvsd/server/test_multiproc.py
int19h: This was meant to be `platform.system()`, which is generally what we use in the tests for OS checks.

fabioz: Well, that's not wrong per-se, but usually I'd advise to use `sys.platform` which is set at compile-time.

From what I remember `platform.system()` does end up spawning a process to get its info... just as an anecdote, I had an issue once -- several years ago -- where that call halted the process startup... I think it was back in Python 2.4, so, it may not be an issue anymore, but just the fact that it needs to spawn a process to get that info which should usually be as good as `sys.platform` made me always default to `sys.platform` -- see: https://github.com/microsoft/ptvsd/blob/master/src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py#L73 for how `pydevd` defines the related constants to know the OS.

src/ptvsd/adapter/ide.py
int19h: It just needs to pass it to the adapter - the adapter already has the code to propagate it to the launcher, and from there to the debuggee.

tests/ptvsd/server/test_multiproc.py
int19h: Interesting, I didn't know that! We should look into it, but given the pervasive use of `platform` in the tests, and occasionally in the adapter, it should be a separate issue to make sure we change them all.

int19h: https://github.com/microsoft/ptvsd/issues/1884

PR #1882, opened 2019-10-29
Report ppid on pydevdSystemInfo on Python 2.7 on Windows. Fixes #1877

### LINKED ISSUE #1877: "pydevdSystemInfo" does not report "ppid" on Python 2.7 on Windows

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -808,6 +806,44 @@ def set_source_mapping(self, py_db, source_filename, mapping):
...
+        if ppid is None and IS_WINDOWS:
+            ppid = self._get_windows_ppid()
+
+        return ppid
+
+    def _get_windows_ppid(self):
+        this_pid = os.getpid()
+
+        list_popen = self._popen(
+            ['wmic', 'process', 'where', '(ProcessId=%s)' % (this_pid,), 'get', 'ParentProcessId'],
```
karthiknadig: @DonJayamanne had mentioned that `wmic` had previously caused a lot of issues with leaving some processes open. Also should this run under the guard to prevent injecting into child process.

fabioz: Regarding the injection, this shouldn't be a problem (i.e.: we don't inject into non-python processes)...

As for the issues with `wmic`, we also use it to list the processes to kill, so, if this usage is a problem, we should also use a different approach there... @DonJayamanne can you give some details on the problems you had with `wmic`? Is the usage here bound to give us problems later on?

karthiknadig: One way I know using native APIs is to call `kernel32!CreateToolhelp32Snapshot` get the Process snapshot. Then use `kernel32!Process32First` and `kernel32!Process32Next` functions to iterate over the snapshot to find the a match for the current pid. The returned object will be https://docs.microsoft.com/en-us/windows/win32/api/tlhelp32/ns-tlhelp32-processentry32 which has `th32ParentProcessID`.

fabioz: @karthiknadig I'll change both usages of `wmic` to use `CreateToolhelp32Snapshot` (it really seems better than spawning a process) -- although I'm still curious on the actual problems with `wmic` ;)

int19h: Note that we only really need `ppid` for subprocesses - having it there for the primary process doesn't hurt, but it doesn't do anything useful, either. And in the subprocess scenario, the parent process is the one injecting the debugger, so it can pass its own PID directly to the child, without querying the OS for it.

fabioz: Well, the latest commit is already using the create snapshot (and it fixed the other use case to avoid using wmic too), so, I think that should be a reasonable solution already.

karthiknadig: @fabioz For Python extension they were using `wmic` to detect and terminate any stray python process. This eventually led to several stray `wmic` processes being left behind. See https://github.com/microsoft/vscode-python/issues/1036
I think this is what @DonJayamanne was talking about as an issue with `wmic`.

The reason why most extensions use `wmic` is because of the lack of native support until Node NAPI support came in. Currently VS Code does not ship with a node version that supports NAPI. So extensions are stuck with `wmic`. Once that is available, extensions are likely to switch to https://www.npmjs.com/package/windows-process-tree which uses native APIs to get this info.

PR #1891, opened 2019-10-31
Harden code which deals with CreateToolhelp32Snapshot. Fixes #1877

### LINKED ISSUE #1877: "pydevdSystemInfo" does not report "ppid" on Python 2.7 on Windows

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
```diff
@@ -969,19 +1066,81 @@ class PROCESSENTRY32(ctypes.Structure):
                     ("szExeFile", ctypes.c_char * 260)]
```
int19h: I think the problem was here, actually. Here's the corresponding C struct:
```c
typedef struct tagPROCESSENTRY32 {
  DWORD     dwSize;
  DWORD     cntUsage;
  DWORD     th32ProcessID;
  ULONG_PTR th32DefaultHeapID;
  DWORD     th32ModuleID;
  DWORD     cntThreads;
  DWORD     th32ParentProcessID;
  LONG      pcPriClassBase;
  DWORD     dwFlags;
  CHAR      szExeFile[MAX_PATH];
} PROCESSENTRY32;
```
Note that `th32DefaultHeapID` is ULONG_PTR, which means that it's 64 bits in a 64-bit process.  But `ctypes.c_ulong` is always going to be 32-bit on Windows. It should be `c_size_t`.

I would also change all DWORD members to `c_uint32`, since that's what the typedef is meant to represent.

int19h: Let me try this change locally and see if that fixes it...

int19h: Yep, now it works! Let me make for that.

int19h:
Given that this was a problem with the way the API was invoked, I don't think we need to do any hardening here. Per @karthiknadig, there's practically nothing there that can fail when invoked correctly, short of extreme scenarios such as running out of memory.

fabioz:
Thanks for the investigation... I thought it was something along the lines of https://github.com/gperftools/gperftools/issues/203 since it was failing spuriously (and I couldn't really replicate locally).

I'll close this pull request given that you'll create a new one with the proper construct.

PR #1902, opened 2019-11-11
Fix #1875: Inject ptvsd into subprocesses

### LINKED ISSUE #1875: Inject ptvsd into subprocesses

int19h:
Use "evaluate" request without "frameId" to import ptvsd into the debuggee as soon as server attaches.

Fix test_subprocess to use the new test API.

### DIFF
--- a/src/ptvsd/_vendored/pydevd/pydevd.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd.py
@@ -164,7 +164,7 @@ class PyDBCommandThread(PyDBDaemonThread):
     @overrides(PyDBDaemonThread._on_run)
     def _on_run(self):
         # Delay a bit this initialization to wait for the main program to start.
-        time.sleep(0.3)
+        self._py_db_command_thread_event.wait(0.3)
 
         if self._kill_received:
             return
@@ -1253,6 +1253,9 @@ class PyDB(object):
         else:
             internal_cmd = InternalThreadCommand(thread_id, method, *args, **kwargs)
         self.post_internal_command(internal_cmd, thread_id)
+        if thread_id == '*':
+            # Notify so that the command is handled as soon as possible in the PyDBCommandThread.
+            self._py_db_command_thread_event.set()
 
     def post_internal_command(self, int_cmd, thread_id):
         """ if thread_id is *, post to the '*' queue"""
@@ -2598,10 +2601,10 @@ def _locked_settrace(
         if not wait_for_ready_to_run:
             py_db.ready_to_run = True
 
+        py_db.start_auxiliary_daemon_threads()
+
         while not py_db.ready_to_run:
             time.sleep(0.1)  # busy wait until we receive run command
-
-        py_db.start_auxiliary_daemon_threads()
 
         if trace_only_current_thread:
             py_db.enable_tracing()
--- a/src/ptvsd/_vendored/pydevd/pydevd.py
+++ b/src/ptvsd/_vendored/pydevd/pydevd.py
@@ -620,6 +620,13 @@ class PyDB(object):
         # Stop the tracing as the last thing before the actual shutdown for a clean exit.
         atexit.register(stoptrace)
 
+    def wait_for_ready_to_run(self):
+        while not self.ready_to_run:
+            # busy wait until we receive run command
+            self.process_internal_commands()
+            self._py_db_command_thread_event.clear()
+            self._py_db_command_thread_event.wait(0.1)
+
     def on_initialize(self):
         '''
         Note: only called when using the DAP (Debug Adapter Protocol).
@@ -631,6 +638,7 @@ class PyDB(object):
         Note: only called when using the DAP (Debug Adapter Protocol).
         '''
         self._on_configuration_done_event.set()
+        self._py_db_command_thread_event.set()
 
     def is_attached(self):
         return self._on_configuration_done_event.is_set()
@@ -652,8 +660,18 @@ class PyDB(object):
         else:
             return system_exit_exc in self._ignore_system_exit_codes
 
-    def block_until_configuration_done(self, timeout=None):
-        return self._on_configuration_done_event.wait(timeout)
+    def block_until_configuration_done(self, cancel=None):
+        if cancel is None:
+            cancel = NULL
+
+        while not cancel.is_set():
+            if self._on_configuration_done_event.is_set():
+                cancel.set()  # Set cancel to prevent reuse
+                return
+
+            self.process_internal_commands()
+            self._py_db_command_thread_event.clear()
+            self._py_db_command_thread_event.wait(1 / 15.)
 
     def add_fake_frame(self, thread_id, frame_id, frame):
         self.suspended_frames_manager.add_fake_frame(thread_id, frame_id, frame)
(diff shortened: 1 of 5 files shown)


int19h:
Do not merge - this breaks things at the moment due to some unclear issue with "evaluate".

int19h:
@fabioz, can you take a look and verify whether I'm using "evaluate" correctly. For some reason, the "evaluate" request doesn't seem to be getting a response from pydevd in most scenarios, but not all. Since the injection is set up to happen even in single-proc, this can be trivially observed with `test_run` - attach-by-socket tests pass, but everything else fails. In pydevd logs, it looks like this:
```
Process EvaluateRequest: {
    "arguments": {
        "expression": "\nimport sys\nsys.path.insert(0, 'c:\\\\git\\\\ptvsd\\\\src')\ntry:\n    import ptvsd\nfinally:\n    del sys.path[0]\n",
        "format": {}
    },
    "command": "evaluate",
    "seq": 2,
    "type": "request"
}
```
and there's no response that follows, so the adapter just times out waiting for it.

I tried to trace the flow of the request inside of pydevd with prints. So far as I can see, it handles the request to the point where `internal_evaluate_expression_json` is posted to the command thread; but it never actually runs.

fabioz:
I'll take a look.

fabioz:
The issue is that in the 'settrace', pydevd does not start the auxiliary threads while it's not ready to run (so, it doesn't handle requests while it's not unblocked).

A fix for this would be starting those threads before doing that wait. I did the needed changes for that in  https://github.com/fabioz/ptvsd/commit/750f1bf16bdf867256c8f6ddd33bc91f6ac3cd26 (so, you can cherry-pick those changes in your branch).

int19h:
@fabioz  Thanks! This solves the issue for `test_run`, but it appears to still be present for subprocesses. Same symptom - "evaluate" is received by pydevd, but no response is sent back, and there's nothing in the logs indicating any failure to process.

`test_subprocess` (any variant, e.g. `test_subprocess[program-launch]`), with changes in this PR, can be used to observe this. The test itself fails due to backchannel timeout, but that happens because the backchannel is being opened by the child process, and that code doesn't execute because the subprocess is not getting attached to.

fabioz:
I'll investigate the issue.

fabioz:
I started investigating the issue on Python 2.7 (where it seems that any launch with `runners.attach_by_socket["cli"]` deadlocks).

In this case the problem is the Python 2.7 import lock -- I'm posting the thread dump below, which I think is clear in showing why the deadlock happens.

```

-------------------------------------------------------------------------------
 Thread pydevd.CommandThread  (daemon: True, pydevd thread: True)

 File "C:\bin\Miniconda\envs\py27_tests\lib\threading.py", line 773, in __bootstrap
   self.__bootstrap_inner()
 File "C:\bin\Miniconda\envs\py27_tests\lib\threading.py", line 800, in __bootstrap_inner
   self.run()
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_comm.py", line 156, in run
   self._on_run()
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\pydevd.py", line 175, in _on_run
   self.py_db.process_internal_commands()
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\pydevd.py", line 1468, in process_internal_commands
   int_cmd.do_it(self)
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_comm.py", line 578, in do_it
   InternalThreadCommand.do_it(self, dbg)
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_comm.py", line 547, in do_it
   self.method(dbg, *self.args, **self.kwargs)
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\_pydevd_bundle\pydevd_comm.py", line 1000, in internal_evaluate_expression_json
   dump_threads(stream=_LoggingGlobals._debug_stream)

# ... when about to evaluate "import ptvsd" it's never able to get the import lock.

-------------------------------------------------------------------------------
 Thread pydevd.Reader  (daemon: True, pydevd thread: True)
 
    -- nothing special here

-------------------------------------------------------------------------------
 Thread pydevd.Writer  (daemon: True, pydevd thread: True)

   -- nothing special here

-------------------------------------------------------------------------------
 Thread pydevd.CheckAliveThread  (daemon: False, pydevd thread: True)

  -- nothing special here

-------------------------------------------------------------------------------
 Thread MainThread  (daemon: False, pydevd thread: False)

 File "c:\Users\fabio\AppData\Local\Temp\pytest-of-fabio\pytest-451\test_step_multi_threads_progra0\code_to_debug.py", line 5, in <module>

   import debug_me  # noqa -- holding the import lock on Python 2.7 here!!!

   self: <Condition(<thread.lock object at 0x00000000039FDE30>, 1)>
 File "d:\x\ptvsd_workspace\ptvsd\tests\DEBUGGEE_PYTHONPATH\debug_me\__init__.py", line 44, in <module>
   eval(_code, {})
 File "<PTVSD_TEST_DEBUG_ME>", line 5, in <module>
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\__init__.py", line 47, in wait_for_attach
   return api.wait_for_attach()
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\server\api.py", line 35, in wait_for_attach
   pydevd._wait_for_attach(cancel=cancel_event)
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\pydevd.py", line 2424, in _wait_for_attach
   if py_db.block_until_configuration_done(0.1):
 File "D:\x\ptvsd_workspace\ptvsd\src\ptvsd\_vendored\pydevd\pydevd.py", line 656, in block_until_configuration_done
   return self._on_configuration_done_event.wait(timeout)
 File "C:\bin\Miniconda\envs\py27_tests\lib\threading.py", line 613, in wait
   self.__cond.wait(timeout)
 File "C:\bin\Miniconda\envs\py27_tests\lib\threading.py", line 358, in wait
   _sleep(delay)
```

fabioz:
Ok, I did another change to support this use case -- you can cherry-pick it at: https://github.com/fabioz/ptvsd/commit/d34bde485812c54331fabb917672ae9520679ea2

I changed to process events in the thread that's currently waiting instead of starting threads as I did previously.

I also had to do a change to check if `ptvsd` was already loaded prior to the import to handle the case where the user does:

```
import ptvsd
ptvsd.enable_attach(...)
ptvsd.wait_for_attach()
```

In this case `ptvsd` will be already loaded and in `sys.modules`, and in that case the evaluation will be done in a thread (which would deadlock in the import statement) because we're no longer waiting for the initial configuration done.

int19h:
Looks good, thanks!

PR #1903, opened 2019-11-11
Ensure attach to pid is bitness agnostic. Fixes #1624

### LINKED ISSUE #1624: Ensure attach to pid is bitness agnostic

fabioz:


src/ptvsd/_vendored/pydevd/pydevd_attach_to_process/windows/inject_dll.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+        const LPVOID procAddress = GetProcAddress(dllHandle, &procName[0]);
+        if(procAddress == nullptr) {
+            std::cout << "Error. Unable to GetProcAddress for: " << procName << ". Error code: " << GetLastError() << "." << std::endl;
+            return 9;            
+        }
+        
+        std::cout << "Calling " << procName << " in dll." << std::endl;
+        
+        const HANDLE remoteThreadCustomProc = CreateRemoteThread(hProcess, nullptr, 0, (LPTHREAD_START_ROUTINE) procAddress, nullptr, 0, nullptr);
```
fabioz: @int19h @karthiknadig 

Note: instead of doing it on `DllMain` I'm creating a new remote thread to call the function (it was a easier than having yet another dll just for the `DllMain` -- because for reattach to work properly we have to load the dll and then free it, but we can't free at the wrong time because of running threads)... 

The catch is the following: on my tests loading the same dll locally/GetProcAddress always yielded proper results to call the method in the other process -- this works locally for me and on the ci -- both for the kernel32 dll as for my own dll... do you know of a case where this won't work?

I didn't really find much documentation regarding this detail -- if there's something I'm missing here, I'll have to revert to creating a new dll and doing it in `DllMain` and dealing with synchronizing `LoadLibrary`/`FreeLibrary` with threads -- providing that it still works for the kernel 32 dll (or using shellcode as I was doing before)... if there's a case where this wouldn't work for my own dll nor the kernel32 dll, then the whole attach would fail regardless of that.

int19h: I don't think it's a hard guarantee. It's going to be the same only so long as the base address of the DLL is the same, *and* the OS could actually load the DLL at that address. Since we're injecting this into some unknown code, it's entirely possible for that address range to already be used by some other DLL that happened to request it, and then ours will be reallocated. This is less likely to happen in 64-bit process than in 32-bit, but it's always a possibility. 

Another thing that's suspicious - this shouldn't reliably work in general if the injected DLL is compiled with /dynamicbase:on (ASLR), which should be the default on all recent VC++ versions. Are we disabling that flag somewhere? We shouldn't, unless we have a good reason to.

fabioz: /dynamicbase is not disabled (as the compilation is pretty simple I just have a .bat -- you can see how the compile is done at: https://github.com/microsoft/ptvsd/pull/1903/files#diff-fb91f3f2a87c9d1771a22ee959437d89R3 -- I also tried explicitly adding the flag and everything still works on my tests).

Anyways, as it's not a hard guarantee, I've changed it to use `DllMain` (with the proper synchronization).

fabioz:
Note: I'll do some changes to use the pid in the mutexes/shared memory so that multiple processes can attach in parallel (so, please wait for that before merging).

fabioz:
@int19h @karthiknadig this is ready to review (if it's ok, please merge for me -- failures are the same from `master`).

PR #1911, opened 2019-11-13
[DONT MERGE] Close debugger connection on exec. Fixes #1909

### LINKED ISSUE #1909: Fork/exec creates two connections from pydevd to adapter
This is causing `test_subprocess` to fail on Python 2.7, and is also the root cause of Flask test failures.

The problem is that on 2.7, `subprocess.Popen` uses fork+exec - and pydevd detours both. So first `fork()` creates a child process, and pydevd is spun up in that process and connects to the adapter. Then the child process does `exec()`, pydevd patches command line to ensure that it gets loaded first, and as it does so, it connects to the adapter again. Since both connections come from a process with the same PID, the adapter rejects the second one.

There is a bug in the adapter here as well (#1908), but the fundamental problem is that it really is the _same_ process trying to connect twice. The adapter can handle it by forcibly closing any existing connection with the same PID when a new one comes in; but I wonder if it might be better to prevent this on pydevd side, e.g. by having the code detouring `exec()` close the socket before replacing the process. Per `man execve`, the OS itself will keep the socket fd open unless it's explicitly closed, so the adapter can't just wait for it to go away.

fabioz:


fabioz:
Note: please don't merge yet (I'm still checking some things on the pydevd side).

PR #1921, opened 2019-11-15
Don't notify about threads before the debugger is ready to run. Fixes #1919

### LINKED ISSUE #1919: pydevd should not notify about threads before ready to run
This was a bad side effect of https://github.com/microsoft/ptvsd/commit/4ce4e709ceae389b2b38db0137e9b35c20d3ff1e (I'll provide a fix shortly).

fabioz:


fabioz:
@int19h or @karthiknadig can you merge for me? (I don't have permissions to do that when the ci is failing -- the failures seem to be the same ones from the master branch)

PR #1925, opened 2019-11-19
Reduce running time for CI test runs

int19h:
- reduce test matrix where not necessary
- switch to `sys.platform`

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
 testpaths=tests
-timeout=20
+timeout=30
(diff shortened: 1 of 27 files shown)


int19h:
@karthiknadig This should be good for review. Unfortunately the runtimes on Linux haven't improved much - I suspect this is because the process spawn overhead is low there (this is also why Travis is only marginally faster). But on Windows, which is the longer running one anyway, this has somewhere in the ballpark of 40-50% improvement. All test failures are known and have bugs logged for them.

PR #1962, opened 2019-12-03
Support for step over and step return with asyncio coroutines. Fixes #1627

### LINKED ISSUE #1627: Consider coroutines when stepping (check with asyncio and trio).
## Environment data

- VS Code version: 1.36.1
- Extension version: 2019.6.24221
- OS and version: Windows 10 / Version 1903 / Windows_NT x64 10.0.18362 /
- Python version: Python 3.7.3 x64
- Type of virtual environment used: venv
- Relevant/affected Python packages and their versions: trio 0.11.0
- Jedi or Language Server?: Language server (false)

## Expected behaviour

When debugging trio async/await code, continue, stepping over, and out of functions doesn't work.

When stopped on a breakpoint, pressing continue (F5) should continue execution to the next breakpoint.

When stopped on a break point, pressing step over (F10) should proceed to next line of code.

When inside an async function, pressing step out(Shift+F11) should return to the the next line of the calling scope

## Actual behaviour

Behaviour seems erratic. Sometimes I get an error. Sometimes something unexpected happens.

Pressing continue stops on the same line as where you had the breakpoint. Sometimes an error is opened in a new tab, with the following or similar message: "Could not load source '<string>': Source unavailable."

Pressing step over stops on the same line as where you had the breakpoint.

Pressing step out at first doesn't do anything. Pressing it again steps out to the same line as where you stepped in.

## Steps to reproduce:
1.  Use example code:

```
import trio

async def func1():
    await trio.sleep(1)

async def func2():
    await trio.sleep(2)

async def func3():
    await trio.sleep(3)

async def main():
    await func1() # <--- Try putting break point here.
    await func2()
    await func3()

trio.run(main)
```

## Logs

Code started with following parameters

```
(.venv) PS C:\Users\Anders\Documents\python\trio-debugging> cd 'c:\Users\Anders\Documents\python\trio-debugging'; ${env:PYTHONIOENCODING}='UTF-8'; ${env:PYTHONUNBUFFERED}='1'; & 'c:\Users\Anders\Documents\python\trio-debugging\.venv\Scripts\python.exe' 'c:\Users\Anders\.vscode\extensions\ms-python.python-2019.6.24221\pythonFiles\ptvsd_launcher.py' '--default' '--client' '--host' 'localhost' '--port' '55265' 'c:\Users\Anders\Documents\python\trio-debugging\debugging.py'
```

fabioz:
As a note, this doesn't completely fix #1627 (it's still missing the step in part which will be done afterwards).


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_cython.pyx
```diff
@@ -526,62 +510,121 @@ cdef class PyDBFrame:
...
+                        return_lines = set()
+                        for x in main_debugger.collect_return_info(frame.f_code):
+                            # Note: cython does not support closures in cpdefs (so we can't use
+                            # a list comprehension).
+                            return_lines.add(x.return_line)
+
+                        frame_skips_cache[returns_cache_key] = return_lines
+
+                    if line not in return_lines:
+                        # Not really a return (coroutine/generator paused).
```
int19h: I wonder if this should perhaps be special-cased for "step in", so that it's possible to step into a yield, and land at the point where `next()` is being called? It can be handy sometimes. 

Not sure whether the equivalent for `await` would make much sense, though...

fabioz: Right now if the user does a step in (instead of a step over), he'll get to the point where the `next()` is being called (which is what I want to fix with asyncio as with generators in general that's where the user wants to stop, but with coroutines I don't think he ever wants to get into the asyncio code).

I agree that for generators in general we should revert to the usual step in (which would get to the `next()` being called). My plan is fixing that soon (when I also fix the step in for coroutines).

PR #1968, opened 2019-12-04
Fix #1967: attach_by_socket tests fail sporadically on MacOS

### LINKED ISSUE #1967: attach_by_socket tests fail sporadically on MacOS + Python 3.6
When it happens, it's usually several dozen tests failing in the same run for a given Python version. The only common theme is the attach method - the tests themselves are completely unrelated.

int19h:
Override inherited timeouts for sockets returned by accept() to ensure that they're blocking.


int19h:
We already do this in a different place, so that's not really helping.

PR #1971, opened 2019-12-06
Disable "redirectOutput" in tests by default.

int19h:


int19h:
This is to reflect the corresponding change to the default value for this property in ptvsd itself and in vscode-python.

PR #1973, opened 2019-12-09
Fix #1711: adapter: access tokens

### LINKED ISSUE #1711: adapter: access tokens
Adapter should implement the counterpart to #1710 in pydevd. This includes:

1. A command line argument that allows to specify the server access token (used by `ptvsd` in `enable_attach()` mode).
1. Generation of a cryptographically secure client access token.
1. Passing the server token to pydevd in the "initialize" request when connecting.
1. Verifying that the client token returned by pydevd in the "initialize" response matches the generated client token.

int19h:
Generate server access token in enable_attach(), propagate it to the adapter, and have the adapter authenticate to the server via "pydevdAuthorize".

Generate client access token in adapter when not spawned by server, and propagate it to pydevd.

Fix pydevd to correctly propagate access tokens for subprocesses that are not forked.

Simplify options handling.


int19h:
@fabioz, can you take a look at the bits that I changed in pydevd?

PR #2005, opened 2019-12-19
Accept new style properties for debugger options. Fixes #2001

### LINKED ISSUE #2001: PTVSD 5.* debugger Flags are not working
I'm setting the debugger flags in the launch json, but they are being ignored by the debugger. For instance, the show return value and break on system exit 0 are not working. 

Launch JSON: 
image

DebugAdapterLogs for PTVSD 5.*
[DebugAdapterHost5.txt](https://github.com/microsoft/ptvsd/files/3980429/DebugAdapterHost5.txt)

DebugAdapterLogs for PTVSD 4.0 (for comparison)
[DebugAdapterHost4.txt](https://github.com/microsoft/ptvsd/files/3980432/DebugAdapterHost4.txt)

fabioz:


fabioz:
Please don't merge/review yet (it seems there are some tests breaking, so, I'll take a look at that...)

int19h:
It looks like the failing tests are those that have to do with redirected output. This is probably because the adapter needs to tell pydevd to *not* do output redirection when the launcher is doing it, and it does so by tweaking "debugOptions" - but it doesn't adjust the property:
https://github.com/microsoft/ptvsd/blob/d9ef3ddd2d98be8d12dfe5743d391ff7965c5aa0/src/ptvsd/adapter/ide.py#L182-L186

src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
```diff
@@ -1,14 +1,99 @@
...
+            self.show_return_value = bool_parser(args['showReturnValue'])
+
+        if 'breakOnSystemExitZero' in args:
+            self.break_system_exit_zero = bool_parser(args['breakOnSystemExitZero'])
+
+        if 'django' in args:
+            self.django_debug = bool_parser(args['django'])
+
+        if 'flask' in args:
+            self.flask_debug = bool_parser(args['flask'])
```
karthiknadig:
```suggestion
            self.flask_debug = bool_parser(args['flask'])

        if 'jinja' in args:
            self.flask_debug = bool_parser(args['jinja'])        
```

fabioz: Done

PR #2007, opened 2019-12-21
Fix #2004: Adapter process is not fully detached from parent server

### LINKED ISSUE #2004: Adapter process is not fully detached from parent server
When `enable_attach()` spawns the adapter, it's just a regular child process. This has some unfortunate consequences.

On Windows, it inherits the console handle, and thus receives Ctrl+C and other console events, and will be terminated if the console is closed. This can be fixed by using `CREATE_NEW_PROCESS_GROUP` and `DETACHED_PROCESS` creation flags.

On Unix, the adapter becomes orphaned, because there's no parent process to reap it (in all scenarios, the adapter exits after the server does). It should be properly daemonized with the usual double-fork technique, so that it becomes a child of `init`, which will then take care of reaping it properly. It should also use `os.setsid()` between the forks to fully detach from the terminal.

int19h:
Close stdin, stdout, and stderr in the adapter.

Double-fork() and setsid() on Unix to daemonize properly.

Use CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW on Win32 to daemonize properly.

### DIFF
--- a/src/ptvsd/server/api.py
+++ b/src/ptvsd/server/api.py
@@ -9,25 +9,28 @@ import contextlib
 import json
 import os
 import pydevd
+import socket
 import sys
 import threading
 
 import ptvsd
-from ptvsd.common import compat, log
+from ptvsd import adapter
+from ptvsd.common import compat, log, sockets
 from ptvsd.server import options
 from _pydevd_bundle.pydevd_constants import get_global_debugger
 from pydevd_file_utils import get_abs_path_real_path_and_base_from_file
 
 
-_QUEUE_TIMEOUT = 10
-_ADAPTER_PATH = os.path.join(os.path.dirname(ptvsd.__file__), "adapter")
+def _settrace(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


src/ptvsd/server/api.py
```diff
@@ -69,54 +72,116 @@ def debug(address, log_dir=None, multiprocess=True):
...
+        )
+        if os.name == "posix":
+            # It's going to fork again to daemonize, so we need to wait on it to
+            # clean it up properly.
+            enable_attach.adapter.wait()
+        else:
+            # Suppress misleading warning about child process still being alive when
+            # this process exits (https://bugs.python.org/issue38890).
+            enable_attach.adapter.returncode = 0
+            pydevd.add_dont_terminate_child_pid(enable_attach.adapter.pid)
```
karthiknadig: Shouldn't we be adding this in `posix` case as well?

int19h: In the POSIX case, the immediate child exits immediately after forking itself (and is waited on above). The forked daemon has a different PID, but in any case, it's not treated as a child anymore.

PR #2018, opened 2019-12-27
Don't show empty string when hovering over bogus word. Fixes #2011

### LINKED ISSUE #2011: Debug hover: shows empty when hovering over bogus words
Hi, VS Code dev here :wave:

1. Have a python program and hover over some bogus words - like part of a string
2. Debug hover appears with no content
3. The issue here is that the python debug for `evaluate` of `"This"` returns a successful response. I suggest that for evaluating bogus expressions the python debugger returns a response with `success = false`

In this picture I am hovering over `This`.

Related issue: have a bogus expression in the WATCH view. The value is a structured object, instead it shuold just be an error like in the debug console

Screenshot 2019-12-24 at 15 16 03

Screenshot 2019-12-24 at 15 17 57

fyi @qubitron @DonJayamanne

fabioz:


src/ptvsd/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -982,7 +982,7 @@ def __create_frame():
         is_error = isinstance(eval_result, ExceptionOnEvaluate)
         if is_error:
             if context == 'hover':  # In a hover it doesn't make sense to do an exec.
-                _evaluate_response(py_db, request, result='')
+                _evaluate_response(py_db, request, result='', error_message='Exception evaluating.')
```
int19h: The error message is unclear - should be something like "Exception occurred during evaluation".

PR #2042, opened 2020-01-15
Accept bytes when patching args. Fixes #2309

fabioz:


fabioz:
I'm closing this one if favor of https://github.com/microsoft/ptvsd/pull/2043 (both fixes were on the same code).

PR #2125, opened 2020-05-04
Deprecate ptvsd

int19h:
Update README to direct to debugpy.


int19h:
Render preview: https://github.com/microsoft/ptvsd/blob/c7dba54d31befb7387740a76e419921580266793/README.md

karthiknadig (approved):

ianlini:
I'm wondering why this is not merged because when I saw ptvsd and debugpy, it's not very obvious that I should use debugpy.

PR #2133, opened 2020-08-06
tox.ini: Add Python 3.8 and remove 3.4 and 3.5

cclauss:
https://devguide.python.org/#status-of-python-branches
https://devguide.python.org/devcycle/#end-of-life-branches  # Python 3.5 EOLs in one month.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,34,35,36,37}{,-cov}
+envlist = py{27,36,37,38}{,-cov}
 
 [testenv]
 deps = -rtests/requirements.txt


int19h:
We're not accepting any more changes to ptvsd; https://github.com/microsoft/debugpy is the new hotness.

But in any case, we do not strictly follow the official EOL dates. They are taken into consideration, but it's not the most important thing. Even if something is EOL'd, but still has a lot of users - e.g. Python 2.7 - we keep supporting it.

cclauss:
Cool.  Would it make sense to mention debugpy on the README of this repo?

int19h:
Yep, there's an outstanding PR for that - we just need to coordinate a few more things for it to go in.

